PR #14, opened 2019-01-13
New resource & data source 'azuread_group'

tiwood:
This PR adds a new resource and data source for Azure Active Directory groups.

The is based on the PR terraform-providers/terraform-provider-azurerm#1839

- [x] New data source 'azuread_group'
- [x] New resource 'azuread_group'
- [x] Add tests

## Example Usage

```hcl

// This creates a new Azure AD group with the display name "my_group"
resource "azuread_group" "my_group" {
 name = "my_group"
}

// This queries an Azure AD group by ID
data "azuread_group" "my_group_by_id" {
  object_id = "78722cfc-8946-11e8-95f1-2200ec79ad01"
}

// This queries an Azure AD group by Name
data "azuread_group" "my_group_by_name" {
  name = "my_group"
}
```

* `id` - The Object ID for the Azure AD Group.
* `name` - The Display Name for the Azure AD Group.

### DIFF
--- a/azuread/data_source_group.go
+++ b/azuread/data_source_group.go
@@ -1,0 +1,98 @@
+package azuread
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/ar"
+	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/validate"
+)
+
+func dataGroup() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceActiveDirectoryGroupRead,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"object_id"},
+				ValidateFunc:  validate.NoEmptyStrings,
+			},
+
+			"object_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"name"},
+				ValidateFunc:  validate.UUID,
+			},
+		},
+	}
+}
+
+func dataSourceActiveDirectoryGroupRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).groupsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	var adgroup graphrbac.ADGroup
+	var groupObj *graphrbac.ADGroup
+
+	if oId, ok := d.GetOk("object_id"); ok {
+		// use the object_id to find the Azure AD group
+
+		objectId := oId.(string)
+		resp, err := client.Get(ctx, objectId)
+		if err != nil {
+			if ar.ResponseWasNotFound(resp.Response) {
+				return fmt.Errorf("Error: AzureAD Group with ID %q was not found", objectId)
+			}
+
+			return fmt.Errorf("Error making Read request on AzureAD Group with ID %q: %+v", objectId, err)
+		}
+
+		adgroup = resp
+
+	} else {
+
+		// use the name to find the Azure AD group
+		name := d.Get("name").(string)
+		filter := fmt.Sprintf("displayName eq '%s'", name)
+		log.Printf("[DEBUG] [data_source_azuread_group] Using filter %q", filter)
+
+		resp, err := client.ListComplete(ctx, filter)
+		if err != nil {
+			return fmt.Errorf("Error listing Azure AD groups: %+v", err)
+		}
+
+		for _, v := range *resp.Response().Value {
+			if v.DisplayName != nil {
+				if strings.EqualFold(*v.DisplayName, name) {
+					log.Printf("[DEBUG] [data_source_azuread_group] %q (API result) matches %q (given value). The group has the objectId: %q", *v.DisplayName, name, *v.ObjectID)
+					groupObj = &v
+					break
+				} else {
+					log.Printf("[DEBUG] [data_source_azuread_group] %q (API result) does not match %q (given value)", *v.DisplayName, name)
+				}
+			}
+		}
+		if groupObj == nil {
+			return fmt.Errorf("Couldn't locate a Azure AD group with a name of %q", name)
+		}
+
+		adgroup = *groupObj
+	}
+
+	d.SetId(*adgroup.ObjectID)
+	d.Set("object_id", adgroup.ObjectID)
+	d.Set("name", adgroup.DisplayName)
+
+	return nil
+}
--- a/azuread/data_source_group.go
+++ b/azuread/data_source_group.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/ar"
 	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/validate"
 )
 
@@ -19,19 +18,9 @@ func dataGroup() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				Computed:      true,
-				ConflictsWith: []string{"object_id"},
-				ValidateFunc:  validate.NoEmptyStrings,
-			},
-
-			"object_id": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				Computed:      true,
-				ConflictsWith: []string{"name"},
-				ValidateFunc:  validate.UUID,
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
 			},
 		},
 	}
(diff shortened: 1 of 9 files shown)


tiwood:
Tests pass:

```shell
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAzureADGroup* -timeout 180m -ldflags="-X=github.com/terraform-providers/terraform-provider-azuread/version.ProviderVersion=acc"
?   	github.com/terraform-providers/terraform-provider-azuread	[no test files]
=== RUN   TestAccAzureADGroup_basic
--- PASS: TestAccAzureADGroup_basic (2.87s)
=== RUN   TestAccAzureADGroup_complete
--- PASS: TestAccAzureADGroup_complete (2.26s)
PASS
```

website/docs/d/group.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+description: |-
+  Gets information about an Azure Active Directory group.
+
+---
+
+# Data Source: azuread_group
+
+Gets information about an Azure Active Directory group.
+
+-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to both `Read directory data` within the `Windows Azure Active Directory` API.
```
perbergland: remove " both " - it seems misplaced if there is only one permission needed

tiwood: Thanks. Fixed in the latest commit.

tombuildsstuff:
thanks for porting this over @tiwood - will take a look shortly :)

azuread/data_source_group.go
```diff
@@ -0,0 +1,98 @@
...
+		filter := fmt.Sprintf("displayName eq '%s'", name)
+		log.Printf("[DEBUG] [data_source_azuread_group] Using filter %q", filter)
+
+		resp, err := client.ListComplete(ctx, filter)
+		if err != nil {
+			return fmt.Errorf("Error listing Azure AD groups: %+v", err)
+		}
+
+		for _, v := range *resp.Response().Value {
+			if v.DisplayName != nil {
```
tombuildsstuff: _minor_ we could probably make this more readable by inverting it? e.g.

```
if v.DisplayName == nil {
  continue
}
```

azuread/data_source_group.go
```diff
@@ -0,0 +1,98 @@
...
+				if strings.EqualFold(*v.DisplayName, name) {
```
tombuildsstuff: in general we tend to use exact matches for comparisons for Data Sources - as such I think it'd be worth doing that here; WDYT @tiwood?

azuread/resource_group.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func resourceGroupRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).groupsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	resp, err := client.Get(ctx, d.Id())
+	if err != nil {
+		if ar.ResponseWasNotFound(resp.Response) {
+			log.Printf("[DEBUG] [resource_azuread_group] Azure AD group with id %q was not found - removing from state", d.Id())
```
tombuildsstuff: We can ditch the `[resource_azuread_group]` prefix since this'll be output with the resource name at the start afaik

```suggestion
			log.Printf("[DEBUG] Azure AD group with id %q was not found - removing from state", d.Id())
```

azuread/data_source_group.go
```diff
@@ -0,0 +1,98 @@
...
+					log.Printf("[DEBUG] [data_source_azuread_group] %q (API result) matches %q (given value). The group has the objectId: %q", *v.DisplayName, name, *v.ObjectID)
+					groupObj = &v
+					break
+				} else {
+					log.Printf("[DEBUG] [data_source_azuread_group] %q (API result) does not match %q (given value)", *v.DisplayName, name)
```
tombuildsstuff: _minor_ we can ditch the `[data_source_azuread_group]` since `data_azuread_group.foo` will be output as the prefix AFAIK

```suggestion
					log.Printf("[DEBUG] %q (API result) does not match %q (given value)", *v.DisplayName, name)
```

azuread/data_source_group.go
```diff
@@ -0,0 +1,98 @@
...
+
+		resp, err := client.ListComplete(ctx, filter)
+		if err != nil {
+			return fmt.Errorf("Error listing Azure AD groups: %+v", err)
+		}
+
+		for _, v := range *resp.Response().Value {
+			if v.DisplayName != nil {
+				if strings.EqualFold(*v.DisplayName, name) {
+					log.Printf("[DEBUG] [data_source_azuread_group] %q (API result) matches %q (given value). The group has the objectId: %q", *v.DisplayName, name, *v.ObjectID)
```
tombuildsstuff: _minor_ we can ditch the `[data_source_azuread_group]` since `data_azuread_group.foo` will be output as the prefix AFAIK

```suggestion
					log.Printf("[DEBUG] %q (API result) matches %q (given value). The group has the objectId: %q", *v.DisplayName, name, *v.ObjectID)
```

azuread/data_source_group.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		adgroup = resp
+
+	} else {
+
+		// use the name to find the Azure AD group
+		name := d.Get("name").(string)
+		filter := fmt.Sprintf("displayName eq '%s'", name)
+		log.Printf("[DEBUG] [data_source_azuread_group] Using filter %q", filter)
```
tombuildsstuff: _minor_ we can ditch the `[data_source_azuread_group]` since `data_azuread_group.foo` will be output as the prefix AFAIK

```suggestion
		log.Printf("[DEBUG] Using filter %q", filter)
```

website/docs/d/group.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+The following arguments are supported:
+
+* `object_id` - (Optional) The ID of the Azure AD Group we want to lookup.
+
+* `name` - (Optional) The ID of the Azure AD Group we want to loopup.
+
+-> **NOTE:** At least one of `name` or `object_id` must be specified.
+
+-> **WARNING:** `name` is not unique within Azure Active Directory. The data source will only return the first Group found.
```
tombuildsstuff: since this is a warning we can use `~>` rather than `->` (which gives us a yellow box rather than a blue info one)

```suggestion
~> **WARNING:** `name` is not unique within Azure Active Directory. The data source will only return the first Group found.
```

website/docs/r/group.markdown
```diff
@@ -0,0 +1,53 @@
...
+Manages a Group within Azure Active Directory.
+
+-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to `Read and write all groups` within the `Windows Azure Active Directory` API.
+
+-> **NOTE:** Additionally, due to a limitation within the API, you have to assign **one** of the following Azure Active Directory Roles to the Service Principal to be able to delete Groups:
+
+* User Account Administrator
+* Company Administrator
+
+You can assign one of the required Azure Active Directory Roles with PowerShell. Please refer to [this documentation](https://docs.microsoft.com/en-us/powershell/module/azuread/add-azureaddirectoryrolemember) for more details.
```
tombuildsstuff: I think it'd be better to be platform agnostic and reference the PowerShell method of doing this here: https://docs.microsoft.com/en-us/cli/azure/role/assignment?view=azure-cli-latest#az-role-assignment-create - WDYT?

website/docs/d/group.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+data "azuread_group" "test_group" {
+  name = "MyTestGroup"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `object_id` - (Optional) The ID of the Azure AD Group we want to lookup.
```
tombuildsstuff:
```suggestion
* `object_id` - (Optional) The UUID of the Azure AD Group
```

tombuildsstuff: thinking about this a little - what's the use-case for looking up an AzureAD group by its ID rather than it's name? I can't think the name as the other available output (at this time) would be particularly useful when you've got the ID?

website/docs/r/group.markdown
tiwood: Please note that you need to assign Azure AD directory roles to the Service Principal, that is not the same as an Azure Role Assignment. This assignment is currently only possible via the PowerShell Module "AzureAD" (available on Windows PowerShell or in the Azure Cloud Shell) or directly through the Graph API.

I'm going to add some more details to the docs to make this more clear.

website/docs/d/group.html.markdown
tiwood: The group display names are not unique within AAD. There may be cases where multiple groups have the same name.

Example: 
- User syncs AD forest foo.mydomain.com to Azure AD
- The forest has multiple sub domains, say bar1 and bar2
- Each domain has a domain users group "employees"
- This results in multiple Azure AD groups called "employees" where the only way to target these is the object id.

This is a bad configuration ofc. but it may happen, therefore I thought i might as well give the option to scope by id. Ofc. we don't have to support each (bad) scenario.

So @tombuildsstuff, are we going to keep the option? :-)

azuread/data_source_group.go
tiwood: This mirrors the behaviour of the API, as the DisplayName is not case-sensitive for this property, but I may be better to follow stay consistent on the Terraform side of things.

azuread/data_source_group.go
tiwood: made requested changes.

azuread/resource_group.go
tiwood: made requested changes.

azuread/data_source_group.go
tiwood: made requested changes.

azuread/data_source_group.go
tiwood: made requested changes.

azuread/data_source_group.go
tiwood: made requested changes.

website/docs/d/group.html.markdown
tiwood: made requested changes.

tiwood:
@tombuildsstuff, this is ready for another review.
(I think I actually have to add a new comment to remove the waiting-response tag)

website/docs/d/group.html.markdown
tombuildsstuff: I was thinking more along the lines of the outputs available, for example if you've only got the group name (and there's a single group matching that name) then the ID is useful as an output; however if you've got the ID I'm not sure of the use-case for what you'd do with the name?

tiwood: Made requested changes in 2ebf0cd

azuread/data_source_group.go
tiwood: made requested changes

website/docs/d/group.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+data "azuread_group" "test_group" {
+  name = "MyTestGroup"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The UUID of the Azure AD Group we want to lookup.
```
tom-dudley: I think you mean name rather than UUID

tombuildsstuff:
@tiwood sorry for the delay - been trying to get the permissions sorted for the service principal - will take a look at that again today

tiwood:
> @tiwood sorry for the delay - been trying to get the permissions sorted for the service principal - will take a look at that again today

No worries, if I can help let me know.

For reference, this is how I do it in cloud shell:

```powershell 
$aadUserAdminRole = Get-AzureADDirectoryRole | ? DisplayName -eq "User Account Administrator"
$terraformServicePrincipal = Get-AzureADServicePrincipal | ? ObjectId -eq "_YourTerraformServicePrincipalObjectId_"
Add-AzureADDirectoryRoleMember -ObjectId $aadUserAdminRole.ObjectId -RefObjectId $terraformServicePrincipal.ObjectId
```

website/docs/r/group.markdown
```diff
@@ -0,0 +1,53 @@
...
+# azuread_group
+
+Manages a Group within Azure Active Directory.
+
+-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to `Read and write all groups` within the `Windows Azure Active Directory` API.
+
+-> **NOTE:** Additionally, due to a limitation within the API, you have to assign **one** of the following Azure Active Directory Roles to the Service Principal to be able to delete Groups:
+
+* User Account Administrator
+* Company Administrator
```
tombuildsstuff: after spending some time looking into this, it appears that `Company Administrator` should be the only role necessary - the `User Account Administrator` doesn't appear to have sufficient rights on it's own (I'm assuming it may for the `azuread_user` resource however?) - as such could we update this?

website/docs/r/group.markdown
```diff
@@ -0,0 +1,53 @@
...
+
+You can assign one of the required Azure Active Directory Roles with the **AzureAD PowerShell Module**, which is available for Windows PowerShell or in the Azure Cloud Shell. Please refer to [this documentation](https://docs.microsoft.com/en-us/powershell/module/azuread/add-azureaddirectoryrolemember) for more details.
```
tombuildsstuff: I initially tried to do this in CloudShell but couldn't get this working (& ended up having to spin up a Windows VM to grant these permissions in the end); whilst it's likely I did something wrong, is this definitely possible in CloudShell?

tombuildsstuff:
@tiwood thanks for the help - got the permissions working in the end - and the tests now run for me:

```
$ acctests azuread TestAccAzureADGroup_
=== RUN   TestAccAzureADGroup_basic
--- PASS: TestAccAzureADGroup_basic (4.89s)
=== RUN   TestAccAzureADGroup_complete
--- PASS: TestAccAzureADGroup_complete (4.57s)
PASS
ok  	github.com/terraform-providers/terraform-provider-azuread/azuread	9.605s
```

website/docs/r/group.markdown
tiwood: `User Account Administrator` should definitely work, see https://docs.microsoft.com/en-us/azure/active-directory/users-groups-roles/directory-assign-admin-roles#user-account-administrator.

I use these permissions exclusively on my Service Principal and it seems to work.

I've found, that if you apply a role to the Service principal, it may take a few min to propagate.

tombuildsstuff:
@tiwood thanks for this PR - this now LGTM 👍

ksix:
@tiwood the docs here don't specify the option to use `object_id` or group `name`, is that still part of this feature?

tiwood:
> @tiwood the docs here don't specify the option to use `object_id` or group `name` (only `name`), is that still part of this feature?

Hi @ksix, can you explain further what you need?
Originally it was possible to use the data source with the object_id parameter, but after some discussions we decided to remove it for now. Do you have a use case for this?

erick-thompson:
Do you have any idea when 0.2.0 might be released? I'd love to start using the group resource instead of littering object_id all over the place.

ksix:
@tiwood I want to create groups in one  script and assign them in another (asynchronously), coupling to the name is unsafe because it can be changed by others or may not even return the right group.

PR #18, opened 2019-01-17
New resource & data source 'azuread_user'

tiwood:
This PR adds a new resource and data source for Azure Active Directory users.

- [x] New data source 'azuread_user'
- [x] New resource 'azuread_user'
- [x] Add tests

## Example Usage

```hcl

// This creates a new Azure AD user
resource "azuread_user" "my_user" {
 user_principal_name = "john.doe@hashicorptest.onmicrosoft.com"
 display_name = "John Doe"
 mail_nickname = "John Doe"
 account_enabled = true
 password = "mySecretPw$$"
 force_password_change = false
}

// This queries the a user
data "azuread_user" "my_user" {
  user_principal_name = "john.doe@hashicorptest.onmicrosoft.com"
}
```

## Use cases

- Assign a role to a user principal
- Create a "service user" where it is not possible to use service principals (example: Databricks)
- Create fallback admin accounts (onmicrosoft)


tiwood:
@katbyte, @tombuildsstuff, how are we going to implement tests which require specific environment configuration? 

Like in this example the Domain Suffix for the user principal name has to be a verified, managed domain (like foo.onmicrosoft.com).

azuread/resource_user.go
```diff
@@ -0,0 +1,188 @@
...
+					Schema: map[string]*schema.Schema{
+						"password": {
+							Type:      schema.TypeString,
+							Required:  true,
+							Sensitive: true,
+						},
+						"force_password_change": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
```
tombuildsstuff: since it's only possible to configure a single password (profile) for a given user - I think it'd be worth moving these fields to the top-level (which means we can ignore setting them in the state / the Set/List requirement here) - WDYT?

tombuildsstuff:
@tiwood thanks for this PR :)

> @katbyte, @tombuildsstuff, how are we going to implement tests which require specific environment configuration?
> Like in this example the Domain Suffix for the user principal name has to be a verified, managed domain (like foo.onmicrosoft.com).

Ideally we'd look the AAD Domain up and have that available as a data source (in a similar way we do for [the `azurerm_client_config` data source](https://www.terraform.io/docs/providers/azurerm/d/client_config.html) - maybe `azuread_current_domain`?).

However in the short-term we should be able to make this an Environment Variable and then use that in the test config via `os.Getenv("ARM_TENANT_DOMAIN")` (we do this in AzureRM for fields like the Client ID/Secret in the Kubernetes tests) - would that approach work here?

I guess the other question is whether we should just accept a username and prepend the domain to it (which would allow us to validate the username matches any requirements)?

tiwood:
> Ideally we'd look the AAD Domain up and have that available as a data source (in a similar way we do for [the `azurerm_client_config` data source](https://www.terraform.io/docs/providers/azurerm/d/client_config.html) - maybe `azuread_current_domain`?).

This makes perfect sense and as I'm going to need something like this for one of my use cases anyway, I've created a PR to add the functionality (#27)

> I guess the other question is whether we should just accept a username and prepend the domain to it (which would allow us to validate the username matches any requirements)?

I'm pretty sure there are a lot of use cases where you want to specify the target domain. I have customers that have multiple domains and want to provision service users (not service principals) in these specific domains.

azuread/resource_user.go
tiwood: Thanks for the review.

That makes sense, I'm going to change it.

azuread/resource_user.go
```diff
@@ -0,0 +1,176 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"user_principal_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
katbyte: Could we get some validation here?
```suggestion
				ValidateFunc: validate.NoEmptyStrings,
			},
```

azuread/resource_user.go
```diff
@@ -0,0 +1,176 @@
...
+
+			"display_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
katbyte: Could we get some validation here?
```suggestion
				ValidateFunc: validate.NoEmptyStrings,
			},
```

azuread/resource_user.go
```diff
@@ -0,0 +1,176 @@
...
+
+			"mail_nickname": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
katbyte: Could we get some validation here?
```suggestion
				ValidateFunc: validate.NoEmptyStrings,
			},
```

azuread/resource_user.go
```diff
@@ -0,0 +1,176 @@
...
+			"account_enabled": {
+				Type:     schema.TypeBool,
+				Required: true,
+			},
+
+			"password": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
```
katbyte: Could we get some validation here?
```suggestion
				ValidateFunc: validate.NoEmptyStrings,
			},
```

azuread/resource_user.go
```diff
@@ -0,0 +1,176 @@
...
+	passwordProfile := graphrbac.PasswordProfile{
+		ForceChangePasswordNextLogin: &forcePasswordChange,
+		Password:                     &password,
+	}
+
+	userCreateParameters := graphrbac.UserCreateParameters{
+		AccountEnabled:    &accountEnabled,
+		DisplayName:       &displayName,
+		MailNickname:      &mailNickName,
+		PasswordProfile:   &passwordProfile,
```
katbyte: _minor_
```suggestion
		PasswordProfile:   graphrbac.PasswordProfile{
			ForceChangePasswordNextLogin: &forcePasswordChange,
			Password:                     &password,
		},
```

azuread/resource_user_test.go
```diff
@@ -0,0 +1,138 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckADApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccADUser_basic(id),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckADUserExists(resourceName),
+					resource.TestCheckResourceAttrSet(resourceName, "user_principal_name"),
+					resource.TestCheckResourceAttr(resourceName, "display_name", fmt.Sprintf("acctestt%s", id)),
```
katbyte: Is there a reason to use `acctestt` over `acctest` 🤔

azuread/resource_user_test.go
```diff
@@ -0,0 +1,138 @@
...
+	return fmt.Sprintf(`
+
+data "azuread_domains" "tenant_domain" {
+	tenant_domain_only = true
+}
+
+resource "azuread_user" "test" {
+	user_principal_name = "acctest%s@${data.azuread_domains.tenant_domain.domains.0.domain_name}"
+	display_name = "acctest%s"
+	mail_nickname = "acctest%s"
```
katbyte: Could we align these assignments?

azuread/data_user.go
```diff
@@ -0,0 +1,59 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validate.NoEmptyStrings,
+			},
+
+			"mail": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
```
katbyte: Could we add `display_name` and `mail_nickname` (if different then mail) here?

MovieStoreGuy:
Yes, I am 100% for this PR

tiwood:
@katbyte, this is ready for another review.

The tests will fail until PR #27 is merged, as this uses the new domains data source to find the target domain for the user creation.

Thanks.

azuread/data_user.go
katbyte: Given that they are exposed by the resource, we should likely have them here. Generally we have parity between the two.

azuread/data_user.go
```diff
@@ -0,0 +1,59 @@
...
+	log.Printf("[DEBUG] Using Get with the following query string: %q", queryString)
+	resp, err := client.Get(ctx, queryString)
+	if err != nil {
+		if ar.ResponseWasNotFound(resp.Response) {
+			return fmt.Errorf("Error: No AzureAD User found with the following query string: %q", queryString)
+		}
+		return fmt.Errorf("Error making Read request on AzureAD User the following query string: %q: %+v", queryString, err)
+	}
+
+	user = resp
```
katbyte: Could we either just use `resp` or rename `resp` to `user`?

azuread/data_user.go
```diff
@@ -0,0 +1,59 @@
...
+
+	d.SetId(*user.ObjectID)
```
katbyte: We will need to nil check this, see my pr #32

azuread/helpers/validate/strings.go
```diff
@@ -18,3 +19,18 @@ func NoEmptyStrings(i interface{}, k string) ([]string, []error) {
 
 	return nil, nil
 }
+
+// StringIsEmailAddress validates that the given string is a valid email address (foo@bar.com)
+func StringIsEmailAddress(i interface{}, k string) ([]string, []error) {
```
katbyte: Could we get a test for this validation?

azuread/resource_user.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	objectId := user.ObjectID
+
+	resp, err := client.Get(ctx, *objectId)
+	if err != nil {
+		return fmt.Errorf("Error retrieving User (%q) with ObjectID %q: %+v", userPrincipalName, *objectId, err)
+	}
+
+	d.SetId(*resp.ObjectID)
```
katbyte: As above, we should nil check this dereference

azuread/resource_user_test.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("Azure AD User still exists:\n%#v", resp)
+	}
+
+	return nil
+}
+
+func testAccADUser_basic(id string, password string) string {
+	return fmt.Sprintf(`
+
+#data "azuread_domains" "tenant_domain" {
```
katbyte: Should this be uncommented?

azuread/resource_user_test.go
```diff
@@ -0,0 +1,145 @@
...
+#	only_initial = true
+#}
+
+resource "azuread_user" "test" {
+	user_principal_name = "acctest%s@${data.azuread_domains.tenant_domain.domains.0.domain_name}"
```
katbyte: Could we align these assignments?

website/docs/r/user.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+
+Manages a User within Azure Active Directory.
+
+-> **NOTE:** If you're authenticating using a Service Principal then it must have permissions to `Directory.ReadWrite.All` within the `Windows Azure Active Directory` API.
+
+## Example Usage
+
+```hcl
+resource "azuread_user" "test_user" {
+  user_principal_name = "john@hashicorp.com"
```
katbyte: Could we align these assignments?

website/docs/r/user.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `user_principal_name` - (Required) The User Principal Name of the Azure AD User.
+* `display_name` - (Required) The name to display in the address book for the user.
+* `mail_nickname`- (Required) The mail alias for the user.
+* `account_enabled` - (Required) `true` if the account should be enabled, otherwise `false`.
+* `password` - (Required) The password for the User. The password must satisfy minimum requirements as specified by the password policy. The maximum length is 16 characters.
+* `force_password_change` - (Required) `true` if the User is forced to change the password during the next sign-in.
```
katbyte: Could we specify the default here?

azuread/data_user.go
tiwood: Thanks for the feedback. 
Code updated accordingly.

azuread/data_user.go
tiwood: Updated as requested.

azuread/data_user.go
tiwood: Updated as requested.

azuread/helpers/validate/strings.go
tiwood: Test added.

azuread/resource_user.go
tiwood: Updated as requested.

azuread/resource_user_test.go
tiwood: Ops. :-)

tiwood:
Hi @katbyte, thank you for the review. I've made the requested changes.

azuread/data_user_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"math/rand"
+	"strconv"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccDataSourceAzureADUser_byUserPrincipalName(t *testing.T) {
+	dataSourceName := "data.azuread_user.test"
+	id := strconv.Itoa(rand.Intn(9999))
```
katbyte: Could we change this to: 
```suggestion
	id := acctest.RandString(7)
```

azuread/resource_user_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/ar"
+)
+
+func TestAccAzureADUser_basic(t *testing.T) {
+	resourceName := "azuread_user.test"
+	id := strconv.Itoa(rand.Intn(9999))
```
katbyte: Could we change this to: 
```suggestion
	id := acctest.RandString(7)
```

azuread/resource_user_test.go
```diff
@@ -0,0 +1,145 @@
...
+					"password", // not returned from API, sensitive
+				},
+			},
+		},
+	})
+}
+
+func TestAccAzureADUser_update(t *testing.T) {
+	resourceName := "azuread_user.test"
+	id := strconv.Itoa(rand.Intn(9999))
```
katbyte: Could we change this to: 
```suggestion
	id := acctest.RandString(7)
```

azuread/resource_user_test.go
```diff
@@ -0,0 +1,145 @@
...
+	password := id + "p@$$wR2"
+	updatedId := strconv.Itoa(rand.Intn(9999))
```
katbyte: Could we change this to: 
```suggestion
	id := acctest.RandString(7)
```

lukasmrtvy:
Is possible to implement also support for 'phone number' ? I dont know if its even possible. Thanks

PR #26, opened 2019-01-22
Switching to Go Modules

tombuildsstuff:


tombuildsstuff:
@alexsomesan tests seem fine (they're eventually consistent because of an open bug, but the failures aren't unexpected)

PR #27, opened 2019-01-22
New data source 'azuread_domains'

tiwood:
This PR adds a new data source for Domains registered in Azure AD.

- [x] New data source 'azuread_domains'
- [x] Add tests

## Example Usage

```hcl

// This gets all domains currently registered in Azure AD
data "azuread_domains" "my_domains" {}

// This will return only the tenant root domain (foo.onmicrosoft.com)
data "azuread_domains" "my_root_domain" {
 tenant_domain_only = true
}

```

## Example Output
```json
domains = [
    {
        domain_name = hashicorp.onmicrosoft.com,
        is_default = 1,
        is_initial = 1,
        is_verified = 1
    }
]
```

## Use cases

* This will be helpful to create objects where the user principal name is required and not known during runtime (testing for example)

### DIFF
--- a/azuread/config.go
+++ b/azuread/config.go
@@ -29,6 +29,7 @@ type ArmClient struct {
 
 	// azure AD clients
 	applicationsClient      graphrbac.ApplicationsClient
+	domainsClient           graphrbac.DomainsClient
 	groupsClient            graphrbac.GroupsClient
 	servicePrincipalsClient graphrbac.ServicePrincipalsClient
 }
@@ -73,6 +74,9 @@ func getArmClient(authCfg *authentication.Config) (*ArmClient, error) {
 func (c *ArmClient) registerGraphRBACClients(endpoint, tenantID string, authorizer autorest.Authorizer) {
 	c.applicationsClient = graphrbac.NewApplicationsClientWithBaseURI(endpoint, tenantID)
 	configureClient(&c.applicationsClient.Client, authorizer)
+
+	c.domainsClient = graphrbac.NewDomainsClientWithBaseURI(endpoint, tenantID)
+	configureClient(&c.domainsClient.Client, authorizer)
 
 	c.groupsClient = graphrbac.NewGroupsClientWithBaseURI(endpoint, tenantID)
 	configureClient(&c.groupsClient.Client, authorizer)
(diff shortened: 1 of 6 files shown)


tiwood:
> hey @tiwood
> 
> Thanks for this PR :)
> 
> I've spent a little bit of time playing with this and on the whole this looks pretty good - however I'm wondering if we want to expose a couple of extra filters here? I've [put together this gist showing what I mean](https://gist.github.com/tombuildsstuff/4abf54d3cf4d831dabcc77fc1c5b58ed) - what do you think?
> 
> Thanks

Hey @tombuildsstuff,

I'm still thinking  about what the default output and the parameters should be, as I'm not 100% sold on **only** filters, as each domain can have multiple of the "only properties" at once.

What do you think about the following:

* azuread_domains without params will only return verified domains
* There will be an option to `include_unverified` (Conflicts with `only_tenant_domain`)
* There will be an option to `only_tenant_domain` (Conflicts with `include_unverified`)
* The following properties will be included in the output for each domain:
`is_verified` `is_default` `authentication_type`

I think this is easier to understand, WDYT?

tombuildsstuff:
@tiwood `include` sounds good 👍 I was otherwise going to suggest `filter` (but it's the same thought)

The only question I've got is around the `tenant_domain` field, I feel like it'd be better to consistently make this `is_initial` - WDYT?

tiwood:
@tombuildsstuff, I've changed the resource like I proposed and updated the original description accordingly.

For consistency I've renamed the attribute `only_tenant_domain` to `only_initial`.

Again a short summary:
* without attributes the data source will return all **VERIFIED** domains registered in AAD.
* if `only_initial` is set to `true` only the initial tenant domain is returned (the foo.onmicrosoft.com domain)
* if you set `include_unverified` to `true` unverified domains will be included in the output.

katbyte:
Thanks @tiwood,

I think you may want to mark the `only_initial` as conflicting with `include_unverified`.

Also `only_default` may be another useful filter to add?

tiwood:
@katbyte, they actually may be use cases where you want to create objects in the default domain, therefore I've added the option to set `only_default`.

Additionally I've added the ConflictsWith validation:
* `include_unverified` conflicts with `only_default` and `only_initial` as default and initial domains have to be verified by design.
* `only_default`and `only_initial`are conflicting with each other.

PR #37, opened 2019-02-06
fix build: update google go get cookie & switch to golangci for linting

katbyte:


alexsomesan:
Is this build supposed to run with go modules or “classic” (from /vendor)?
I suspect the failure in Travis is because GOPATH is set.

PR #51, opened 2019-02-27
Allow settings group_membership_claims on ad application

jamielennox:
Set the groupMembershipClaims into AdditionalProperties[1]. This is an
unvalidated map that can be used to set arbitrary manifest entries.

This is kind of new for terraform however in this case the inputs are
well documented[2] and can be validated.

This imports the suppress case difference in strings helper from the
azuread provider.

[1] https://godoc.org/github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac#ApplicationCreateParameters
[2] https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-app-manifest

### DIFF
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/suppress"
 	"github.com/terraform-providers/terraform-provider-azuread/azuread/helpers/tf"
 
 	"github.com/hashicorp/terraform/helper/validation"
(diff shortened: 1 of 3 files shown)


jamielennox:
So i'm interested in this kind of as a test case of well defined parameters that exist on manifest but aren't explicitly in the SDK. I'm interested in this variable in particular because it's required as part of: https://docs.microsoft.com/en-us/azure/aks/aad-integration which is what I want here. 

If this is acceptable the next bit about exposing oauth2 permissions might require a bit more discussion.

azuread/resource_application.go
```diff
@@ -75,6 +76,16 @@ func resourceApplication() *schema.Resource {
 				Computed: true,
 			},
 
+			"group_membership_claims": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: suppress.CaseDifference,
+				ValidateFunc: validation.StringInSlice(
+					[]string{"None", "SecurityGroup", "All"},
+					true,
```
katbyte: Could we not ignore case here? We are slowly moving to treating as much as case sensitive as possible to get ahead of weird casing bugs we occasionally run into.

janschumann:
@katbyte I added tests, docs, etc. via #78

katbyte:
Does this mean i can close #51 in favour of #78?

jamielennox:
Sorry, I had planned to come back and finish this. But yes, thank you @janschuman. I'm happy to use that PR instead.

katbyte:
👍 closing in favour of #78

PR #52, opened 2019-02-27
azuread_group: set mail nickname to a guid by default

### LINKED ISSUE #48: Group Resource fails if group name contains spaces
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request

Terraform v0.11.11
+ provider.azuread (unversioned) master/83d4ab7

### Affected Resource(s)

* `azuread_group`

```hcl
resource "azuread_group" "test" {
  name = "Test Security Group"
}
```

### Exception Output

Invalid value specified for property 'mailNickname' of resource 'Group'.
{"item":"PropertyName","value":"mailNickname"},
{"item":"PropertyErrorCode","value":"InvalidValue"}

### Expected Behaviour

I would expect to be allowed to create groups with spaces in the name, but currently the provider just passes the exact same value as group name to the mailNickname field, which doesn't support spaces:

```go
properties := graphrbac.GroupCreateParameters{
	DisplayName:     &name,
	MailEnabled:     p.Bool(false), //we're defaulting to false, as the API currently only supports the creation of non-mail enabled security groups.
	MailNickname:    &name,
	SecurityEnabled: p.Bool(true), //we're defaulting to true, as the API currently only supports the creation of non-mail enabled security groups.
}
```

### Actual Behaviour

Failed to create group because having spaces in the mailNickname is unsupported.

### Steps to Reproduce

1. `terraform apply`

### Suggested Fix

Can we just set the MailNickname to a generated GUID instead of the `name` field? That is what the Azure Portal does if creating a non-MailEnabled group from the UI.

katbyte:
fixes #48


katbyte:
As groups has not been released yet this isn't a breaking change.

PR #61, opened 2019-03-13
Ignore order for reply_urls

cgroschupp:
Use TypeSet for reply_urls to ignore order.


cgroschupp:
@katbyte I haven't found a solution except to set reply_url to a fixed string to get a static set ID.

Issue #36, opened 2019-02-05
azuread_group: Support owners and members management 


perbergland:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azuread_group

### Potential Terraform Configuration

```hcl
resource "azuread_group" "mygroup" {
name= "mygroup"
members = ["${azuread_member.member1.id}","guid-guid-guid-guid"]
owners = ["${azuread_member.member1.id}","guid-guid-guid-guid"]
}

# or maybe this is better to allow a mixture of managed and unmanaged owners/members

resource "azuread_group_owner" "managed_owner" {
group = "${azuread_group.mygroup.id}"
owner = "${azuread_member.member1.id}"
}

resource "azuread_group_member" "managed_member" {
group = "${azuread_group.mygroup.id}"
member = "${azuread_member.member1.id}"
}
```

ksix:
The model of mixed managed and unmanaged owners & members is what I need. It also gives the most flexibility.

tom-dudley:
I'm interested in seeing this implemented. I've done a bit of work [here](https://github.com/tom-dudley/terraform-provider-azuread/tree/ds_r_azuread_group_membership) which implements group membership (not ownership yet) using UPNs:

```
resource "azuread_group" "mygroup" {
    name = "MyGroup"
    members = ["foo.bar@example.com"]
 }
```

It's still a WIP and lacking tests, but does currently work for membership.

Given my use case involves creating new groups and adding pre-existing users, the UPNs seem more maintainable than guids, although I can see the following may be a better option:

```
data "azuread_user" "foo_bar" {
    user_principal_name = "foo.bar@example.com"
 }

resource "azuread_group" "mygroup" {
    name = "MyGroup"
    members = ["${azuread_user.foo_bar.id}"]
 }
```

The only downside I see here is having to have a data element for each user, although the loops in `v 0.12` may solve that.

Couple of questions:
* Was the `azuread_member` meant to be `azuread_user`, or are you suggesting a new resource?
[...]

PR #62, opened 2019-03-13
azuread_group - support for the `owners` property

tiwood:
(1/2 of #36)

This adds a new resource to manage Azure AD Group Owners with Terraform. 
Additionally this adds support for `owners` in the resource azuread_group.

- [x] New resource 'azuread_group_owner'
- [x] New attribute 'owners' for resource 'azuread_group'
- [x] Add tests
- [x] Add documentation
- [ ] azuread_group: If `owners` is not set, it should leave existing owners as is.

## Example Usage

```hcl
resource "azuread_group" "my_group" {
  name = "MyGroup"
}

data "azuread_user" "my_user" {
  user_principal_name = "johndoe@hashicorp.com"
}

resource "azuread_group_owner" "default_owner" {
  group_object_id = "${azuread_group.my_group.id}"
  owner_object_id = "${data.azuread_user.my_user.id}"
}
```

## Example for azuread_group

```hcl
resource "azuread_group" "my_group" {
  name     = "my group"
  owners  = ["9c9b3992-78ac-11e9-a3ef-a31d2f3e9d2c"] 
}
```

## Argument Reference

* `group_object_id` - (Required) The object id of the Azure AD Group where the Owner should be added.
* `owner_object_id` - (Required) The object id of the Azure AD User you want to add as Owner.

## Caveats

Azure requires at least one owner per Azure AD group. That means the destruction of 'azuread_group_owner' resources will fail if no additional owner is present on the group.


tiwood:
Related: #36

tiwood:
@katbyte, @tombuildsstuff, do you guys have any ideas how we could tackle the issue of the 'not deletable' (see known issues above) group owners?

BTW you have the same behaviour you can see in the Azure Portal.
1. Create a group
2. Add a new owner
3 Try to remove the owner (this fails if its the last owner of the group)

Should we leave this up to the User?

twendt:
Would it be an option to not implement it as a separate resource but instead integrate it into the group resource? There you could always pass the full list of owners in and not allow the list to be empty.

tiwood:
@twendt, I think this is the right way to do it. I'm going to look into it.
Thanks.

katbyte:
@tiwood & @twendt,

There are reasons to have it as a separate resource (you can separate out creating & managing the owners) & within the resource (central location, ability to explicitly define who the owners should be)

And as such we should support both.

as to the issue of not being able to remove the last user, I say we fail destruction/update and leave that up to the user as theres not much we can do about it.

tiwood:
I will look into this in the next days.

azuread/resource_group_owner.go
```diff
@@ -0,0 +1,131 @@
...
+	tenantID := client.TenantID
+
+	ownerGraphURL := fmt.Sprintf("https://graph.windows.net/%s/directoryObjects/%s", tenantID, ownerID)
+
+	properties := graphrbac.AddOwnerParameters{
+		URL: &ownerGraphURL,
+	}
+
+	_, err := client.AddOwner(ctx, groupID, properties)
+	if err != nil {
```
katbyte: _minor_ these two lines could be merged:
```
	if _, err := client.AddOwner(ctx, groupID, properties); err != nil {
```

azuread/resource_group_owner.go
```diff
@@ -0,0 +1,131 @@
...
+		user, _ := owners.Value().AsUser()
+		if user != nil {
+			if *user.ObjectID == ownerID {
+				ownerObjectID = *user.ObjectID
+				break
+			}
+		}
+
+		err = owners.NextWithContext(ctx)
+		if err != nil {
```
katbyte: These two lines could be merged

website/docs/r/group_owner.markdown
```diff
@@ -0,0 +1,52 @@
...
+  group_object_id = "${azuread_group.my_group.id}"
+  owner_object_id = "${data.azuread_user.my_user.id}"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `group_object_id` - (Required) The object id of the Azure AD Group where the Owner should be added.
```
katbyte: Could we mark this as force new
```suggestion
* `group_object_id` - (Required) The object id of the Azure AD Group where the Owner should be added. Changing this forces a new resource to be created.
```

website/docs/r/group_owner.markdown
```diff
@@ -0,0 +1,52 @@
...
+* `owner_object_id` - (Required) The object id of the Azure AD User you want to add as Owner.
```
katbyte: and here
```suggestion
* `owner_object_id` - (Required) The object id of the Azure AD User you want to add as Owner. Changing this forces a new resource to be created.
```

tiwood:
I'm almost done with the changes, but I'm currently trying to figure out something.

Is it possible to suppress a diff if the ordering of a list has changed?

```
testing.go:538: Step 0 error: After applying this step, the plan was not empty:

        DIFF:

        UPDATE: azuread_group.test
          owners.0: "0911ca82-5526-4d57-bfd6-88cb71057df4" => "dd447a48-663e-4b2a-9d0a-db9db3287d72"
          owners.1: "dd447a48-663e-4b2a-9d0a-db9db3287d72" => "0911ca82-5526-4d57-bfd6-88cb71057df4"
```

The schema property is defined as follows:
```
Schema: map[string]*schema.Schema{
			"name": {
				Type:         schema.TypeString,
				Required:     true,
				ForceNew:     true,
				ValidateFunc: validation.NoZeroValues,
			},
			"owners": {
				Type:     schema.TypeList,
				Optional: true,
				MinItems: 1,
				MaxItems: 100, //Group owners are maxed out at 100
				Elem: &schema.Schema{
					Type:         schema.TypeString,
					ValidateFunc: validate.UUID,
				},
			},
		},
```

website/docs/r/group_owner.markdown
tiwood: Changed as requested.

website/docs/r/group_owner.markdown
tiwood: Changed as requested.

azuread/resource_group_owner.go
tiwood: Changed as requested.

azuread/resource_group_owner.go
tiwood: Changed as requested.

azuread/resource_group.go
```diff
@@ -83,11 +93,22 @@ func resourceGroupCreate(d *schema.ResourceData, meta interface{}) error {
 	if v, ok := d.GetOk("members"); ok {
 		members := tf.ExpandStringSlicePtr(v.(*schema.Set).List())
 
+		// we could lock here against the group ember resource, but the should not be used together (todo conflicts with at a resource level?)
```
mbfrahry:
```suggestion
		// we could lock here against the group member resource, but they should not be used together (todo conflicts with at a resource level?)
```

azuread/resource_group.go
```diff
@@ -83,11 +93,22 @@ func resourceGroupCreate(d *schema.ResourceData, meta interface{}) error {
...
 		if err := graph.GroupAddMembers(client, ctx, *group.ObjectID, *members); err != nil {
 			return err
 		}
 	}
 
+	// Add owners if specified
+	if v, ok := d.GetOk("owners"); ok {
+		members := tf.ExpandStringSlicePtr(v.(*schema.Set).List())
+
+		// we could lock here against the group owner resource, but the should not be used together (todo conflicts with at a resource level?)
```
mbfrahry:
```suggestion
		// we could lock here against the group owner resource, but they should not be used together (todo conflicts with at a resource level?)
```

katbyte:
Closes #36

PR #71, opened 2019-04-03
New resource 'azuread_application_password'

twendt:
This adds a new resource to manage applications passwords. It is based on the service principal password resource.


twendt:
@katbyte I agree that it would make sense. I looked into the code to see how it can be achieved and I am not sure how to handle the different clients that are required. I was thinking about defining an interface, but the Get function returns return a concrete type of an application or service principal so I guess that won't work.
What is the best approach for this? Should I create a little wrapper type for this?

The other question is about the input. Should we use different inputs for the application id and service principal id and use that to figure out what context we are running in? Or should we just accept an object id and the resource then figures out if the id belongs to an application or service principal?

twendt:
I updated it now so that only a single object_password resource is used for applications and service principals.
I create a little wrapper type that handles the different clients.

The locking is an issue at the moment though. I need to update it to figure the resource_name out that the object_id belongs to to make sure that I lock the application or service principal instead of the object_password resource.

twendt:
I left the service_principal_password in as it would otherwise break existing installations that use it.

katbyte:
👋 @twendt,

After some internal discussion we decided to change the approach to one that is a little less general. I hope you don't mind but i have pushed some changed to this effect so we could get this into the next release. Mainly it keeps the schema and client calls separate while moving the shared graph logic and schema out into a helper.

twendt:
@katbyte Thank you very much for helping me out here.
I really like the new approach, because it is not a breaking change for users that already used the service principal password resource before and I also think that this is more intuitive for the end user.

PR #78, opened 2019-05-14
allow setting group membeship claims on aad application

janschumann:
This is based on #51 while adding tests and documentation.

As requested in #51, the value of `group_membership_claims` is not forced to be case sensitive. So that `All`, `all`, `SecurityGroup`, `securitygroup`, `None`, `none` are all accepted.


azuread/data_application.go
```diff
@@ -70,6 +73,16 @@ func dataApplication() *schema.Resource {
 				Computed: true,
 			},
 
+			"group_membership_claims": {
```
katbyte: As this is the datasource, we only need it to be computed:

```go
			"group_membership_claims": {
				Type:             schema.TypeString,
				Computed:         true,
			},
```

azuread/helpers/suppress/string.go
```diff
@@ -0,0 +1,11 @@
+package suppress
+
+import (
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func CaseDifference(_, old, new string, _ *schema.ResourceData) bool {
```
katbyte: Could we remove this?

azuread/resource_application.go
```diff
@@ -75,6 +76,16 @@ func resourceApplication() *schema.Resource {
 				Computed: true,
 			},
 
+			"group_membership_claims": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: suppress.CaseDifference,
```
katbyte: I was asking this to not be case sensitive. Unless the API is returning varying values with random casing could we be case sensitive here?

katbyte:
I hope you don't mind but i am going to push some changes to get this merged today 🙂

PR #79, opened 2019-05-14
allow reading oauth2 permissions from aad application

janschumann:
Adds the `oauth2Permissions` manifest attribute to the exported attributes of a `azuread_application` resource and data resource.

https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-app-manifest


dominik-lekse:
@janschumann Thanks for doing the "boring" stuff. I am looking forward for this being merged into the provider soon.

azuread/resource_application.go
```diff
@@ -340,3 +394,43 @@ func flattenADApplicationResourceAccess(in *[]graphrbac.ResourceAccess) []interf
...
+func flattenADApplicationOauth2Permissions(in []interface{}) []map[string]interface{} {
+	if in == nil {
+		return []map[string]interface{}{}
+	}
+
+	result := make([]map[string]interface{}, 0, len(in))
+	for _, oauth2Permissions := range in {
+		rawPermission := oauth2Permissions.(map[string]interface{})
+		permission := make(map[string]interface{})
+		if rawPermission["adminConsentDescription"] != nil {
```
katbyte: Could we changes these to:
```go
		if v := rawPermission["adminConsentDescription"]; v != nil {
			permission["admin_consent_description"] = v
```

website/docs/r/application.html.markdown
```diff
@@ -69,7 +69,7 @@ The following arguments are supported:
 
 * `oauth2_allow_implicit_flow` - (Optional) Does this Azure AD Application allow OAuth2.0 implicit flow tokens? Defaults to `false`.
 
-* `required_resource_access` - (Optional) A collection of `required_resource_access` blocks as documented below.
+* `required_resource_access` - (Optional) A collection of `required_resource_access` blocks as documented below.
```
katbyte:
```suggestion
* `required_resource_access` - (Optional) A collection of `required_resource_access` blocks as documented below.
```

website/docs/d/application.html.markdown
```diff
@@ -66,3 +67,23 @@ output "azure_ad_object_id" {
 * `id` - The unique identifier for one of the `OAuth2Permission` or `AppRole` instances that the resource application exposes. 
 
 * `type` - Specifies whether the id property references an `OAuth2Permission` or an `AppRole`.
+
+---
+
+`oauth2_permission` block exports the following:
```
katbyte: Is this block always exported or only when it is of type `OAuth2Permission`?

azuread/resource_application.go
```diff
@@ -121,6 +170,7 @@ func resourceApplicationCreate(d *schema.ResourceData, meta interface{}) error {
 	name := d.Get("name").(string)
 
 	properties := graphrbac.ApplicationCreateParameters{
+		AdditionalProperties:    make(map[string]interface{}),
```
katbyte: What is the reason for this here if we aren't setting anything? will the API not return it otherwise?

website/docs/d/application.html.markdown
janschumann: Yes it is always exported, but it might be an empty list.

azuread/resource_application.go
janschumann: Yes

azuread/resource_application.go
janschumann: This is a fragment left from #78. Will remove this here.

jfcoz:
@katbyte, is the review ok now ?

azuread/resource_application.go
```diff
@@ -75,6 +75,55 @@ func resourceApplication() *schema.Resource {
 				Computed: true,
 			},
 
+			"oauth2_permissions": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
katbyte: As this block is only exported/computed this can be removed
```suggestion
```

bluemalkin:
Hi, is there an ETA for a release with this fix please ?

dbourcet:
I am curious about the syntax to use. If I have an application like so : 
```
resource "azuread_application" "server" {
    name      = "server"
    reply_urls = ["http://server"]
```
what is the syntax to access the id of an oauth2_permissions block?
`${azuread_application.server.oauth2_permissions[0]["id"]` ?
`${azuread_application.server.oauth2_permissions["id"]` ?
I can't find how to use it.

dbourcet:
I manage to do it this way: `${lookup(azuread_application.server.oauth2_permissions[0], "id")}`.

PR #81, opened 2019-05-21
Increase maximum password length from 16 to 256

landro:
More info here:
https://techcommunity.microsoft.com/t5/Azure-Active-Directory-Identity/Removal-of-the-16-character-limit-for-passwords-in-Azure-AD/ba-p/565275


katbyte:
Thanks for this @landro, LGTM 👍

Issue #88, opened 2019-05-29
azuread_application type native reply_urls validation error


steve-hawkins:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v0.11.14
+ provider.azuread v0.3.1 (built from the following master commit [ac25838](https://github.com/terraform-providers/terraform-provider-azuread/commit/ac258382a782bf22786e8c4a3e663dc965c5584d))

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
resource "azuread_application" "native" {
  name                       = "native"
  type                       = "native"
  reply_urls                 = ["urn:ietf:wg:oauth:2.0:oob"]
  available_to_other_tenants = false
  oauth2_allow_implicit_flow = true
}
```

### Debug Output

[Terraform DEBUG output](https://gist.github.com/steve-hawkins/3bfb477e110a5a2987ffbde7e8e8e63b)

### Expected Behavior

Creates an azuread_application of type native with the following reply URL:-

"urn:ietf:wg:oauth:2.0:oob"

### Actual Behavior

`Error: azuread_application.native: "reply_urls.0" url has no host: "urn:ietf:wg:oauth:2.0:oob"`

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

This is a valid reply URL:-

```
az ad app show --id [REDACTED] | ConvertFrom-Json | Select-Object -ExpandProperty replyUrls
urn:ietf:wg:oauth:2.0:oob
```

### References

* #34

steve-hawkins:
suggest this bug is added to the [0.4.0 milestone](https://github.com/terraform-providers/terraform-provider-azuread/milestone/3) as the azuread_application type is also introduced in there

PR #89, opened 2019-05-29
removed reply_urls validation

steve-hawkins:
this should address #88 the alternative was to either add some very complicated regex or an if on the validation dependant on type

I have left a comment in the test if anyone could help me understand what is going on there and make that better that would be great

### DIFF
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -55,7 +55,8 @@ func resourceApplication() *schema.Resource {
 				Optional: true,
 				Computed: true,
 				Elem: &schema.Schema{
-					Type: schema.TypeString,
+					Type:         schema.TypeString,
+					ValidateFunc: validate.NoEmptyStrings,
 				},
 			},
 
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -55,8 +55,7 @@ func resourceApplication() *schema.Resource {
 				Optional: true,
 				Computed: true,
 				Elem: &schema.Schema{
-					Type:         schema.TypeString,
-					ValidateFunc: validate.URLIsHTTPOrHTTPS,
+					Type: schema.TypeString,
 				},
 			},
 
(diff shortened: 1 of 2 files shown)


azuread/resource_application_test.go
```diff
@@ -414,6 +444,16 @@ resource "azuread_application" "test" {
 `, id)
 }
 
+func testAccADApplication_nativeReplyUrls(id string) string {
+	return fmt.Sprintf(`
+resource "azuread_application" "test" {
+  name       = "acctest%s"
+	type       = "native"
```
katbyte: could we use spaces instead of tabs here?

azuread/resource_application_test.go
```diff
@@ -203,6 +202,37 @@ func TestAccAzureADApplication_native(t *testing.T) {
...
+		CheckDestroy: testCheckADApplicationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccADApplication_nativeReplyUrls(id),
+				Check: resource.ComposeTestCheckFunc(
+					testCheckADApplicationExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "name", fmt.Sprintf("acctest%s", id)),
+					resource.TestCheckResourceAttr(resourceName, "type", "native"),
+					resource.TestCheckResourceAttr(resourceName, "reply_urls.#", "1"),
+					// I didn't understand the reason reply_urls.3714513888 was used in TestAccAzureADApplication_update
```
katbyte: when the type is a set u need to get the set "hash value". I believe this can be grabbed from the state/plan when running locally and then put in here as it will always be the same for a given value.

azuread/resource_application.go
```diff
@@ -55,8 +55,7 @@ func resourceApplication() *schema.Resource {
 				Optional: true,
 				Computed: true,
 				Elem: &schema.Schema{
-					Type:         schema.TypeString,
-					ValidateFunc: validate.URLIsHTTPOrHTTPS,
```
katbyte: Could we use `validate.NoEmptyStrings` here?

azuread/resource_application_test.go
steve-hawkins: gets me every time, vscode swaps out spaces for tabs in .go files by default

steve-hawkins:
thanks for merging this one

PR #90, opened 2019-05-29
data.azuread_[group|user]: add lookup by object_id

katbyte:
This allows looking up by object_id.

As well fixes a bug where id there are multiple groups or users found for a given name it will now error instead of only returning the first.

### DIFF
--- a/azuread/data_user_test.go
+++ b/azuread/data_user_test.go
@@ -31,14 +31,44 @@ func TestAccDataSourceAzureADUser_byUserPrincipalName(t *testing.T) {
 	})
 }
 
-func testAccAzureADUserDataSource_byUserPrincipalName(id, password string) string {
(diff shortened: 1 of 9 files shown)


azuread/data_user_test.go
```diff
@@ -31,14 +31,44 @@ func TestAccDataSourceAzureADUser_byUserPrincipalName(t *testing.T) {
...
+	})
+}
+
 func testAccAzureADUserDataSource_byUserPrincipalName(id, password string) string {
-	template := testAccADUser_basic(id, password)
 	return fmt.Sprintf(`
+%s
 
+data "azuread_user" "test" {
+	user_principal_name = "${azuread_user.test.user_principal_name}"
```
mbfrahry: Should this be by `object_id`

PR #99, opened 2019-06-06
add explicit object_id property to resources

katbyte:


website/docs/r/user.html.markdown
```diff
@@ -39,5 +39,6 @@ The following arguments are supported:
 
 The following attributes are exported:
 
+* `id` - The Object ID of the Azure AD User.
```
mbfrahry: Is this one supposed to be `object_id`
```suggestion
* `object_id` - The Object ID of the Azure AD User.
```

PR #63, opened 2019-03-13
New resource 'azuread_group_member'

tiwood:
This adds a new resource to manage Azure AD Group Membership with Terraform.

- [x] New resource 'azuread_group_member'
- [x] Add tests
- [x] Add documentation

## Example Usage

```hcl

data "azuread_user" "my_user" {
  user_principal_name = "johndoe@hashicorp.com"
}

resource "azuread_group" "my_group" {
  name = "my_group"
}

resource "azuread_group_member" "test" {
  group_object_id   = "${azuread_group.my_group.id}"
  member_object_id  = "${data.azuread_user.my_user.id}"
}
```

## Argument Reference

The following arguments are supported:

* `group_object_id` - (Required) The Object ID of the Azure AD Group you want to add the Member to.
* `member_object_id` - (Required) The Object ID of the Azure AD Object you want to add as a Member to the Group. Supported Object types are Users, Groups or Service Principals.

PR #100, opened 2019-06-06
Resource: 'azuread_group_member'

evenh:
This PR continues where #63 left off. In addition to fixing the comments pointed out in the review, it introduces support for adding members to a group.

Go is not my main language and I'm not that familiar with Terraform so don't be afraid to do a proper review :-)

### DIFF
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -1,0 +1,31 @@
+package graph
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
+)
+
+func GroupAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.Context) ([]string, error) {
+	it, err := client.GetGroupMembersComplete(ctx, groupId)
+
+	if err != nil {
+		return nil, fmt.Errorf("Error listing existing group members from Azure AD Group with ID %q: %+v", groupId, err)
+	}
+
+	existingMembers := make([]string, 0)
+
+	for it.NotDone() {
+		currUser, _ := it.Value().AsUser()
+		existingMembers = append(existingMembers, *currUser.ObjectID)
+		if err := it.NextWithContext(ctx); err != nil {
+			return nil, fmt.Errorf("Error during pagination of group members from Azure AD Group with ID %q: %+v", groupId, err)
+		}
+	}
+
+	log.Printf("[DEBUG] %d members in Azure AD group with ID: %q", len(existingMembers), groupId)
+
+	return existingMembers, nil
+}
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -17,9 +17,27 @@ func GroupAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.
 
 	existingMembers := make([]string, 0)
 
+	var memberObjectID string
 	for it.NotDone() {
-		currUser, _ := it.Value().AsUser()
-		existingMembers = append(existingMembers, *currUser.ObjectID)
+		// possible members are users, groups or service principals
+		// we try to 'cast' each result as the corresponding type and diff
+		// if we found the object we're looking for
+		user, _ := it.Value().AsUser()
+		if user != nil {
+			memberObjectID = *user.ObjectID
+		}
+
+		group, _ := it.Value().AsADGroup()
+		if group != nil {
+			memberObjectID = *group.ObjectID
+		}
+
+		servicePrincipal, _ := it.Value().AsServicePrincipal()
+		if servicePrincipal != nil {
+			memberObjectID = *servicePrincipal.ObjectID
+		}
+
+		existingMembers = append(existingMembers, memberObjectID)
 		if err := it.NextWithContext(ctx); err != nil {
 			return nil, fmt.Errorf("Error during pagination of group members from Azure AD Group with ID %q: %+v", groupId, err)
 		}
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -29,3 +29,18 @@ func GroupAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.
 
 	return existingMembers, nil
 }
+
+func GroupAddMember(groupId string, member string, client graphrbac.GroupsClient, ctx context.Context) error {
+	memberGraphURL := fmt.Sprintf("https://graph.windows.net/%s/directoryObjects/%s", client.TenantID, member)
+
+	properties := graphrbac.GroupAddMemberParameters{
+		URL: &memberGraphURL,
+	}
+
+	log.Printf("[DEBUG] Adding member with id %q to Azure AD group with id %q", member, groupId)
+	if _, err := client.AddMember(ctx, groupId, properties); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -57,7 +57,7 @@ func GroupAddMember(client graphrbac.GroupsClient, ctx context.Context, groupId 
 
 	log.Printf("[DEBUG] Adding member with id %q to Azure AD group with id %q", member, groupId)
 	if _, err := client.AddMember(ctx, groupId, properties); err != nil {
-		return err
+		return fmt.Errorf("Error adding group member %q to Azure AD Group with ID %q: %+v", member, groupId, err)
 	}
 
 	return nil
@@ -68,7 +68,7 @@ func GroupAddMembers(client graphrbac.GroupsClient, ctx context.Context, groupId
 		err := GroupAddMember(client, ctx, groupId, memberUuid)
 
 		if err != nil {
-			return err
+			return fmt.Errorf("Error while adding members to Azure AD Group with ID %q: %+v", groupId, err)
 		}
 	}
 
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -62,3 +62,15 @@ func GroupAddMember(client graphrbac.GroupsClient, ctx context.Context, groupId 
 
 	return nil
 }
+
+func GroupAddMembers(client graphrbac.GroupsClient, ctx context.Context, groupId string, members []string) error {
+	for _, memberUuid := range members {
+		err := GroupAddMember(client, ctx, groupId, memberUuid)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -8,7 +8,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
 )
 
-func GroupAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.Context) ([]string, error) {
+func GroupAllMembers(client graphrbac.GroupsClient, ctx context.Context, groupId string) ([]string, error) {
 	it, err := client.GetGroupMembersComplete(ctx, groupId)
 
 	if err != nil {
@@ -48,7 +48,7 @@ func GroupAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.
 	return existingMembers, nil
 }
 
-func GroupAddMember(groupId string, member string, client graphrbac.GroupsClient, ctx context.Context) error {
+func GroupAddMember(client graphrbac.GroupsClient, ctx context.Context, groupId string, member string) error {
 	memberGraphURL := fmt.Sprintf("https://graph.windows.net/%s/directoryObjects/%s", client.TenantID, member)
 
 	properties := graphrbac.GroupAddMemberParameters{
--- a/azuread/resource_group.go
+++ b/azuread/resource_group.go
@@ -117,7 +117,7 @@ func resourceGroupRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("name", resp.DisplayName)
 	d.Set("object_id", resp.ObjectID)
 
-	members, err := fetchAllMembers(d.Id(), client, ctx)
+	members, err := graph.GroupAllMembers(d.Id(), client, ctx)
 	if err != nil {
 		return err
 	}
@@ -132,7 +132,7 @@ func resourceGroupUpdate(d *schema.ResourceData, meta interface{}) error {
 	ctx := meta.(*ArmClient).StopContext
 
 	if v, ok := d.GetOk("members"); ok && d.HasChange("members") {
-		existingMembers, err := fetchAllMembers(d.Id(), client, ctx)
+		existingMembers, err := graph.GroupAllMembers(d.Id(), client, ctx)
 		if err != nil {
 			return err
 		}
@@ -171,28 +171,6 @@ func resourceGroupDelete(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	return nil
-}
-
-func fetchAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.Context) ([]string, error) {
-	it, err := client.GetGroupMembersComplete(ctx, groupId)
-
-	if err != nil {
-		return nil, fmt.Errorf("Error listing existing group members from Azure AD Group with ID %q: %+v", groupId, err)
-	}
-
-	existingMembers := make([]string, 0)
-
-	for it.NotDone() {
-		currUser, _ := it.Value().AsUser()
-		existingMembers = append(existingMembers, *currUser.ObjectID)
-		if err := it.NextWithContext(ctx); err != nil {
-			return nil, fmt.Errorf("Error during pagination of group members from Azure AD Group with ID %q: %+v", groupId, err)
-		}
-	}
-
-	log.Printf("[DEBUG] %d members in Azure AD group with ID: %q", len(existingMembers), groupId)
-
-	return existingMembers, nil
 }
 
 func addMember(groupId string, member string, client graphrbac.GroupsClient, ctx context.Context) error {
--- a/azuread/resource_group.go
+++ b/azuread/resource_group.go
@@ -130,7 +130,7 @@ func resourceGroupUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*ArmClient).groupsClient
 	ctx := meta.(*ArmClient).StopContext
 
-	if v, ok := d.GetOk("members"); ok && d.HasChange("members") {
+	if v, ok := d.GetOkExists("members"); ok && d.HasChange("members") {
 		existingMembers, err := graph.GroupAllMembers(d.Id(), client, ctx)
 		if err != nil {
 			return err
--- a/azuread/resource_group.go
+++ b/azuread/resource_group.go
@@ -82,12 +82,8 @@ func resourceGroupCreate(d *schema.ResourceData, meta interface{}) error {
 	if v, ok := d.GetOk("members"); ok {
 		members := tf.ExpandStringSlicePtr(v.(*schema.Set).List())
 
-		for _, memberUuid := range *members {
-			err := graph.GroupAddMember(client, ctx, *group.ObjectID, memberUuid)
-
-			if err != nil {
-				return err
-			}
+		if err := graph.GroupAddMembers(client, ctx, *group.ObjectID, *members); err != nil {
+			return err
 		}
 	}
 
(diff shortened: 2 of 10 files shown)


evenh:
ping @katbyte :)

azuread/resource_group.go
```diff
@@ -88,9 +116,50 @@ func resourceGroupRead(d *schema.ResourceData, meta interface{}) error {
 
 	d.Set("name", resp.DisplayName)
 	d.Set("object_id", resp.ObjectID)
+
+	members, err := fetchAllMembers(d.Id(), client, ctx)
```
katbyte: This might be better as a graph helper: `graph.GroupAllMembers()`

azuread/resource_group.go
```diff
@@ -88,9 +116,50 @@ func resourceGroupRead(d *schema.ResourceData, meta interface{}) error {
...
+			if resp, err := client.RemoveMember(ctx, d.Id(), existingMember); err != nil {
+				if !ar.ResponseWasNotFound(resp) {
+					return fmt.Errorf("Error Deleting group member %q from Azure AD Group with ID %q: %+v", existingMember, d.Id(), err)
+				}
+			}
+		}
+
+		for _, newMember := range membersToAdd {
+			if err := addMember(d.Id(), newMember, client, ctx); err != nil {
+				return err
```
katbyte: Could we wrap this error like above

azuread/resource_group.go
```diff
@@ -103,3 +172,40 @@ func resourceGroupDelete(d *schema.ResourceData, meta interface{}) error {
...
+			return nil, fmt.Errorf("Error during pagination of group members from Azure AD Group with ID %q: %+v", groupId, err)
+		}
+	}
+
+	log.Printf("[DEBUG] %d members in Azure AD group with ID: %q", len(existingMembers), groupId)
+
+	return existingMembers, nil
+}
+
+func addMember(groupId string, member string, client graphrbac.GroupsClient, ctx context.Context) error {
```
katbyte: This might be better as `graph.GroupAddMember()`

azuread/resource_group_member.go
```diff
@@ -0,0 +1,151 @@
...
+	return resourceGroupMemberRead(d, meta)
+}
+
+func resourceGroupMemberRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ArmClient).groupsClient
+	ctx := meta.(*ArmClient).StopContext
+
+	id := strings.Split(d.Id(), "/")
+	if len(id) != 2 {
+		return fmt.Errorf("ID should be in the format {groupObjectId}/{memberObjectId} - but got %q", d.Id())
```
katbyte: Could we change this format to: `{groupObjectId}/member/{memberObjectId}` otherwise it'll conflict with the group_owner resource

azuread/resource_group_member.go
```diff
@@ -0,0 +1,151 @@
...
+	memberID := d.Get("member_object_id").(string)
+	tenantID := client.TenantID
+
+	memberGraphURL := fmt.Sprintf("https://graph.windows.net/%s/directoryObjects/%s", tenantID, memberID)
+
+	properties := graphrbac.GroupAddMemberParameters{
+		URL: &memberGraphURL,
+	}
+
+	if _, err := client.AddMember(ctx, groupID, properties); err != nil {
```
katbyte: Will this fail if the member already exists?

katbyte: Could we also reuse the `graph.GroupAddMember` function discussed above?

azuread/resource_group_test.go
```diff
@@ -123,3 +167,12 @@ resource "azuread_group" "test" {
 }
 `, id)
 }
+
+func testAccAzureADGroupWithMembers(id string, members []string) string {
+	return fmt.Sprintf(`
+resource "azuread_group" "test" {
+  name = "acctest%s"
```
katbyte:
```suggestion
  name   = "acctest%s"
```

azuread/resource_group_member_test.go
```diff
@@ -0,0 +1,224 @@
...
+	display_name          = "acctestA%[1]s"
+	password              = "%[2]s"
+}
+	
+resource "azuread_group" "test" {
+	name = "acctest%[1]s"
+}
+
+resource "azuread_group_member" "test" {
+	group_object_id 	= "${azuread_group.test.id}"
```
katbyte: I have added an object_id property to all resources, so could we use that instead to be explicit about what we are using here?
```suggestion
	group_object_id 	= "${azuread_group.test.object_id}"
```

website/docs/r/group.markdown
```diff
@@ -15,17 +15,35 @@ Manages a Group within Azure Active Directory.
...
+```hcl
+resource "azuread_user" "my_user" {
+  display_name          = "John Doe"
+  password              = "notSecure123"
+  user_principal_name   = "john.doe@terraform.onmicrosoft.com"
+}
+
+resource "azuread_group" "my_group" {
+  name = "MyGroup"
+  members = [ azuread_user.my_user.id /*, more users */ ]
```
katbyte: Same here
```suggestion
  members = [ azuread_user.my_user.object_id /*, more users */ ]
```

website/docs/r/group.markdown
```diff
@@ -15,17 +15,35 @@ Manages a Group within Azure Active Directory.
...
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
-* `name` - (Required) The display name for the Group.
+* `name` - (Required) The display name for the Group. Changing this forces a new resource to be created.
+* `members` (Optional) A set of users who should be members of this Group.
```
katbyte: This should support both groups and SPs as well?

website/docs/r/group.markdown
```diff
@@ -37,6 +55,8 @@ The following attributes are exported:
 
 * `name` - The Display Name of the Group.
 
+* `members` - The Group Members in the Group.
```
katbyte:
```suggestion
* `members` - The Members of the Group.
```

azuread/resource_group.go
```diff
@@ -103,3 +172,40 @@ func resourceGroupDelete(d *schema.ResourceData, meta interface{}) error {
...
+	it, err := client.GetGroupMembersComplete(ctx, groupId)
+
+	if err != nil {
+		return nil, fmt.Errorf("Error listing existing group members from Azure AD Group with ID %q: %+v", groupId, err)
+	}
+
+	existingMembers := make([]string, 0)
+
+	for it.NotDone() {
+		currUser, _ := it.Value().AsUser()
```
katbyte: We should probably support all types here. That way should be able to reuse this function for the group_member object

website/docs/r/group_member.markdown
```diff
@@ -0,0 +1,55 @@
+---
+layout: "azuread"
+page_title: "Azure Active Directory: azuread_group_member"
+sidebar_current: "docs-azuread-resource-azuread-group-member"
+description: |-
+  Manages a Group Membership within Azure Active Directory.
```
katbyte: This might sound better:
```suggestion
  Manages a single Group Membership within Azure Active Directory.
```

website/docs/r/group_member.markdown
```diff
@@ -0,0 +1,55 @@
...
+
+---
+
+# azuread_group_member
+
+Manages a Group Membership within Azure Active Directory.
```
katbyte: Could we add a warning here to not use this resource and the `azuread_group.members` property at the same time?

website/docs/r/group.markdown
katbyte: AS well a warning to not use the `azuread_group_member` at the same time might be handy here

katbyte:
I have also opened a PR for a data source to get object IDs for multiple users (#109) that would be nice to be used here in the example/tests here somewhere

azuread/resource_group_member.go
evenh: I've refactored to use the new helper function. However, if a member already exists it fails. What is the desired outcome here? Given the following HCL:

```hcl
// User one already exists
resource "azuread_group" "some_group" {
  name = "SomeGroup"
}

resource "azuread_group_member" "one" {
  group_object_id   = azuread_group.some_group.id
  member_object_id  = azuread_user.one.id
}

resource "azuread_group_member" "another" {
  group_object_id   = azuread_group.some_group.id
  member_object_id  = azuread_user.one.id
}
```

In that case Terraform sees two "new" resource in `azuread_group_member.one` and  `azuread_group_member.another`. After applying the first `one`, `another` fails with

```
graphrbac.GroupsClient#AddMember: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="Unknown" Message="Unknown service error" Details=[{"odata.error":{"code":"Request_BadRequest","date":"2019-06-18T11:09:20","message":{"lang":"en","value":"One or more added object references already exist for the following modified properties: 'members'."},"requestId":"c7ad6982-559e-46eb-bfed-56b6455141c8"}}]
```

azuread/resource_group_member.go
evenh: Just for this specific error?

evenh:
Some of the comments is now fixed!

I've used a day trying to make acceptance tests work from my local setup but to no avail (makes writing new acceptance tests a little hard). Any pointers on how the service principal used for running acceptance tests is configured?

katbyte:
Hi @evenh,

You should be able to follow the instructions here: https://www.terraform.io/docs/providers/azuread/auth/service_principal_client_secret.html

Make sure you grant it the appropriate access here: https://www.terraform.io/docs/providers/azuread/auth/service_principal_configuration.html

I believe there is also an example on master in the `examples` folder demonstrating this too if you'd prefer a terraform'y way

azuread/resource_group.go
```diff
@@ -44,6 +44,7 @@ func resourceGroup() *schema.Resource {
 			"members": {
 				Type:     schema.TypeSet,
 				Optional: true,
+				Computed: true,
```
landro: What's the rationale behind this field being computed, @evenh ? To me, this sounds strange?

katbyte: @landro, this is because members can also be managed via the separate resource or cli or azure. Having it be computed allows the resource to read them in if unset. and then if set try and force them to be X.

katbyte:
@evenh, is this ready for re-review?

evenh:
Not quite yet @katbyte. I managed to get the acceptance test service principal set up correctly from the links you posted. I fixed some of the existing tests but have not yet added the tests you requested. 

Will let you know as soon as they are done.

katbyte:
Thanks for the update!

evenh:
Ready for a re-review now @katbyte. Also note the two open inline comments :-)

azuread/resource_group_member.go
katbyte: The ID for this resource should be in that format, left an additional comment above

azuread/resource_group_member.go
```diff
@@ -0,0 +1,113 @@
...
+	ctx := meta.(*ArmClient).StopContext
+
+	groupID := d.Get("group_object_id").(string)
+	memberID := d.Get("member_object_id").(string)
+
+	if err := graph.GroupAddMember(groupID, memberID, client, ctx); err != nil {
+		return err
+	}
+
+	id := fmt.Sprintf("%s/%s", groupID, memberID)
```
katbyte: Could we change this to:
```suggestion
	id := fmt.Sprintf("%s/member/%s", groupID, memberID)
```

azuread/resource_group.go
```diff
@@ -61,6 +78,19 @@ func resourceGroupCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.SetId(*group.ObjectID)
 
+	// Add members if specified
+	if v, ok := d.GetOk("members"); ok {
+		members := tf.ExpandStringSlicePtr(v.(*schema.Set).List())
+
+		for _, memberUuid := range *members {
```
katbyte: I think it might be nice to have another function `graph.GroupAddMembers()` that wraps this functionality around the singular function.

azuread/helpers/graph/group.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/graphrbac/1.6/graphrbac"
+)
+
+func GroupAllMembers(groupId string, client graphrbac.GroupsClient, ctx context.Context) ([]string, error) {
```
katbyte: Could we change the signiture to:
```suggestion
func GroupAllMembers(client graphrbac.GroupsClient, ctx context.Context, groupId string) ([]string, error) {
```

As it will better match the actual API call

azuread/helpers/graph/group.go
```diff
@@ -0,0 +1,64 @@
...
+			return nil, fmt.Errorf("Error during pagination of group members from Azure AD Group with ID %q: %+v", groupId, err)
+		}
+	}
+
+	log.Printf("[DEBUG] %d members in Azure AD group with ID: %q", len(existingMembers), groupId)
+
+	return existingMembers, nil
+}
+
+func GroupAddMember(groupId string, member string, client graphrbac.GroupsClient, ctx context.Context) error {
```
katbyte: As above:
```suggestion
func GroupAddMember(client graphrbac.GroupsClient, ctx context.Context, groupId string, member string) error {
```

website/docs/r/group_member.markdown
```diff
@@ -0,0 +1,55 @@
...
+page_title: "Azure Active Directory: azuread_group_member"
+sidebar_current: "docs-azuread-resource-azuread-group-member"
+description: |-
+  Manages a single Group Membership within Azure Active Directory.
+
+---
+
+# azuread_group_member
+
+Manages a single Group Membership within Azure Active Directory. Do not use this resource at the same time as `azuread_group.members`.
```
katbyte: Could we make the warning at the end a note?

website/docs/r/group.markdown
```diff
@@ -15,17 +15,35 @@ Manages a Group within Azure Active Directory.
...
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
-* `name` - (Required) The display name for the Group.
+* `name` - (Required) The display name for the Group. Changing this forces a new resource to be created.
+* `members` (Optional) A set of members who should be present in this Group. Supported Object types are Users, Groups or Service Principals. Do not use `azuread_group_member` at the same time as this argument.
```
katbyte: Could we make the warning at the end a note?

azuread/resource_group_test.go
```diff
@@ -116,10 +231,112 @@ func testCheckAzureADGroupDestroy(s *terraform.State) error {
...
+}
+
+resource "azuread_service_principal" "test_sp_%[1]s" {
+  application_id = azuread_application.test_app_%[1]s.application_id
+}
+
+`, id)
+}
+
+func group(id string) string {
```
katbyte: You could probably reuse the test config for a single group?

azuread/resource_group_test.go
```diff
@@ -116,10 +231,112 @@ func testCheckAzureADGroupDestroy(s *terraform.State) error {
...
+
+func formatMembersAsUser(members []string) []string {
+	vsm := make([]string, len(members))
+	for i, v := range members {
+		vsm[i] = "azuread_user.acctest_user_" + v + ".object_id"
+	}
+	return vsm
+}
+
+func servicePrincipal(id string) string {
```
katbyte: Same here

azuread/resource_group_test.go
```diff
@@ -55,17 +87,100 @@ func TestAccAzureADGroup_complete(t *testing.T) {
...
+				Check:  assertResourceWithMemberCount(id, "0"),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Group with 1 member
+			{
+				Config: user(id) + groupWithMembers(id, fmt.Sprintf("azuread_user.acctest_user_%[1]s.object_id", id)),
```
katbyte: Could we move this into a names function for consistency with the rest of the provider?

azuread/resource_group_test.go
```diff
@@ -55,17 +87,100 @@ func TestAccAzureADGroup_complete(t *testing.T) {
...
+				Check:  assertResourceWithMemberCount(id, "1"),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Group with multiple members
+			{
+				Config: user(id+"a") + user(id+"b") + user(id+"c") + groupWithMembers(id, fmt.Sprintf("azuread_user.acctest_user_%[1]s.object_id, azuread_user.acctest_user_%[2]s.object_id, azuread_user.acctest_user_%[3]s.object_id", id+"a", id+"b", id+"c")),
```
katbyte: As above

azuread/resource_group_test.go
```diff
@@ -55,17 +87,100 @@ func TestAccAzureADGroup_complete(t *testing.T) {
...
+				Check:  assertResourceWithMemberCount(id, "3"),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Group with a different member
+			{
+				Config: servicePrincipal(id) + groupWithMembers(id, fmt.Sprintf("azuread_service_principal.test_sp_%[1]s.object_id", id)),
```
katbyte: As above

azuread/resource_group_test.go
```diff
@@ -116,10 +231,112 @@ func testCheckAzureADGroupDestroy(s *terraform.State) error {
...
 func testAccAzureADGroup(id string) string {
 	return fmt.Sprintf(`
 resource "azuread_group" "test" {
   name = "acctest%s"
 }
 `, id)
 }
+
+func testAccAzureADGroupWithDiverseMembers(id string) string {
+	var sb strings.Builder
```
katbyte: Could we just use a large here string here with a fmt.Sprintf for consistency with the rest of the provider?

azuread/resource_group_test.go
```diff
@@ -116,10 +231,112 @@ func testCheckAzureADGroupDestroy(s *terraform.State) error {
...
+	sb.WriteString(servicePrincipal(id))
+	sb.WriteString(group(id))
+	sb.WriteString(user(id))
+	sb.WriteString(groupWithMembers(id, fmt.Sprintf("azuread_user.acctest_user_%[1]s.object_id, azuread_group.test_g_%[1]s.object_id, azuread_service_principal.test_sp_%[1]s.object_id", id)))
+
+	return sb.String()
+}
+
+func testAccAzureADGroupWithMembers(id string, members []string) string {
+	var sb strings.Builder
```
katbyte: As above

azuread/resource_group_test.go
evenh: If using `testAccAzureADGroup(id)` instead of `group(id)`, duplicate resources named `azuread_group.test` will exist, and therefore fail.

azuread/resource_group_test.go
evenh: Not using `fmt.Sprintf` gives us a few benefits:

- Reuse
- Looping and other code functionality

On the other hand I can understand an argument of making it less obvious what it does since the HCL is not visible in plain text.

evenh: I prefer the current approach but if it's not feasible I could try to rewrite the test configuration.

evenh:
I've fixed most of your comments @katbyte and added some of my own :)

azuread/resource_group_test.go
katbyte: you could change it to use `azuread_group.test%s` and pass in an additional argument?

azuread/resource_group_test.go
katbyte: We definitely would prefer HCL as this raises the barriers for maintainers & community contributors as well as people will often copy and past the tests for their use. And most importantly is not consistent with any of the azure provider tests.

if you'd like to loop maybe just have a simple loop at the start appending to a string.

evenh:
I've converted all of `resource_group_test.go` to HCL @katbyte :)

azuread/resource_group.go
```diff
@@ -88,9 +114,50 @@ func resourceGroupRead(d *schema.ResourceData, meta interface{}) error {
...
+		for _, existingMember := range membersForRemoval {
+			log.Printf("[DEBUG] Removing member with id %q from Azure AD group with id %q", existingMember, d.Id())
+			if resp, err := client.RemoveMember(ctx, d.Id(), existingMember); err != nil {
+				if !ar.ResponseWasNotFound(resp) {
+					return fmt.Errorf("Error Deleting group member %q from Azure AD Group with ID %q: %+v", existingMember, d.Id(), err)
+				}
+			}
+		}
+
+		for _, newMember := range membersToAdd {
```
katbyte: You should be able to use `GroupAddMembers` from above here?

azuread/resource_group.go
katbyte: Still not wrapped:
```suggestion
				return fmt.Errorf("Error adding group members to Azure AD Group with ID %q: %+v", d.Id(), err)
```

evenh:
There you go @katbyte :-)

evenh:
Could you please have another look @katbyte?

aambert:
Up

PR #117, opened 2019-07-11
Wait on password replication

jlpedrosa:
Hi @katbyte 

This PR waits on Password replication, same as done in #93 for Application and service principal.
Please note that I have not run the integration tests as get a workable AAD, may be bit complex for me.

### DIFF
--- a/azuread/resource_service_principal_password.go
+++ b/azuread/resource_service_principal_password.go
@@ -62,7 +62,7 @@ func resourceServicePrincipalPasswordCreate(d *schema.ResourceData, meta interfa
 		return client.ListPasswordCredentials(ctx, id.ObjectId)
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


azuread/resource_service_principal_password.go
```diff
@@ -58,6 +58,13 @@ func resourceServicePrincipalPasswordCreate(d *schema.ResourceData, meta interfa
 
 	d.SetId(id.String())
 
+	_, err = graph.WaitForReplication(func() (interface{}, error) {
+		return client.Get(ctx, id.ObjectId)
```
katbyte: So i think this should be:
```suggestion
		return client.ListPasswordCredentials(ctx, id.ObjectId)
```

as that is how we are checking if the credentials exist?

PR #127, opened 2019-07-24
azuread_service_principal - support for the AppRoleAssignmentRequired property

dmeenhuis:
This adds support for setting the `AppRoleAssignmentRequired` property on a service principal.

I'm not sure if I implemented the update function correctly, but since it was not already there I think I had to create it.

### DIFF
--- a/azuread/resource_service_principal.go
+++ b/azuread/resource_service_principal.go
@@ -35,6 +35,11 @@ func resourceServicePrincipal() *schema.Resource {
 				ValidateFunc: validate.UUID,
 			},
 
+			"app_role_assignment_required": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+
 			"display_name": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -43,11 +48,6 @@ func resourceServicePrincipal() *schema.Resource {
 			"object_id": {
 				Type:     schema.TypeString,
 				Computed: true,
-			},
-
-			"app_role_assignment_required": {
-				Type:     schema.TypeBool,
-				Optional: true,
 			},
 
 			"oauth2_permissions": graph.SchemaOauth2Permissions(),
--- a/azuread/resource_service_principal.go
+++ b/azuread/resource_service_principal.go
@@ -21,6 +21,7 @@ func resourceServicePrincipal() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceServicePrincipalCreate,
 		Read:   resourceServicePrincipalRead,
+		Update: resourceServicePrincipalUpdate,
 		Delete: resourceServicePrincipalDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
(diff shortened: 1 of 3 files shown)


azuread/resource_service_principal_test.go
```diff
@@ -131,7 +133,8 @@ resource "azuread_application" "test" {
 }
 
 resource "azuread_service_principal" "test" {
-  application_id = "${azuread_application.test.application_id}"
+	application_id = "${azuread_application.test.application_id}"
+	app_role_assignment_required = true
```
katbyte: COuld we align these assingments here?

website/docs/r/service_principal.html.markdown
```diff
@@ -27,6 +27,7 @@ resource "azuread_application" "example" {
 
 resource "azuread_service_principal" "example" {
   application_id = "${azuread_application.example.application_id}"
+  app_role_assignment_required  = false
```
katbyte: Could we aling these assingments here?

azuread/resource_service_principal.go
```diff
@@ -44,6 +45,11 @@ func resourceServicePrincipal() *schema.Resource {
 				Computed: true,
 			},
 
+			"app_role_assignment_required": {
```
katbyte: Could we move this above the computed properties?

azuread/resource_service_principal_test.go
dmeenhuis: Done

website/docs/r/service_principal.html.markdown
dmeenhuis: Done

azuread/resource_service_principal.go
dmeenhuis: Done. What about the `oauth2_permissions` and `tags`, would you like me to move those above the computed properties as well, for consistency's sake?

dmeenhuis:
I added a test to check the update function. Wasn't entirely sure how to go about it, so I took the update test from application and followed a similar approach for service principal and the test seems to pass. Could you let me know whether I implemented it correctly?

website/docs/r/service_principal.html.markdown
```diff
@@ -36,43 +36,47 @@ resource "azuread_service_principal" "example" {
 
 The following arguments are supported:
 
-* `application_id` - (Required) The ID of the Azure AD Application for which to create a Service Principal.
+- `application_id` - (Required) The ID of the Azure AD Application for which to create a Service Principal.
```
katbyte: Could we not change these from `*` to `-` for consistency with the rest of the provider docs?

azuread/resource_service_principal_test.go
katbyte: `oauth2_permissions` would be great, tags however we always leave at the end. Thanks!

dmeenhuis:
Thanks for merging these changes! I wasn't able to respond earlier since I was on holidays for the past 3 weeks.

PR #141, opened 2019-08-27
azuread_user - support for the `usage_location` property

landro:
Setting usage location is required for users that will be assigned licenses (e.g. Azure Active Directory Premium P1/2 licenses) due to legal requirement to check for availability of services in countries.


evenh:
ping @katbyte

PR #167, opened 2019-10-25
Uses random_password resource for sp examples

artburkart:
Hi 👋 ,

I noticed `random_string` was used in the examples. Thought it might be helpful to use `random_password` instead. I believe @katbyte is maybe the person to ping?

Thanks!


artburkart:
Thanks! :+1:

Issue #162, opened 2019-10-15
azuread_application: support for setting oauth2 permissions (exposing APIs)


joakimhellum:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently Oauth2 permissions can be read for an application, but not set. It would be nice to add support for setting Oauth2 permissions as well (exposing APIs). However normally it's hard to maintain all Oauth2 permissions in Terraform, since some would be automatically created outside of Terraform. And such the suggestion is to add separate resource `azuread_application_oauth2_permission` for this.

I am happy to create a PR for this.

### New or Affected Resource(s)

* azuread_application_oauth2_permission

### Potential Terraform Configuration

```hcl
resource "azuread_application_oauth2_permission" "test" {
  application_object_id = azuread_application.main.object_id

  type = "User"

  admin_consent_description = "Allow access to example on behalf of the signed-in user."
  admin_consent_display_name = "Access example"

  user_consent_description = "Allow access to example on your behalf."
  user_consent_display_name = "Access example"

  value = "user_impersonation"
}
```

### References

* N/A

rdbartram:
would love to see this happen. Why would it be its own resource though and not a block on azuread_application?

joakimhellum-in:
@rdbartram You are right, there is no real justification for this currently. The reason for suggestion this initially is because i'm still not completely sure how to best manage the default ouath2 permission (`user_impersonation`) that is created for every application. 

> **oauth2Permissions: What actions does the app expose?**
>
> The oauth2Permissions collection publishes the list of things that client applications can do with your app—the scopes the app admits, mostly, but that comes into play only in case your app is a web API. If your app is a web application with a UX, the expectation is that browsers will request tokens for your app with the goal of signing in. That does not require any entry for web sign-on, the scenario considered in this chapter, so I thought of deferring coverage of this property until I get to exposing your own web API, but some of the concepts will come in handy sooner than that, so I’ll give you a bit of background now. Let’s take a closer look at the only entry in the oauth2Permissions collection for the sample application:
>
> {
[...]

PR #187, opened 2019-12-10
Added support for OAuth2 permissions in resource application.

jorgecarleitao:
See #162 for details.


jorgecarleitao:
I've tested both create, destroy, and change of the resource. The config I used was

```
provider "azuread" {

  subscription_id = "..."
  client_id       = "..."
  client_secret   = "..."
  tenant_id       = "..."
}

# Create an application
  name = "ExampleApp"

  oauth2_permissions {
      admin_consent_description  = "example"
      admin_consent_display_name = "example"
      type                       = "Admin"
      value                      = "access"
  }

  oauth2_permissions {
      admin_consent_description  = "example"
      admin_consent_display_name = "example"
      type                       = "User"
      value                      = "write"
  }
}
```

and 
1. `terraform apply -auto-approve`
2. erased one of the items and `terraform apply -auto-approve`
3. added the item back and `terraform apply -auto-approve`
4. changed value of second item from `write` to `read` and `terraform apply -auto-approve`
5. `terraform destroy -auto-approve`

I validated each change both on the local state and on the portal.

azuread/resource_application.go
```diff
@@ -202,7 +202,61 @@ func resourceApplication() *schema.Resource {
 				Computed: true,
 			},
 
-			"oauth2_permissions": graph.SchemaOauth2PermissionsComputed(),
+			"oauth2_permissions": {
+				Type:     schema.TypeSet,
+				Optional: true,
```
ennui93: Per [katbyte's comment](https://github.com/terraform-providers/terraform-provider-azuread/issues/162#issuecomment-543807695) and the other discussion in #162, should this also be computed?

website/docs/r/application.html.markdown
```diff
@@ -125,6 +127,24 @@ The following arguments are supported:
 
 * `value` - (Optional) Specifies the value of the roles claim that the application should expect in the authentication and access tokens.
 
+---
+
+`oauth2_permission` supports the following:
```
ennui93:
```suggestion
`oauth2_permissions` supports the following:
```

stevebering:
I would really like to see this merged and released as it is preventing me from using Terraform to automate our App Registrations.

PeterStephenson:
We're currently also getting stuck by this, is there an ETA for the feature or anything we can do to help finish it?

MetalGuardian:
I'd also like to see this merge. Ho can we make it happen?

katbyte:
@MetalGuardian - iirc there needs to be a good ux for leaving it empty (probably marking it as computed) and then changing/adding/updating new permissions. i pulled it locally and tried a few things quickly a couple months ago but nothing stuck. So the PR basically needs to be update as per above comments & then tweaked to have a better UX around the default values.

manicminer:
Closing in favor of #252

Issue #206, opened 2020-01-17
azuread_user support for immutable_id


derek-burdick:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

ImmutableID is used for SAML IDP integration with Azure Active Directory. Being able to add an immutable id on user creation will simplify integration with SSO providers.

### New or Affected Resource(s)

* azuread_user

### Potential Terraform Configuration

```hcl
resource azuread_user newuser {
     user_principal_name  = "newuser1"
     display_name = "New User"
     password = "super secret"
     immutable_id = "abc123xyz"
}
```

### References

* #0000

PR #207, opened 2020-01-17
azuread_user support for immutable_id property

derek-burdick:
Implements #206

### DIFF
--- a/azuread/data_user.go
+++ b/azuread/data_user.go
@@ -61,6 +61,11 @@ func dataUser() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"immutable_id": {
+				Type: schema.TypeString,
+				Computed: false,
+			}
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


azuread/data_user.go
```diff
@@ -105,6 +105,7 @@ func dataSourceUserRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("mail", user.Mail)
 	d.Set("mail_nickname", user.MailNickname)
 	d.Set("usage_location", user.UsageLocation)
+	d.Set("immutable_id", user.ImmutableID)
```
katbyte: You'll need to tht thi to the DS schema

azuread/resource_user.go
```diff
@@ -201,6 +213,11 @@ func resourceUserUpdate(d *schema.ResourceData, meta interface{}) error {
 		userUpdateParameters.UsageLocation = p.String(usageLocation)
 	}
 
+	if d.HasChange("immutable_id") {
```
katbyte: If this is immutable is it really updateable? 🤔

derek-burdick: It appears to be an unfortunate naming convention. From the documents it can be added to the user after creation, and it can be modified once set.

PR #231, opened 2020-03-19
Remove MinItems from data groups

jpreese:
Was trying to use the `azuread_groups` with the optional `names` parameter, but was getting an error about a value must be returned as they are both set to 1 and return an error.

Looks like a similar issue to: https://github.com/terraform-providers/terraform-provider-azuread/pull/227

### DIFF
--- a/azuread/data_groups.go
+++ b/azuread/data_groups.go
@@ -26,7 +26,6 @@ func dataGroups() *schema.Resource {
 				Type:          schema.TypeList,
 				Optional:      true,
 				Computed:      true,
-				MinItems:      1,
 				ConflictsWith: []string{"names"},
 				Elem: &schema.Schema{
 					Type:         schema.TypeString,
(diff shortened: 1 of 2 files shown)


jpreese:
For context, my use case is that a user can supply an _optional_ value which is a list of Azure AD group names.

I take that variable, and pass it into the `azuread_groups` data provider, but since its optional it might be empty.

I then have another resource that does a `count` on the `azuread_groups` provider to add a resource per group that was returned. If no groups were returned, especially in the case of an empty input, don't recreate any resources.

jpreese:
As an aside, for a future PR, I think there's also value in changing the Object_IDs to be a `set` type. This would make it easier to use `for_each` and the keys should be unique.

manicminer:
Hi @jpreese, thanks for this change suggestion! We've discussed and believe that in this case, it's OK to support empty lists for search inputs. I've moved this to a new PR #257, preserving your contribution, as I've added a test case and clarified the new behavior in the docs.

PR #248, opened 2020-05-12
Update Azure Go SDK version to v42.1.0

magodo:
There is a breaking change on `UserClient.List()` function in package **graphrbac**. This blocks terraform-provider-azurerm from updating the Azure Go SDK to v42.1.0.


katbyte:
Already have an open and approved pr @ #247 🙂 as such will close this

magodo:
@katbyte Sorry for not checking that 😪

Issue #73, opened 2019-04-09
azuread_application: unexpectedly has an 'exposed API' when viewed in the preview portal


DavidR91:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
resource "azuread_application" "app" {
	name = "Example"
	homepage = "https://example"
	reply_urls = "https://example"
	oauth2_allow_implicit_flow = true

	// Microsoft Graph 
	// (not in order):
	// - Directory.Read.All
	// - Group.Read.All
	// - User.Read
	// - User.ReadBasic.All
	// - User.ReadWrite
	required_resource_access {
		resource_app_id = "00000003-0000-0000-c000-000000000000"
		
		resource_access {
			id = "06da0dbc-49e2-44d2-8312-53f166ab848a",
			type = "Scope"
		},

		resource_access {
			id = "5f8c59db-677d-491f-a6b8-5f174b11ec1d",
			type = "Scope"
		},

		resource_access {
			id = "e1fe6dd8-ba31-4d61-89e7-88639da4683d",
			type = "Scope"
		},

		resource_access {
			id = "b340eb25-3456-403f-be2f-af7a0d370277",
			type = "Scope"
		},

		resource_access {
			id = "b4e74841-8e56-480b-be8b-910348b18b4c",
			type = "Scope"
		}
	}

	// Azure management API
	//(not in order):
	// - user_impersonation
	required_resource_access {
		resource_app_id = "797f4846-ba00-4fd7-ba43-dac1f8f63013"
		
		resource_access {
			id = "41094075-9dad-400e-a0bd-54e686782033",
			type = "Scope"
		},
	}
}
```
### Expected Behavior
The Azure AD application created should have the specified API permissions when viewed in the preview portal

### Actual Behavior
Although the application has the permissions specified, it also gains an 'exposed API' visible in the preview portal:

![image](https://user-images.githubusercontent.com/282737/55795937-c5c8c280-5ac0-11e9-8ff1-476655f68680.png)

[...]

PR #252, opened 2020-05-18
application: support for setting oauth2_permissions

manicminer:
Enable setting OAuth 2.0 permission scopes for an AAD application.

If omitted, will be computed from scopes defined out of band.
If specified, will replace any existing scopes configured for an application.
If specified as empty value, will remove any existing scopes from an application.

Thanks to @jorgecarleitao for initial work on #187

Fixes: #73, #162
Replaces: #187

### DIFF
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -211,7 +211,61 @@ func resourceApplication() *schema.Resource {
 				Computed: true,
 			},
 
-			"oauth2_permissions": graph.SchemaOauth2PermissionsComputed(),
+			"oauth2_permissions": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"admin_consent_description": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validate.NoEmptyStrings,
+						},
+
+						"admin_consent_display_name": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validate.NoEmptyStrings,
+						},
+
+						"id": {
+							Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


katbyte:
@manicminer - would it make sense to include a flag `oauth2_permissions_include_default` that injects the default permissions? that way you can add your own and not worry about specifying them, as well as clear & reset to the default

manicminer:
Hi @katbyte, thinking this over a few ways I'm not sure that would work. I can see two options:

1. Hardcode the default scope in the provider and add it on behalf of the user, which feels like it will break at some point?
2. Assume a default scope(s) will be created, which it seems to be, but I've had cases where it was not though I'm not sure exactly what configuration causes this. Also, then we'd have to save the state of any default scopes because they are only added at create time?

Just adding my 2c, as an administrator my personal preference is to be explicit about my application scopes.

manicminer:
Tests pass

Screenshot 2020-05-27 03 09 28

imbatman4711:
Is there an expected date for the 0.10.0? Because I need exactly this feature for my current project :)

manicminer:
Hi @imbatman4711, we're planning on releasing this week which will include this change.

nephinj:
Can you provide a sample or better documentation of an empty value?  I tried setting oauth2_permissions = [] but it does not seem to work for me and I see values in the manifest.  

resource "azuread_application" "Chatbot-adapp" {
  name                       = "${var.prefix}-wapp"
  oauth2_permissions = []
  oauth2_allow_implicit_flow = false
 reply_urls = null
}

manicminer:
Hi @nephinj, I'm observing some extreme delays at the moment with changes reflecting in the portal for applications. I've just ran a few tests myself and the correct updates are being made (and accepted by the API), but it's been over 10 minutes with updates not yet reflecting.

PR #254, opened 2020-05-19
azuread_application: Allow app_role.id to be specified manually

### LINKED ISSUE #150: azuread_application: app_role id can't be specified manually
I'm trying to provision AD groups and applications in Azure AD for AWS accounts, as described for example in the following document
https://www.cloudreach.com/en/insights/blog/multi-aws-account-federation-with-microsoft-azure-active-directory-as-idp/

Unfortunately, if I specify the AD group I create in the "app_role" block below (see the line in bold in my template below) I get: **Error: "app_role.0.id": this field cannot be set**

Looking at the source on https://github.com/terraform-providers/terraform-provider-azuread/blob/1fcdc7197dbf0303d6111d0663c04aa56c3e9684/azuread/resource_application.go (line 103-106) I see that it is "computed", which I guess is why it can't be set manually.

Should the logic here not rather be that if "id" is provided in the terraform template then that id should be used, instead of the computed one?

[root@5578ed8c70dd /]# terraform -v
Terraform v0.12.9
+ provider.aws v2.28.1
+ provider.azuread v0.6.0

=====
provider "azuread" {}
provider "aws" {region = "eu-west-1"}

variable "aws-account-id" {}
variable "aws-role-name" {}

locals {
   group = "AWS-${var.aws-account-id}-${var.aws-role-name}"
}

resource "azuread_group" "example" {
   name = "${local.group}"
}

  name                       = "AWS-${var.aws-account-id}"
  homepage                   = "https://signin.aws.amazon.com/all_role id "
  identifier_uris            = ["https://signin.aws.amazon.com/saml#${var.aws-account-id}"]
  reply_urls                 = ["https://signin.aws.amazon.com/saml"]
  available_to_other_tenants = false
  oauth2_allow_implicit_flow = true
  type                       = "webapp/api"

  required_resource_access {
    resource_app_id = "00000003-0000-0000-c000-000000000000"

    resource_access {
      id   = "${azuread_group.example.id}"
      type = "Role"
    }
  }

  app_role {
      allowed_member_types = [ "User" ]
      description = "Role mapped from AD group AWS-${var.aws-account-id}-${var.aws-role-name}"
      display_name = "10x-test-adminaccess,Azure-AD-2"
      is_enabled = true
**#      id   = "${azuread_group.example.id}"**
      value = "arn:aws:iam::${var.aws-account-id}:role/${var.aws-role-name},arn:aws:iam::${var.aws-account-id}:saml-provider/Azure-AD-2"
  }
}

resource "azuread_service_principal" "example" {
   application_id = azuread_application.example.application_id
   tags = [
    "WindowsAzureActiveDirectoryIntegratedApp",
  ]
}

manicminer:
Allow `azuread_application.app_role.id` to be specified manually. This is an accepted use case since the API permits it, and is required for some advanced scenarios where UUIDs are carefully coordinated.

Fixes #150


manicminer:
Hitting some seesawing test failures on this, will update when resolved.

manicminer:
Closing in favor of #298

PR #256, opened 2020-05-25
Allow for graceful handling of errored users within users data source

patrickmarabeas:
Return all user data in `users` parameter aray

* Add users parameter - an array of users which mirrors that of the user
data source. This allows for additional validation of users - such as
checking whether the account is enabled.

Allow user lookup to gracefully fail

* Add graceful_errors parameter to allow user lookup to gracefully fail.
This allows only users which were successfully looked up to return. The
data source will still result in an error if no users were returned.

Documentation updated to include the two new parameters.
 
Add missing MailNickname validation logic

### DIFF
--- a/azuread/data_users.go
+++ b/azuread/data_users.go
@@ -139,7 +139,7 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
 				if ignoreMissing && ar.ResponseWasNotFound(u.Response) {
 					break
 				}
-				return fmt.Errorf("Error making Read request on AzureAD User with ID %q: %+v", v.(string), err)
+				return fmt.Errorf("making Read request on AzureAD User with ID %q: %+v", v.(string), err)
 			}
 			users = append(users, &u)
 		}
@@ -151,7 +151,7 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
 				if ignoreMissing && ar.ResponseWasNotFound(u.Response) {
 					break
 				}
-				return fmt.Errorf("Error finding Azure AD User with object ID %q: %+v", v.(string), err)
+				return fmt.Errorf("finding Azure AD User with object ID %q: %+v", v.(string), err)
 			}
 			users = append(users, u)
 		}
@@ -163,52 +163,51 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
 				if ignoreMissing && ar.ResponseWasNotFound(u.Response) {
 					break
 				}
-				return fmt.Errorf("Error finding Azure AD User with email alias %q: %+v", v.(string), err)
+				return fmt.Errorf("finding Azure AD User with email alias %q: %+v", v.(string), err)
 			}
 			users = append(users, u)
 		}
 	}
 
 	if !ignoreMissing && len(users) != expectedCount {
-		return fmt.Errorf("Unexpected number of users returned (%d != %d)", len(users), expectedCount)
+		return fmt.Errorf("unexpected number of users returned (%d != %d)", len(users), expectedCount)
 	}
 
-	if ignoreMissing && len(users) == 0 {
-		return fmt.Errorf("No users were returned")
-	}
+	// TODO: consider disallowing no results in v1.0
+	//if len(users) == 0 {
+	//	return fmt.Errorf("no users were returned")
+	//}
 
 	upns := make([]string, 0, len(users))
 	oids := make([]string, 0, len(users))
 	mailNicknames := make([]string, 0, len(users))
-	var userList []map[string]interface{}
+	userList := make([]map[string]interface{}, 0, len(users))
 	for _, u := range users {
 		if u.ObjectID == nil || u.UserPrincipalName == nil {
-			return fmt.Errorf("User with nil ObjectId or UPN was found: %v", u)
+			return fmt.Errorf("user with nil ObjectId or UPN was found: %v", u)
 		}
 
-		if u.ObjectID != nil || u.UserPrincipalName != nil {
-			oids = append(oids, *u.ObjectID)
-			upns = append(upns, *u.UserPrincipalName)
-			mailNicknames = append(mailNicknames, *u.MailNickname)
+		oids = append(oids, *u.ObjectID)
+		upns = append(upns, *u.UserPrincipalName)
+		mailNicknames = append(mailNicknames, *u.MailNickname)
 
-			user := make(map[string]interface{})
-			user["account_enabled"] = u.AccountEnabled
-			user["display_name"] = u.DisplayName
-			user["immutable_id"] = u.ImmutableID
-			user["mail"] = u.Mail
-			user["mail_nickname"] = u.MailNickname
-			user["object_id"] = u.ObjectID
-			user["onpremises_sam_account_name"] = u.AdditionalProperties["onPremisesSamAccountName"]
-			user["onpremises_user_principal_name"] = u.AdditionalProperties["onPremisesUserPrincipalName"]
-			user["usage_location"] = u.UsageLocation
-			user["user_principal_name"] = u.UserPrincipalName
-			userList = append(userList, user)
-		}
+		user := make(map[string]interface{})
+		user["account_enabled"] = u.AccountEnabled
+		user["display_name"] = u.DisplayName
+		user["immutable_id"] = u.ImmutableID
+		user["mail"] = u.Mail
+		user["mail_nickname"] = u.MailNickname
+		user["object_id"] = u.ObjectID
+		user["onpremises_sam_account_name"] = u.AdditionalProperties["onPremisesSamAccountName"]
+		user["onpremises_user_principal_name"] = u.AdditionalProperties["onPremisesUserPrincipalName"]
+		user["usage_location"] = u.UsageLocation
+		user["user_principal_name"] = u.UserPrincipalName
+		userList = append(userList, user)
 	}
 
 	h := sha1.New()
 	if _, err := h.Write([]byte(strings.Join(upns, "-"))); err != nil {
-		return fmt.Errorf("Unable to compute hash for upns: %v", err)
+		return fmt.Errorf("unable to compute hash for UPNs: %v", err)
 	}
 
 	d.SetId("users#" + base64.URLEncoding.EncodeToString(h.Sum(nil)))
--- a/azuread/data_users.go
+++ b/azuread/data_users.go
@@ -26,8 +26,8 @@ func dataUsers() *schema.Resource {
 			"object_ids": {
 				Type:         schema.TypeList,
 				Optional:     true,
-				Computed:      true,
-				ConflictsWith: []string{"user_principal_names"},
+				Computed:     true,
+				ExactlyOneOf: []string{"object_ids", "user_principal_names", "mail_nicknames"},
 				Elem: &schema.Schema{
 					Type:         schema.TypeString,
 					ValidateFunc: validate.UUID,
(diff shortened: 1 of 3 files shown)


manicminer:
Hi @patrickmarabeas, thanks for this change suggestion. I'm looking to add similar functionality to the `azuread_users` data source (#258), however I'm wary of adding this to a singular data source as we'd usually expect a data source to always err when a resource isn't found. Would you be able to make use of #258 in your configuration?

patrickmarabeas:
@manicminer that data source would also need to be uplifted to gracefully fail when a user being queried returns an error. I'm certainly happy to expand the scope of this PR to provide graceful failure to multiple data sources, but would like to confirm the approach before spending the time.

manicminer:
@patrickmarabeas thanks for replying, I got mixed up and see now your change is quite different - sorry about that.

If I'm honest, I'm not sure we should enable that behavior in a data source as it abstracts from API semantics and may arguably go against provider design principles. However, if we did enable this, I also don't think we could implement it reliably. There are cases where the API returns 404 for a resource that does or should exist - due to replication delays or other circumstances. In such events, we couldn't distinguish between a genuine error and one the operator wishes to ignore.

patrickmarabeas:
@manicminer We are simply looking for graceful failure - regardless of the error. A renaming of the parameters would probably help this intent. But basically, if you opt into graceful failure you are expecting this result. This is certainly not something that should be enabled by default.

In our case, we are looking for accounts to be gracefully and automatically removed from teams if the user data source returns an error. On/off-boarding of accounts in AzureAD is disconnected from the management of teams and we are looking to solve the blocker of Terraform failing when accounts no longer exist. We were hoping that accounts were simply being disabled (which we can check for via the `account_enabled` prop) - but this is only true for AD.

tombuildsstuff:
@patrickmarabeas it's worth mentioning there's a couple of conventions in Terraform for Data Sources:

* Singular Data Sources (e.g. `azuread_user`) should fail if the specified resource can't be found
* Plural Data Sources (e.g. `azuread_users`) should fail if 0 items can be found

As such in this instance I'm wondering if the `azuread_users` data source should be updated to support passing in a list of email addresses (or similar) identifiers (potentially with an "enabled" filter) - which should solve this the other way.

Whilst we've (unintentionally) had some Data Sources in the past which didn't raise an error when the resource wasn't found - we ended up reverting these to bring this into line with Terraform, as such I think we'd be better to add this to the Plural Data Source here rather than the Singular.

What do you think? :)

patrickmarabeas:
@tombuildsstuff I'm happy with this approach. I will migrate the changes across to the `azuread_users` data source allowing for graceful handling of errors - **even in the case of all accounts erroring**.

tombuildsstuff:
@patrickmarabeas to match the behaviour of the rest of Terraform, we'd still need this to error if no accounts are returned from the `azurerm_users` data source. That said the Azure API should still return accounts when they're disabled - presumably just with the "Disabled/Enabled" flag set to "Disabled" - so are these accounts being deleted?

patrickmarabeas:
Yes, the accounts are being deleted.

Having thought on it, the behaviour of an error when no users are returned makes sense - as the group would have no members. Probably a worthy time to call it out.

I will follow this logic.

patrickmarabeas:
@manicminer @tombuildsstuff 

`graceful_errors` parameter added. Data source will error if no users are returned, however.

`users` parameter added which mirrors the `user` data source.

`account_error` was not required, as it's obviously filtered out on error.

azuread/data_users.go
```diff
@@ -54,6 +54,70 @@ func dataUsers() *schema.Resource {
 					ValidateFunc: validate.NoEmptyStrings,
 				},
 			},
+
+			"graceful_errors": {
```
manicminer, line 58: To be more explicit, could we rename this property to `ignore_missing`?

azuread/data_users.go
```diff
@@ -62,54 +126,83 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
...
 
+	gracefulError := d.Get("graceful_errors").(bool)
 	if upns, ok := d.Get("user_principal_names").([]interface{}); ok && len(upns) > 0 {
 		expectedCount = len(upns)
 		for _, v := range upns {
-			resp, err := client.Get(ctx, v.(string))
+			u, err := client.Get(ctx, v.(string))
 			if err != nil {
-				return fmt.Errorf("Error making Read request on AzureAD User with ID %q: %+v", v.(string), err)
+				if !gracefulError {
```
manicminer, line 138: Can we also check the response was a 404, to make sure we aren't ignoring other errors?

e.g.
```hcl
if !gracefulError && ar.ResponseWasNotFound(u.Response) {
	return fmt.Errorf("...")
}
```

azuread/data_users.go
```diff
@@ -62,54 +126,83 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
...
-			users = append(users, resp)
+			users = append(users, &u)
 		}
 	} else if oids, ok := d.Get("object_ids").([]interface{}); ok && len(oids) > 0 {
 		expectedCount = len(oids)
 		for _, v := range oids {
 			u, err := graph.UserGetByObjectId(&client, ctx, v.(string))
 			if err != nil {
-				return fmt.Errorf("Error finding Azure AD User with object ID %q: %+v", v.(string), err)
+				if !gracefulError {
```
manicminer, line 150: Can we have a 404 check here too?

azuread/data_users.go
```diff
@@ -62,54 +126,83 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
...
-			users = append(users, *u)
+			users = append(users, u)
 		}
 	} else if mailNicknames, ok := d.Get("mail_nicknames").([]interface{}); ok && len(mailNicknames) > 0 {
 		expectedCount = len(mailNicknames)
 		for _, v := range mailNicknames {
 			u, err := graph.UserGetByMailNickname(&client, ctx, v.(string))
 			if err != nil {
-				return fmt.Errorf("Error finding Azure AD User with email alias %q: %+v", v.(string), err)
+				if !gracefulError {
```
manicminer, line 162: Another 404 check :)

azuread/data_users.go
```diff
@@ -62,54 +126,83 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
...
+		return fmt.Errorf("No users were returned")
+	}
+
 	upns := make([]string, 0, len(users))
 	oids := make([]string, 0, len(users))
 	mailNicknames := make([]string, 0, len(users))
+	var userList []map[string]interface{}
 	for _, u := range users {
-		if u.ObjectID == nil || u.UserPrincipalName == nil {
+		if !gracefulError && (u.ObjectID == nil || u.UserPrincipalName == nil || u.MailNickname == nil) {
```
manicminer, line 184: Did you find it necessary to skip this error condition when testing? I'm thinking we may want to err here in every case since it indicates a bad API response. Also I think `MailNickname` might be optional for some accounts so it being nil could be ok?

azuread/data_users.go
```diff
@@ -62,54 +126,83 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
...
 			return fmt.Errorf("User with nil ObjectId or UPN was found: %v", u)
 		}
 
-		oids = append(oids, *u.ObjectID)
-		upns = append(upns, *u.UserPrincipalName)
-		mailNicknames = append(mailNicknames, *u.MailNickname)
+		if u.ObjectID != nil || u.UserPrincipalName != nil || u.MailNickname != nil {
```
manicminer, line 188: Same question(s) as per L184

website/docs/d/users.html.markdown
```diff
@@ -31,6 +31,8 @@ The following arguments are supported:
 
 * `mail_nicknames` - (Optional) The email aliases of the Azure AD Users.
 
+* `graceful_errors` - (Optional) Whether Terraform will return an error if any user lookup fails. An error will still occur if no users are returned.
```
manicminer, line 34: In addition to the property rename, could we word this:

> (Optional) Ignore missing users and return users that were found. The data source will still fail if no users are found. Defaults to `false`.

Also can we shift this down to L36 (underneath the _NOTE_ about the required search properties)

azuread/data_users.go
```diff
@@ -26,7 +26,7 @@ func dataUsers() *schema.Resource {
 				Type:          schema.TypeList,
 				Optional:      true,
 				Computed:      true,
-				ConflictsWith: []string{"user_principal_names"},
+				ConflictsWith: []string{"user_principal_names", "mail_nicknames"},
```
manicminer, line 29: As a tangential change, would you mind changing the `ConflictsWith` field to `ExactlyOneOf`? This is a general improvement that would be nice to add at the same time :)

azuread/data_users.go
patrickmarabeas: Ah, didn't realise that about `MailNickname`. Yes this can be reverted.

patrickmarabeas:
Changes have been amended and branch rebased.

azuread/data_users.go
```diff
@@ -62,54 +127,83 @@ func dataSourceUsersRead(d *schema.ResourceData, meta interface{}) error {
...
 
+	ignoreMissing := d.Get("ignore_missing").(bool)
 	if upns, ok := d.Get("user_principal_names").([]interface{}); ok && len(upns) > 0 {
 		expectedCount = len(upns)
 		for _, v := range upns {
-			resp, err := client.Get(ctx, v.(string))
+			u, err := client.Get(ctx, v.(string))
 			if err != nil {
-				return fmt.Errorf("Error making Read request on AzureAD User with ID %q: %+v", v.(string), err)
+				if !ignoreMissing && ar.ResponseWasNotFound(u.Response) {
```
manicminer, line 139: @patrickmarabeas Thanks for making those changes. I realize I didn't think out the logic here - my intention was to have the loop break if we're ignoring missing users and the error was a 404. Sorry about that, how about this instead?

```hcl
if ignoreMissing && ar.ResponseWasNotFound(u.Response) {
	break
}
return fmt.Errorf("Error making Read request on AzureAD User with ID %q: %+v", v.(string), err)
```

patrickmarabeas: Updated these blocks.

patrickmarabeas:
@manicminer I don't have anyway to run the tests, so if you could do this that would be fantastic. I've given you collaborator access on my fork.

manicminer:
I've added a test for the `ignoreMissing` attribute, and altered behavior slightly to allow returning no results in all cases, as this is probably expected given the current behavior of the data source (with a note to consider changing this in future).

Since I've touched the code, I'm opening for review from another contributor :)

website/docs/d/users.html.markdown
```diff
@@ -33,10 +33,29 @@ The following arguments are supported:
...
+
+The `user` object consists of:
+
+* `id` - The Object ID of the Azure AD User.
+* `user_principal_name` - The User Principal Name of the Azure AD User.
+* `account_enabled` - `True` if the account is enabled; otherwise `False`.
+* `display_name` - The Display Name of the Azure AD User.
+* `mail` - The primary email address of the Azure AD User.
+* `mail_nickname` - The email alias of the Azure AD User.
+* `mail_nickname` - The email alias of the Azure AD User.
```
tracypholmes, line 57: It looks like `mail_nickname` was duplicated.

manicminer: Good spot, thanks!

patrickmarabeas:
@manicminer 

Getting an error when testing this now.

```
panic: runtime error: invalid memory address or nil pointer dereference
..
2020-06-30T18:33:37.488+1000 [DEBUG] plugin.terraform-provider-azuread_v0.11.0: 	/Users/marabeap/Code/github-com/terraform-providers/terraform-provider-azuread/azuread/data_users.go:163 +0x1b27
```

Which is this block:

manicminer:
@patrickmarabeas Can you share configuration that reproduces the panic? Tests seem to work for me.

manicminer:
I found the issue, will work up a fix shortly.

manicminer:
@patrickmarabeas this should fix it: #289 

Thanks for testing and reporting!

Issue #145, opened 2019-09-03
Add support for optionalClaims in azuread_application


rcurd:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I would like to set the optionalClaims field in Azure AD Application Registrations via Terraform. Currently this does not appear to be supported.

### New or Affected Resource(s)

azuread_application

* azuread_XXXXX

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims

* #0000

katbyte:
(maybe good first issue)

supported by the SDK with `OptionalClaims` property

RavulaChetan:
Any new update on this?

manicminer:
This has been added with #261 and we'll include it in the next release of the provider.

PR #260, opened 2020-05-29
azuread_application: support optional_claims

manicminer:
Support configuration of optional claims as described [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-optional-claims) and export same attriobutes with data source.

Does not yet support optional claims for SAML tokens due to SDK bug: https://github.com/Azure/azure-sdk-for-go/issues/9714

Fixes: #145


manicminer:
Tests pass
Screenshot 2020-06-02 23 15 58

PR #264, opened 2020-06-04
New ID format for azuread_application_certificate, azuread_application_password, azuread_service_principal_certificate, azuread_service_principal_password

manicminer:
* New ID format inserts a string corresponding to resource type
* Should minimise risk of UUID collision
* FIx a bug in `azuread_application` where possibly nil optionalClaims was not handled


manicminer:
Tests pass
Screenshot 2020-06-05 17 12 16
Screenshot 2020-06-05 17 12 03

Issue #232, opened 2020-03-23
Document minimum permissions required for creating Service Principals


bhummerstone:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

After a lot of trial and error, I've found that the minimum permissions required for creating/deleting Service Principals is Application.ReadWrite.OwnedBy under API permissions; see the below screenshot:

![image](https://user-images.githubusercontent.com/13648840/77326708-1c4f0100-6d12-11ea-8a43-ef22bffab068.png)

However, this doesn't appear to be documented anywhere, and the recommended permissions for the Terraform service principal are User/Company Administrator, which is overkill just for Service Principal management.

### New or Affected Resource(s)

* azuread_service_principal

PR #271, opened 2020-06-11
Stop requiring subscription_id and update configuration guides

manicminer:
Stop requiring a subscription ID to be configured (though still allow it for compatibility)
  - See 44829f2621da31096fe4b06d3ca810631dc078b2

Configuration guide updates
  - Walkthroughs reflect current portal experience
  - Amend guidance on AAD permissions
    - Document some common AAD roles and API scopes
    - Note on subscription ID
  - Remove references to configuring subscriptions

Also fix up certificate related tests by shortening validity period to 6 months

**DEPRECATION WARNING**

The `subscription_id` attribute will be removed from the `azuread_client_config` data source in v1.0. The subscription ID here is only an echo of the subscription ID supplied by the user and there is no possible way to use it currently elsewhere in the provider. Configurations can switch over to the `azurerm_client_config` data source.

Closes: #232

### DIFF
--- a/azuread/data_client_config.go
+++ b/azuread/data_client_config.go
@@ -27,8 +27,9 @@ func dataClientConfig() *schema.Resource {
 			},
 
 			"subscription_id": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:       schema.TypeString,
+				Computed:   true,
+				Deprecated: fmt.Sprintf("The %q attribute will be removed in version 1.0 of the provider. If you are using this attribute, you should instead use the %q data source from the AzureRM provider", "subscription_id", "azurerm_client_config"),
 			},
 
 			"object_id": {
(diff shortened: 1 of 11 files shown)


manicminer:
OK I think that's ready now! 😊

manicminer:
One more rebase to split a commit 😅

azuread/data_client_config.go
```diff
@@ -26,11 +26,6 @@ func dataClientConfig() *schema.Resource {
 				Computed: true,
 			},
 
-			"subscription_id": {
```
katbyte, line 29: I think it would be best to deprecate this and remove it from all documentation & examples so we don't break people's config - we can remove it in 1.0

manicminer: Thanks, that's a good shout. I've reworked it so that for existing configurations, where a subscription ID is configured for the provider, it's returned by `data.azuread_client_config`. For newer configurations where it's omitted, the data source attribute is a blank string.

manicminer:
> thanks @manicminer! overall this is good but i think we should keep the subscription property in the data source until 1.0 - can deprecate it for now and remove it then.

Sounds good. I've reworked it so that for existing configurations, where a subscription ID is configured for the provider, it's returned by data.azuread_client_config. For newer configurations where it's omitted, the data source attribute is a blank string. Do you see any problems with doing it this way?

manicminer:
Tests pass

Screenshot 2020-06-15 17 14 18

Issue #268, opened 2020-06-09
azuread_application: homepage can't be blank


magne:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v0.12.26
+ provider.azuread v0.10.0

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
resource "azuread_application" "api" {
  name     = "My App"
  homepage = ""

  oauth2_permissions {
    type                       = "User"
    value                      = "API.All"
    admin_consent_description  = "Access to the API."
    admin_consent_display_name = "Access to the API."
  }
}
```

### Expected Behavior

I expected the application to be created with a blank homepage (I'm creating an internal API app registration, and it doesn't have an obvious homepage).

### Actual Behavior

Terraform reports an error:
```
Error: expected "homepage" url to not be empty
```

Also, if I try to leave out `homepage`, terraform fails with:
```
Error: graphrbac.ApplicationsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="Unknown" Message="Unknown service error" Details=[{"odata.error":{"code":"Request_BadRequest","date":"2020-06-09T09:35:04","message":{"lang":"en","value":"Url is malformed paramName: WwwHomepage, paramValue: https://My App, objectType: Microsoft.Online.DirectoryServices.Application"},"requestId":"07137f59-663e-4292-bc3f-28ddddc14628","values":[{"item":"PropertyName","value":"homepage"},{"item":"PropertyErrorCode","value":"NotAWellFormedUrl"}]}}]
```
This happens because of the default value of `https://{name}` for `homepage`, since my app registration has a name that is unsuitable for an url.

The obvious workaround is of course to set `homepage` to a valid url, but this should not be necessary.

### Steps to Reproduce

1. `terraform init`
1. `terraform apply`

manicminer:
[...]

PR #278, opened 2020-06-16
azuread_application: no longer derive a default value for the homepage property

manicminer:
- The homepage property (now known as 'signinUrl') used to be a required
  field with a default value when using the Azure Portal.
- It's no longer presented to the user when creating an application, and
  defaults to being unset (`null` in the application manifest)
- Also, it cannot currently be unset by the SDK once set. See https://github.com/Azure/azure-sdk-for-go/issues/10463

Mitigates, but does not solve: #268 
Related: https://github.com/Azure/azure-sdk-for-go/issues/10463

### DIFF
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -58,7 +58,6 @@ func resourceApplication() *schema.Resource {
 			"homepage": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				Computed:     true,
 				ValidateFunc: validate.URLIsHTTPOrHTTPS,
 			},
 
@@ -316,11 +315,6 @@ func resourceApplicationCreate(d *schema.ResourceData, meta interface{}) error {
 
 	if v, ok := d.GetOk("homepage"); ok {
 		properties.Homepage = p.StringI(v)
-	} else {
-		// continue to automatically set the homepage with the type is not native
-		if appType != "native" {
-			properties.Homepage = p.String(fmt.Sprintf("https://%s", name))
-		}
 	}
 
 	if v, ok := d.GetOk("logout_url"); ok {
--- a/azuread/resource_application_test.go
+++ b/azuread/resource_application_test.go
@@ -28,7 +28,6 @@ func TestAccAzureADApplication_basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testCheckADApplicationExists(resourceName),
 					resource.TestCheckResourceAttr(resourceName, "name", appName),
-					resource.TestCheckResourceAttr(resourceName, "homepage", fmt.Sprintf("https://%s", appName)),
 					resource.TestCheckResourceAttr(resourceName, "oauth2_allow_implicit_flow", "false"),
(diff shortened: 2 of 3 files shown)


azuread/resource_application_test.go
```diff
@@ -642,18 +638,19 @@ func testAccADApplication_basicEmpty(ri int) string {
 	return fmt.Sprintf(`
 resource "azuread_application" "test" {
   name                    = "acctest-APP-%[1]d"
+  homepage                = "https://homepage-%[1]d" // Cannot be unset once set, see https://github.com/Azure/azure-sdk-for-go/issues/10463
```
katbyte, line 641: Should it be force new then?

azuread/resource_application.go
```diff
@@ -316,11 +315,6 @@ func resourceApplicationCreate(d *schema.ResourceData, meta interface{}) error {
 
 	if v, ok := d.GetOk("homepage"); ok {
 		properties.Homepage = p.StringI(v)
-	} else {
-		// continue to automatically set the homepage with the type is not native
-		if appType != "native" {
-			properties.Homepage = p.String(fmt.Sprintf("https://%s", name))
-		}
```
katbyte, line 323: this is changing existing behaviour.. should this not wait until 1.0 to change as i will probably break existing configs?

website/docs/r/application.html.markdown
```diff
@@ -109,7 +109,9 @@ The following arguments are supported:
 
 * `name` - (Required) The display name for the application.
 
-* `homepage` - (optional) The URL to the application's home page. If no homepage is specified this defaults to `https://{name}`.
+* `homepage` - (optional) The URL to the application's home page. This property is now also known as `signInUrl`.
```
katbyte, line 112: i don't see a property known as signin url, so maybe this would be better worder as
```suggestion
* `homepage` - (optional) The URL to the application's home page. This property is also referred to as `signInUrl` in [portal?].
```

manicminer: Ah right, I see the Branding blade does actually still show "Home page URL" even though the underlying property changed. I'll amend this.

azuread/resource_application_test.go
manicminer: ForceNew would make sense but there is a case where you can go from no homepage to some homepage, e.g. from
```hcl
  name = "test"
}
```
to
```hcl
  name = "test"
  homepage = "https://test"
}
```
which this works, and I think we should enable that.

azuread/resource_application.go
manicminer: Agreed, I will bump to v1.0 👍

azuread/resource_application_test.go
katbyte: we could use customize diff to conditionally force new?

katbyte:
Still need to revert the homepage generation change to 1.0

manicminer:
Given the underlying SDK issue, we'll bump this PR to 1.0 since we can't improve things without making the breaking change to the `homepage` property 👍

manicminer:
Closing this as the change is included in #298

Issue #182, opened 2019-11-27
azuread_group can create multiple groups with the same name


simongottschlag:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
```
Terraform v0.12.16
+ provider.azuread v0.7.0
+ provider.azurerm v1.37.0
```

### Affected Resource(s)
* `azuread_group`

### Terraform Configuration Files
```hcl
resource "azuread_group" "aadGroupOwner" {
    name        = "aad-sec-u-dev-test-owner"
}
```

### Expected Behavior
Create a group if another one with the same name does not exists.

### Actual Behavior
Creates a group even if there's another one with the exact same name but different ID.

### Steps to Reproduce
Create the group in two different states.

### Important Factoids
When using Azure Portal, it's not possible to create a group if there's already an existing one.

katbyte:
Hi @simongottschlag ,

Unfortunately the API works on GUID IDs. For now you can import the group via the app ID. In the future we could add a flag `check_name_availiblity`/`prevent_duplicate_names` which checks if the name is in use during creation and fails if it is.

simongottschlag:
@katbyte thank you for the clarification! Not a big issue, as long as we don't loose our state. Looking forward to the new release!

Issue #200, opened 2020-01-13
duplicate azuread_application resources are created when redeploying


diirib:
When creating an azuread_application resource, we define the display name. If we try to redeploy the same resource, since the applicationId is created automatically and is different from the one deployed the first time, terraform creates a new application with the same name but a different applicationId.

I would expect to at least have a parameter to specify if I want a new resource to be created or I'm just redeploying. 

Is the applicationId the only way to check if there's an existing application?

katbyte:
Hi @diirib,

Unfortunately the API works on GUID IDs. For now you can import the application via the app ID. In the future we could add a flag `check_name_availiblity`/`prevent_duplicate_names` which checks if the name is in use during creation and fails if it is.

PR #279, opened 2020-06-16
azuread_application & azuread_group: Duplicate name detection

manicminer:
- New property `prevent_duplicate_names`, defaults to false
- When true, searches for an existing application / group and errs if one found having the same name
- For applications, also performs the check on update

Closes: #182, #200

### DIFF
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -285,6 +285,11 @@ func resourceApplication() *schema.Resource {
 					},
 				},
 			},
+			"prevent_duplicate_names": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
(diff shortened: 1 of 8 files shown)


azuread/resource_application.go
```diff
@@ -904,3 +928,31 @@ func adApplicationSetOwnersTo(client graphrbac.ApplicationsClient, ctx context.C
 
 	return nil
 }
+
+func aadApplicationFindByName(client graphrbac.ApplicationsClient, ctx context.Context, name string) (*graphrbac.Application, error) {
```
katbyte, line 932: i think this would make sense in the graph helper package as graph.ApplicationFindByName?

azuread/resource_application.go
```diff
@@ -904,3 +928,31 @@ func adApplicationSetOwnersTo(client graphrbac.ApplicationsClient, ctx context.C
...
+	for _, app := range resp.Values() {
+		if *app.DisplayName == name {
+			return &app, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func aadApplicationCheckNameAvailability(client graphrbac.ApplicationsClient, ctx context.Context, name string) error {
```
katbyte, line 949: same here

azuread/resource_group.go
```diff
@@ -238,3 +255,31 @@ func resourceGroupDelete(d *schema.ResourceData, meta interface{}) error {
 
 	return nil
 }
+
+func aadGroupFindByName(client graphrbac.GroupsClient, ctx context.Context, name string) (*graphrbac.ADGroup, error) {
```
katbyte, line 259: ditto for these two functions

azuread/resource_application.go
manicminer: Yeah absolutely, I'll move them there 👍

manicminer:
Tests pass:

Screenshot 2020-06-23 03 08 17
Screenshot 2020-06-23 03 08 02

PR #281, opened 2020-06-17
azuread_user: Add job info and contact info fields

nauxliu:


azuread/resource_user.go
```diff
@@ -101,6 +101,62 @@ func resourceUser() *schema.Resource {
 					"Required for users that will be assigned licenses due to legal requirement to check for availability of services in countries. " +
 					"Examples include: `NO`, `JP`, and `GB`. Not nullable.",
 			},
+
```
manicminer, line 104: These properties should probably all be computed

azuread/resource_user.go
```diff
@@ -152,7 +208,7 @@ func resourceUserCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Error waiting for User (%s) with ObjectId %q: %+v", upn, *user.ObjectID, err)
 	}
 
-	return resourceUserRead(d, meta)
+	return resourceUserUpdate(d, meta)
```
manicminer, line 211: Is it necessary to update the user in order to add AdditionalProperties? It would be preferable to set these in the create function

azuread/resource_user.go
```diff
@@ -220,6 +323,15 @@ func resourceUserRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("usage_location", user.UsageLocation)
 	d.Set("immutable_id", user.ImmutableID)
 
+	d.Set("job_title", user.AdditionalProperties["jobTitle"])
```
manicminer, line 326: I think it's worth testing each of these keys exist before referencing them.

azuread/resource_user.go
nauxliu: Why they should be computed as they are not actually computed, By setting them as computed, We'll lose the ability to clear the value of the attributes.

manicminer: That's true, however without the Computed flag, these properties will be cleared if they are not specified in configuration, potentialy wiping all of those fields for existing users. For example:

Screenshot 2020-06-23 01 19 23

nauxliu: I prefer not to mark them as computed because the ability to clear up data is essential. For users who don't want to manage some attributes with terraform, there is a workaround to use `lifecycle.ignore_changes`。

katbyte: @nauxliu - these should be computed. If a user wishes to clear them couldn't they just set it to ""?

nauxliu: I've tried to pass `""`  to computed attributes and some other workaround, there is no way to clear computed data.

nauxliu: @manicminer what do you think?

manicminer: @nauxliu Sorry for the delay, I've done some testing on this and spoken with the SDK team. This is an unexpected SDK behavior, possibly a regression. For now we will have to settle with partial support - where clearing the attributes requires a manual or scripted action. We can unfortunately not consider adding these as non-computed. It creates the unnecessary risk of accidentally wiping these attributes for existing users, in a database that for many organisations where federation is not in use, could be a primary source of truth for this data.

To compound things a little, we are currently readying a v1.0 release of the AzureAD provider, and won't releasing any new features in the current 0.x tree. We can of course look at adding these attributes into a post-1.0 release but this PR would need to be rebased as many files have been moved around, in addition to the other changes previously requested.

nauxliu:
Looks like `TestCheckResourceAttr` assertions have been removed from v1.0,  Should I add these attr tests back to test new attributes? @manicminer

manicminer:
@nauxliu Thanks for following up. It's not necessary to check these attributes unless they are only computed (not user specifiable) - an import step will take care of that.

If you can address the above comments, and rebase your changes on master to use the new file tree and latest changes in v1.0, we can look at getting this merged.

manicminer:
@nauxliu Would you mind if I picked this up and rebased/reviewed to get it merged for v1.1?

nauxliu:
@manicminer Sure, thanks. I'm too busy recently, don't have much time to work on this.

manicminer:
Closing this in favor of #351 since this was opened pre-1.0

Issue #246, opened 2020-05-05
data.azuread_group: name should be case insensitive


Tbohunek:
Hi,

Why did you opt to make Azure AD group lookup case sensitive in https://github.com/terraform-providers/terraform-provider-azuread/commit/b82e58ff53ddce5be32c6d55e07bdd67665ed8c3 ?

Azure AD is in essence case-insensitive so there's no reason for such restriction.
Instead this makes it difficult to use common variables such as `environment=dev` if someone named the AD group `group_Dev`. Same for Service Principals.

Thanks for explanation,
Tomas

katbyte:
Hey @Tbohunek, we default to case sensitivity for many reasons and have no plans to change this. One reason is if someone wanted to change the casing we wouldn't be able to. You can see the comment that prompted that to be case sensitive here: https://github.com/terraform-providers/terraform-provider-azuread/pull/14#discussion_r247437359

Tbohunek:
@katbyte Thanks, but... can you at least add an optional parameter that would allow case-insensitive data lookup? I understand one would want case-sensitive with `resource {}` but why to need it with `data {}` is beyond me.

Tbohunek:
Btw thanks for closing the issue instantly without any chance for a discussion. Maybe I misunderstood the linked comment but it tells me the only reason you do case sensitive is because _you always have_, not because it made sense technically nor because customers asked for it.

tombuildsstuff:
hey @Tbohunek 

Thanks for opening this issue.

Whilst certain Azure API's are case insensitive - a much larger proportion are not, meaning that since everything isn't case insensitive - by definition everything has to be treated as case sensitive.

In this instance, I'd agree this makes sense to make the Data Source case-insensitive here (but as you've mentioned leave the Resource case-sensitive) - as such I'm going to re-open this issue to track this feature request.

Thanks!

Tbohunek:
Thanks @tombuildsstuff 
Would that be possible and logical also for `data {}` on Service principals? Or should I open a separate Issue for that?
Thank you

manicminer:
@Tbohunek I've updated the issue title to reflect the change requested. Please go ahead and open another issue if the service principal data source needs looking at too. Thanks!

manicminer:
@Tbohunek I've implemented this in #291, however since this is a potentially breaking change this will only make it into the next major release of the provider.

DanielFrei64:
[...]

PR #291, opened 2020-07-01
data.azuread_group: name argument is now case insensitive

manicminer:
**NOTE: potentially breaking change destined for v1.0**

Fixes: #246


manicminer:
Closing as this change is included in #298

PR #304, opened 2020-08-17
updated to latest azuread and changed existing group error to warning

akingscote:
If youve selected `prevent_duplicate_names` you are expecting there to be an already existing group.
In my scenario, my terraform is being deployed via a CICD pipeline and the error breaks the pipeline.

If you are managing groups in terraform, then the existing group should look exactly like the one you are trying to create.
The only place an error make sense if you are trying to create a group that isnt managed outside of terraform.

If you have further dependencies on a group down the line, then your terraform still wont work, regardless of a warning or an error message.

### DIFF
--- a/azuread/helpers/graph/group.go
+++ b/azuread/helpers/graph/group.go
@@ -221,7 +221,7 @@ func GroupCheckNameAvailability(client graphrbac.GroupsClient, ctx context.Conte
 		return err
 	}
 	if existingGroup != nil {
-		return fmt.Errorf("existing Azure Active Directory Group with name %q (ObjID: %q) was found and `prevent_duplicate_names` was specified", name, *existingGroup.ObjectID)
+		return fmt.Warnf("existing Azure Active Directory Group with name %q (ObjID: %q) was found and `prevent_duplicate_names` was specified", name, *existingGroup.ObjectID)
 	}
 	return nil
 }


manicminer:
Hi @akingscote, thanks for opening this PR. Unfortunately we don't support warnings so we'll be unable to merge this change as it stands. Can you explain your scenario a little more? It sounds like you might want to omit the `prevent_duplicate_names` attribute (or set it to false). If you are looking to manage an existing group, you should import it into your state using `terraform import`.

akingscote:
@manicminer 
Yeah i have `prevent_duplicate_names` set, but i dont want it to error as it breaks my CICD pipeline.

I am creating the group via CICD, but the next time i run the job it will error on the Azure AD group section. I am using hundreds of different modules and this is the only one which has this goofy functionality of not remembering the state.
I have a state file and there is a entry for the group, but it just creates a new group each time.

So i added the `prevent_duplicate_names` flag and it dosent try to create the group but errors. I just want it to either recognise that it has a state entry for this resource or just fail quietly.

Im deploying a hub and spoke architecture. Every time i deploy a spoke, one of the many things i am doing is to create a group with a user in it. If i redeploy the an existing spoke, i expect the job to succeed. It shouldnt create any new resources, but it should recognise that they are all being managed by terraform.
Ive been using terraform for a while now and this is the **only** module i have found which has this functionality. Its very frustrating.

I dont want to overcomplicate things by using datasources, i want the group creation to be managed by terraform. Im very suprised im the first person to bring this up as its not consistent with the wider terraform design?

Not that its particularly relevant, but Im using terragrunt as a wrapper around terraform.
Any help would be greatly appreciated.

I may take a proper look at the code and see why it tries to create a new group every time. Perhaps others want that functionality as the group name isnt a unique id.

manicminer:
Hi @akingscote, thanks for clarifying. I saw and replied to your related issue after my reply above. I'll do some testing with the group resource and see if I can reproduce the issue.

manicminer:
@akingscote could you share a configuration snippet that reproduces the issue? I have been unable to reproduce locally - the group state is remembered correctly between Terraform runs and the group is not recreated (or duplicated) on subsequent runs.

akingscote:
@manicminer Sure, ill do it first thing tomorrow. cheers

akingscote:
@manicminer 
Apologies for wasting your time here, I cannot for the life of me replicate the issue outside my CICD pipeline which hints at a problem on my side somewhere.
Ill keep investigating and if it does turn out to be a terraform/provider problem i will raise a detailed issue. Thanks

manicminer:
@akingscote No problem, glad you got to the bottom of it!

Issue #208, opened 2020-01-19
Cannot create Microsoft Login capable app registration using azuread_application


adrianhall:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v0.12.19
├── provider.azuread ~> 0.7.0

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
resource "azuread_application" "appregistration" {
    name                        = "Title"
    reply_urls                  = [ "msauth://foobar.azurewebsites.net/auth" ]
    available_to_other_tenants  = false
    public_client               = true
    type                        = "native"
}
```

### Expected Behavior

App Registrations provide for three settings on who can log in:

* Accounts in this organizational directory only (Default Directory only - Single tenant)
* Accounts in any organizational directory (Any Azure AD directory - Multitenant)
* Accounts in any organizational directory (Any Azure AD directory - Multitenant) and personal Microsoft accounts (e.g. Skype, Xbox)

I should be able to select which of these I want to choose.

### Actual Behavior

There is no choice provided, and the choice picked is always single tenant.  Once created, there is no ability to select "personal Microsoft accounts" - only single tenant and multi-tenant.

### Steps to Reproduce

1. `terraform apply`

### References

I am actually trying to duplicate the "configure Microsoft Login for authentication" flow documented here: https://docs.microsoft.com/en-us/azure/app-service/configure-authentication-provider-microsoft

phenixdotnet:
Hi, I got the same issue with our terraform and it looks like this provider is using [Azure SDK for Go](https://github.com/Azure/azure-sdk-for-go) which still use the old graph.windows.net API.

As a workaround I use a script which make a curl PATCH on the MS Graph API:
```bash
GRAPH_API_URL="https://graph.microsoft.com/v1.0"
RESPONSE=$(curl --location --silent --request POST "https://login.microsoftonline.com/${tenantId}/oauth2/v2.0/token" \
                --header 'Content-Type: application/x-www-form-urlencoded' \
                --data-urlencode 'grant_type=client_credentials' \
[...]

PR #309, opened 2020-08-20
azuread_application: Adding signInAudience field

RobertvBraam:
I'm trying to solve #208 by adding a `sign_in_audience` field to enable users to declare the audience type when creating an `azuread_application` resource.

As specified in the documentation updating this value on an existing resource is not advised due to different constraints per audience type. Therefor an error will be thrown when updating the `sign_in_audience` field.

The test ran successfully, but I'm not sure my setup was correct so please verify.

A question about testing the changes. Is there a possibility/documentation on how to use the the Terraform provider with the changes locally with my own scripts?

Since this is my first time with `golang`, so please let me know if there are improvements.

### DIFF
--- a/azuread/resource_application.go
+++ b/azuread/resource_application.go
@@ -285,10 +285,18 @@ func resourceApplication() *schema.Resource {
 					},
 				},
 			},
+
 			"prevent_duplicate_names": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
+			},
+
+			"sign_in_audience": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "AzureADMyOrg",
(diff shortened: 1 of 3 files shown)


manicminer:
Hi @RobertvBraam, thanks for submitting this PR. It'd be great to add this property but when I try to set it I get a 400 error.

```json
[{"odata.error":{"code":"Request_BadRequest","date":"2020-08-20T21:37:15","message":{"lang":"en","value":"Property 'signInAudience' is read-only and cannot be set."},"requestId":"79954ce2-1cc8-4601-a1c3-4fc739609b06"}}]
```

Have you been able to successfully set this to a non-default value (`AzureADMultipleOrgs` or `AzureADandPersonalMicrosoftAccount`) when creating an application?

RobertvBraam:
@manicminer thank you for your time to take a look at my pull request!

PR #313, opened 2020-09-04
remove reference to subscription_id in provider doc

benjyiw:
Version 1.0.0 removed the `subscription_id` argument for configuring the azuread provider, which causes an error if you configured that previously. It was a bit confusing to reference the documentation and see that it was still an optional argument.
```
Error: Unsupported argument

  on providers.tf line 37, in provider "azuread":
  37:   subscription_id = var.azure-infra-subscription-id

An argument named "subscription_id" is not expected here.
```

This PR removes the reference to that configuration option in the website doc.


manicminer:
@benjyiw Thanks for this documentation fix. LGTM.

PR #318, opened 2020-09-16
Add group availability data source

patrickmarabeas:
Adds a data source which will cause an error if the specified group name
exists within AzureAD. This resolves the issue of group resource only
failing on apply when using "prevent_duplicate_names".

If there is a better way, I'm all ears.


patrickmarabeas:
This doesn't even work in real world. I'm at a loss how to validate a resource during the plan phase.

patrickmarabeas:
99% sure this is a limitation with Terraform.

Given we are only managing a sub section of our AzureAD instance with Terraform we've fallen into Terraform's blind spot. Looks like this can be solved by tracking group names / id's outside of TF.

Issue #334, opened 2020-10-13
Application identifier_uris from allows array but does not store array in state


MarkDordoy:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform version: 0.12.29
AzureAD provider version : 1.0.0

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files
```hcl
resource "azuread_application" "example" {
  name                       = "example"
  homepage                   = "https://homepage"
  identifier_uris            = ["https://uri","https://uri2"]
  reply_urls                 = ["https://replyurl"]
  available_to_other_tenants = false
  oauth2_allow_implicit_flow = true
  type                       = "webapp/api"
}

resource "azuread_service_principal" "example" {
  application_id               = azuread_application.example.application_id
  app_role_assignment_required = false

  tags = ["WindowsAzureActiveDirectoryIntegratedApp", "WindowsAzureActiveDirectoryCustomSingleSignOnApplication", "WindowsAzureActiveDirectoryGalleryApplicationNonPrimaryV1"]
}
```

### Expected Behavior

I would expect the array of strings allowed for the `identifier_uris` to be stored in state. 

### Actual Behavior

State stores the values and works when only a single value of identifier_uris is set in the application. However when you pass an array terraform state never stores the array. Looking at the backend API, multiple identifier_uris are set on the application, but terraform fails to recognise this and causes it to try and apply again and again.
### Steps to Reproduce

1. `terraform apply`

manicminer:
@MarkDordoy Thanks for reporting. Can you share the output from `terraform apply` that you note is stuck in a persistent diff? `identifierUris` for an application are exposed in the service principal manifest as `servicePrincipalNames` but to my knowledge we don't currently store this in state.

MarkDordoy:
Ive removed anything relating specific to my company and replaced with fake data. 

This is the plan

```hcl
 ~ resource "azuread_application" "application" {
        application_id             = "4bc611ef-7cb1-4ca9-ba0c-5a1f89df4ea0"
        available_to_other_tenants = false
[...]

PR #335, opened 2020-10-14
Update AzureAD Application IdentifierUris to Support list 

MarkDordoy:


manicminer:
@MarkDordoy Thanks for submitting, however unfortunately as mentioned in #334 the behavior seeking to be fixed is upstream and this should probably remain a list.

PR #338, opened 2020-10-21
application: Don't perform duplicate name check on update unless the name was changed

### LINKED ISSUE #310: azuread_application using prevent_duplicate_names does not ignore provisioned app
https://github.com/terraform-providers/terraform-provider-azuread/blob/d2dfb01976da48d3ff14e92b6858c182da6c2536/azuread/helpers/graph/application.go#L288-L290

When utilizing prevent_duplicate_names, the provisioning process will error out after the app has been created and during the application of other values. The check should ignore the provisioned application that it is updating.

manicminer:
Fixes #310

### DIFF
--- a/internal/services/aadgraph/application_resource.go
+++ b/internal/services/aadgraph/application_resource.go
@@ -422,7 +422,7 @@ func applicationResourceUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	name := d.Get("name").(string)
 
-	if d.Get("prevent_duplicate_names").(bool) {
+	if d.HasChange("name") && d.Get("prevent_duplicate_names").(bool) {
 		err := graph.ApplicationCheckNameAvailability(ctx, client, name)
 		if err != nil {
 			return err


tareks:
Can someone please review/approve this? We seem to be running into the same issue while resources are being created/updated.

jackofallops (approved):
Thanks @manicminer - LGTM 👍

scott-doyland-burrows:
Hi,
The problem still seems to exist with provider 1.1.0.
I try to create an app registration which does NOT yet exist using:
```prevent_duplicate_names = true```

and still get the error.  The application ID in the error IS the application ID of the app registration that terraform has partially managed to create before it has the error.
```
Error: existing Application with name "burrows-appreg-scott-cadabraui" (AppID: "a79564a0-72cb-4b2d-855e-83079eb05929") was found and `prevent_duplicate_names` was specified

  on cadabraui.tf line 1, in resource "azuread_application" "cadabraui":
   1: resource "azuread_application" "cadabraui" {


terraform version
Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/azuread v1.1.0
+ provider registry.terraform.io/hashicorp/azurerm v2.37.0
```

PR #344, opened 2020-10-28
Fix missing link in SP configuration documentation

hfurubotten:
There is a missing link in the [Service Principal Configuration documentation](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/guides/service_principal_configuration#method-2-api-access-with-admin-consent), which made it a bit harder to read but still understandable. This change cleans up the broken link format rendered to users and make it more readable and usable.


manicminer:
@hfurubotten Many thanks for this documentation fix! We'll publish this as soon as we can.

Issue #353, opened 2020-11-11
Owners on azuread_application resource can't be removed


ChrisF987:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
```
Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/azuread v1.0.0
```

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

Initial Terraform Apply
```hcl
provider "azuread" {
  version = "1.0.0"
}

resource "azuread_application" "app1" {
  name = "chris-awesome-application"
}

resource "azuread_service_principal" "service_principal_1" {
  application_id = azuread_application.app1.application_id
}

resource "azuread_application" "app2" {
  name   = "MyAzureADApplication"
  owners = [azuread_service_principal.service_principal_1.object_id]
}
```

Second Terraform Apply (owners attribute removed from app2 azuread_application)
```hcl
provider "azuread" {
  version = "1.0.0"
}

resource "azuread_application" "app1" {
  name = "chris-awesome-application"
}

resource "azuread_service_principal" "service_principal_1" {
  application_id = azuread_application.app1.application_id
}

resource "azuread_application" "app2" {
  name   = "MyAzureADApplication"
}
```

### Expected Behavior

When the `owners` attribute is removed from the azuread_application resource the terraform apply should remove the existing owners from the resource. 

### Actual Behavior

After the `owners` attribute is removed from the azuread_application resource, the subsequent terraform plan or apply show no changes.

```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

data.azuread_service_principal.this: Refreshing state... [id=f24c33cf-5514-4a45-8232-77e5d834077c]
azuread_application.this: Refreshing state... [id=cf5e3ca3-9ad1-4026-b945-8062ee7eb14e]

------------------------------------------------------------------------

No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
```
[...]

Issue #352, opened 2020-11-11
Owners on azuread_application resource can't be specified as an empty list


ChrisF987:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
```
Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/azuread v1.0.0
```

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
provider "azuread" {
  version = "1.0.0"
}

resource "azuread_application" "this" {
  name = "MyAzureADApplication"
  owners = []
}
```

### Expected Behavior

Terraform plan and then apply creates the azuread_application resource without any Owners. This is what occurs when the Owners attribute is not supplied. 

### Actual Behavior

terraform plan errors with the following:
```
Error: owners: attribute supports 1 item as a minimum, config has 0 declared

  on main.tf line 5, in resource "azuread_application" "this":
   5: resource "azuread_application" "this" {
```
### Steps to Reproduce

1. `terraform plan`

### References

#353 is another issue I've raised with this `owners` attribute

manicminer:
@ChrisF987 Thanks for reporting this bug. A fix for this is forthcoming in #355

PR #355, opened 2020-11-13
azuread_application: enable removal of owners

manicminer:
Enables removal of owners on existing applications, and creation of applications with no owners (by default a new application gets the creator assigned as a single owner).

Closes: #352 
Closes: #353


katbyte (approved):
LGTM 👍

PR #367, opened 2020-11-26
azuread_application: avoid update after create

### LINKED ISSUE #364: Azure app registration fails when using prevent_duplicate_names
Hi,
The problem still seems to exist with provider 1.1.0.
I try to create an app registration which does NOT yet exist using:
```prevent_duplicate_names = true```

and still get the error.  The application ID in the error IS the application ID of the app registration that terraform has partially managed to create before it has the error.
```
Error: existing Application with name "burrows-appreg-scott-cadabraui" (AppID: "a79564a0-72cb-4b2d-855e-83079eb05929") was found and `prevent_duplicate_names` was specified

  on cadabraui.tf line 1, in resource "azuread_application" "cadabraui":
   1: resource "azuread_application" "cadabraui" {


terraform version
Terraform v0.13.5
+ provider registry.terraform.io/hashicorp/azuread v1.1.0
+ provider registry.terraform.io/hashicorp/azurerm v2.37.0
```

_Originally posted by @scott-doyland-burrows in https://github.com/terraform-providers/terraform-provider-azuread/issues/310#issuecomment-733863600_

manicminer:
There are a set of small bugs which can occur by updating immediately after creating an application, mostly around the reliance on `d.GetChange()`, which is always true in this scenario.

This PR refactors the update logic for the `app_role` and `oauth2_permissions` properties into a helper function, and stops returning `applicationResourceUpdate()` at the end of `applicationResourceCreate()`. Additionally fixes up the setting of `prevent_duplicate_names` on read to avoid unwanted diffs, and adds regression test coverage for that attribute.

Sets `ConfigMode: schema.SchemaConfigModeAttr` on the `app_role` attribute to enable specifying an empty list in order to remove existing app roles. As with the other optional+computed attributes, this only works on update and if a config change is detected. Test also amended for this case.

Resolves #364

### DIFF
--- a/internal/services/aadgraph/application_resource.go
+++ b/internal/services/aadgraph/application_resource.go
@@ -108,9 +108,10 @@ func applicationResource() *schema.Resource {
 			},
 
 			"app_role": {
-				Type:     schema.TypeSet,
-				Optional: true,
-				Computed: true,
+				Type:       schema.TypeSet,
+				Optional:   true,
+				Computed:   true,
+				ConfigMode: schema.SchemaConfigModeAttr,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"id": {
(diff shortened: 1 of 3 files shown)


jackofallops (approved):
Thanks @manicminer - LGTM 👍

manicminer:
Linters passing locally

Acceptance test results:

Screenshot 2020-11-26 at 16 25 13

PR #368, opened 2020-12-01
Terraform Plugin SDK v2 and acceptance test wrappers

manicminer:
- Move to Terraform Plugin SDK v2
- Implement acceptance test wrappers for cleaner tests and great justice
- Handle 400 error on group member removal, which can be considered a 404 in disguise

### Latest Test Results

Screenshot 2021-01-13 at 18 34 25
  
Screenshot 2020-12-07 at 16 29 01


internal/provider/provider.go
```diff
@@ -195,14 +196,18 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
...
+		return buildClient(ctx, p, builder, partnerId)
 	}
 }
 
-func buildClient(p *schema.Provider, b *authentication.Builder, partnerId string) (*clients.AadClient, error) {
+func buildClient(ctx context.Context, p *schema.Provider, b *authentication.Builder, partnerId string) (*clients.AadClient, diag.Diagnostics) {
 	config, err := b.Build()
 	if err != nil {
-		return nil, fmt.Errorf("building AzureAD Client: %s", err)
+		return nil, diag.Diagnostics{diag.Diagnostic{
```
katbyte, line 206: Could we replace these with a helper function?

internal/provider/provider.go
```diff
@@ -211,15 +216,17 @@ func buildClient(p *schema.Provider, b *authentication.Builder, partnerId string
...
+		stopCtx = ctx
 	}
 
-	// replaces the context between tests
-	p.MetaReset = func() error { //nolint unparam
-		client.StopContext = p.StopContext()
-		return nil
+	client, err := clientBuilder.Build(stopCtx)
+	if err != nil {
+		return nil, diag.Diagnostics{diag.Diagnostic{
```
katbyte, line 226: And i think there is a helper function in the SDK so this can. be one line?

internal/services/aadgraph/application_app_role_resource.go
```diff
@@ -96,7 +97,7 @@ func applicationAppRoleResourceCreateUpdate(d *schema.ResourceData, meta interfa
 	} else {
 		rid, err := uuid.GenerateUUID()
 		if err != nil {
-			return fmt.Errorf("generating App Role for Object ID %q: %+v", objectId, err)
+			return tf.ErrorDiag(fmt.Sprintf("Generating App Role for application with object ID %q", objectId), err.Error(), "")
```
katbyte, line 100: I think we could make this a little cleaner, for ones that don't nee a "path"
```suggestion
			return tf.ErrorDiagF(err, "Generating App Role for application with object ID %q", objectId)
```

internal/services/aadgraph/application_app_role_resource.go
```diff
@@ -128,9 +129,9 @@ func applicationAppRoleResourceCreateUpdate(d *schema.ResourceData, meta interfa
 	app, err := client.Get(ctx, id.ObjectId)
 	if err != nil {
 		if utils.ResponseWasNotFound(app.Response) {
-			return fmt.Errorf("Application with ID %q was not found", id.ObjectId)
+			return tf.ErrorDiag(fmt.Sprintf("Application with object ID %q was not found", id.ObjectId), "", "application_object_id")
```
katbyte, line 132: And these could become something like
```suggestion
			return tf.ErrorDiagPathF(nil, "application_object_id", "Application with object ID %q was not found", id.ObjectId))
```

internal/services/aadgraph/application_app_role_resource.go
```diff
@@ -128,9 +129,9 @@ func applicationAppRoleResourceCreateUpdate(d *schema.ResourceData, meta interfa
...
 		}
-		return fmt.Errorf("retrieving Application ID %q: %+v", id.ObjectId, err)
+		return tf.ErrorDiag(fmt.Sprintf("retrieving Application with object ID %q", id.ObjectId), err.Error(), "application_object_id")
```
katbyte, line 134:
```suggestion
		return tf.ErrorDiagPathF(err, application_object_id, "retrieving Application with object ID %q", id.ObjectId)
```

internal/services/aadgraph/application_app_role_resource.go
```diff
@@ -184,34 +184,57 @@ func applicationAppRoleResourceRead(d *schema.ResourceData, meta interface{}) er
...
-	d.Set("application_object_id", id.ObjectId)
-	d.Set("role_id", id.RoleId)
-	d.Set("allowed_member_types", role.AllowedMemberTypes)
-	d.Set("description", role.Description)
-	d.Set("display_name", role.DisplayName)
-	d.Set("is_enabled", role.IsEnabled)
-	d.Set("value", role.Value)
+	if err := d.Set("application_object_id", id.ObjectId); err != nil {
+		return tf.ErrorDiag("Could not set attribute", err.Error(), "application_object_id")
+	}
```
katbyte, line 203: We now need to check for simple properties?? if so maybe we could write a helper:
```suggestion
	if diag := tf.Set("application_object_id", id.ObjectId); err != nil {
		return diag
	}
```

katbyte (changes requested):
THanks for this @manicminer - overall looks good but i think we can improve the diag wrappers. I've left some comments inline, WDYT?

jackofallops (approved):
LGTM 👍

internal/provider/provider.go
manicminer: Yes indeed!

internal/services/aadgraph/application_app_role_resource.go
manicminer: I'm not certain whether we _have_ to, but it makes the linter happy and I figured it was good practice? Thanks for the tip, have implemented that and it tidies it up nicely :+1:

manicminer:
@katbyte Many thanks for the review and suggestions. I've redone the ErrorDiag* functions and it reads a lot better. I considered having versions without Sprintf() but it works just the same even with no substitution vars. Thoughts? :)

internal/services/aadgraph/application_oauth2_permission_resource.go
```diff
@@ -201,26 +200,51 @@ func applicationOAuth2PermissionResourceRead(d *schema.ResourceData, meta interf
...
-	d.Set("admin_consent_description", permission.AdminConsentDescription)
-	d.Set("admin_consent_display_name", permission.AdminConsentDisplayName)
-	d.Set("is_enabled", permission.IsEnabled)
-	d.Set("type", permission.Type)
-	d.Set("user_consent_description", permission.UserConsentDescription)
-	d.Set("user_consent_display_name", permission.UserConsentDisplayName)
-	d.Set("value", permission.Value)
+	if dg := tf.Set(d, "application_object_id", id.ObjectId); dg != nil {
+		return dg
+	}
```
katbyte, line 205: We can ignore the err here in the linting setting

katbyte (approved):
Thanks @manicminer - looks good aside from one comment i've left inline

Issue #371, opened 2020-12-11
Documentation: azuread_application_password: Description has a 32 char limit


xenon8:
Please can you add a note to the Documentation (this [page](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/application_password#description)) that the Description field has a 32 char limit.

Thanks.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
Terraform 0.13.3 and Azure AD Provider 1.1.1

### Affected Resource(s)

* azuread_application_password

### Expected Behavior

Documentation should detail char limit.

### Actual Behavior

No limit is mentioned.

manicminer:
Hi @xenon8, many thanks for reporting. The constraints for this property are going to change in the next release and we'll be making a lot of accompanying documentation updates. It's unlikely we'll push any docs updates in the meantime, as our current process for doing so relies on publishing a new release.

In the meantime, we can keep this issue open as a reminder and it will be closed at the time of the next release.

manicminer:
Oh, and this length limit is going to change soon as the current constraint is due to an API limitation which won't apply in future.

PR #372, opened 2020-12-11
adding change to docs to detail char limit

xenon8:
related to issue #371


manicminer:
Thanks for this suggestion @xenon8. I also commented on the linked issue explaining in more detail, but at this time I don't think we'll merge this as there pending documentation updates which include this property and which will conflict. Accordingly I'm going to close this one, but thanks again and please feel free to open further issues or PRs (although it's probably worth asking in an issue before starting any major work).

PR #373, opened 2020-12-14
Initial support for Microsoft Graph with opt-in beta

manicminer:
This PR introduces support for Microsoft Graph using https://github.com/manicminer/hamilton, alongside the existing Azure Active Directory Graph API implementation, and aims to maintain forward and backward compatibility. Users can opt-in to using MS Graph by setting the `use_microsoft_graph` provider attribute or the `AAD_USE_MICROSOFT_GRAPH` environment variable.

This is broken into multiple commits to aid review.

### Authentication

- Authentication for MS Graph is handled by hamilton/auth similarly to go-azure-helpers. Different auth methods can be configured and are tried in order, until one succeeds.
  - Service principal auth is always tried first, if configured
  - If managed identity is enabled, it will raise an error on failure (current behavior)
  - Azure CLI is the last attempted method and serves as a fallback

- Authentication via hamilton/auth is only performed when MS Graph is enabled, to avoid erroring if no API permissions are granted to the principal.
- Authentication via go-azure-helpers is always performed, even when MS Graph is enabled. Both methods must work for provider configuration to succeed. This will be removed in v2.0.
- Client configuration is sourced from the AuthConfig supplied by go-azure-helpers. This will be switched in v2.0.
- All national clouds are supported
- The US Gov cloud is now [two clouds in MS Graph](https://docs.microsoft.com/en-us/graph/deployments) - L4 and L5.

### Resources

- All resource CRUD functions have been replaced with temporary shim functions to conditionally call the corresponding aadgraph or msgraph function.
- A number of `resource.StateChangeConf{}.WaitForState()` method usages have been removed since MS Graph has been observed to be consistent in most cases.
- Lots of deprecations and new fields to accommodate new API schema and nomenclature

### Documentation

- Several documentation fixes, deprecations added
- Upgrade guide for v2.0 and Microsoft Graph, with section explaining the MS Graph Beta


katbyte (changes requested):
Thanks @manicminer - did an initial review and left some comments inline

internal/acceptance/helpers/check_destroyed.go
```diff
@@ -10,25 +10,25 @@ import (
...
 			// Destroy is unconcerned with an error checking the status, since this is going to be "not found"
 			result, err := testResource.Exists(ctx, client, resourceState.Primary)
-			if result == nil && err != nil {
-				return fmt.Errorf("should have either an error or a result when checking if \"%s.%s\" has been destroyed", resourceType, resourceLabel)
+			if result == nil && err == nil {
+				return fmt.Errorf("should have either an error or a result when checking if %q has been destroyed", resourceName)
 			}
 			if result != nil && *result {
-				return fmt.Errorf("\"%s.%s\" still exists", resourceType, resourceLabel)
+				return fmt.Errorf("%q still exists", resourceName)
```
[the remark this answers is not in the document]
manicminer: TestData.ResourceTest() has been passing in the full resource name (`foo.bar`) for the 4th argument rather than just the label (`bar`), which is why I've gone for renaming the argument - previously this would have outputted something like "foo.foo.bar still exists".

internal/provider/provider.go
```diff
@@ -89,14 +91,15 @@ func AzureADProvider() *schema.Provider {
...
 				Description: "The Hostname which should be used for the Azure Metadata Service.",
 			},
 
 			"environment": {
 				Type:        schema.TypeString,
 				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("ARM_ENVIRONMENT", "public"),
-				Description: "The Cloud Environment which should be used. Possible values are `public`, `usgovernment`, `german`, and `china`. Defaults to `public`.",
+				DefaultFunc: schema.EnvDefaultFunc("ARM_ENVIRONMENT", "global"),
+				Description: "The national cloud environment which should be used. Possible values are `global`, `usgovernment`, `dod`, `germany`, and `china`. Defaults to `global`.",
```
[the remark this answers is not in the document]
manicminer: That's fair. MS literature calls everything a 'national' cloud but I agree it's misleading. I think it's ok for the dod cloud to be documented sice this is the newly split USGovernmentL5 cloud and is well advertised in Azure docs?

internal/services/applications/application_app_role_resource_msgraph.go
```diff
@@ -0,0 +1,198 @@
...
+		roleId = rid
+	}
+
+	allowedMemberTypesRaw := d.Get("allowed_member_types").(*schema.Set).List()
+	allowedMemberTypes := make([]string, 0, len(allowedMemberTypesRaw))
+	for _, a := range allowedMemberTypesRaw {
+		allowedMemberTypes = append(allowedMemberTypes, a.(string))
+	}
+
+	role := models.AppRole{
```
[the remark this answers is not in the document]
manicminer: Thanks this is a great tip and I refactored to use the `msgraph` package for both clients and models which is much nicer to import and use.

At the moment this conflicts with the `helpers/msgraph` package but this will itself be refactored in v2.0 once the aadgraph stuff is out.

internal/provider/provider.go
```diff
@@ -218,3 +256,24 @@ func buildClient(ctx context.Context, p *schema.Provider, b *authentication.Buil
 
 	return client, nil
 }
+
+func environment(name string) (env environments.Environment, aadEnv string) {
```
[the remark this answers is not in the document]
manicminer: I looked into this a little more, and whilst the Management API metadata lists the available token endpoints, MS Graph is not represented, nor are endpoints documented in it's own metadata (https://graph.microsoft.com/beta/$metadata), so as best I can tell there isn't any way to query these.

internal/services/applications/application_app_role_resource_aadgraph.go
```diff
@@ -0,0 +1,192 @@
...
+		rid, err := uuid.GenerateUUID()
+		if err != nil {
+			return tf.ErrorDiagF(err, "Generating App Role for application with object ID %q", objectId)
+		}
+		roleId = rid
+	}
+
+	allowedMemberTypesRaw := d.Get("allowed_member_types").(*schema.Set).List()
+	allowedMemberTypes := make([]string, 0, len(allowedMemberTypesRaw))
+	for _, a := range allowedMemberTypesRaw {
```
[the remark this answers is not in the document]
manicminer: The *_aadgraph.go files are literal copies of the current resources / data sources and in the pursuit of minimising change I think it's probably best to leave those as they are, warts and all? 😬

internal/services/users/user_resource_msgraph.go
```diff
@@ -0,0 +1,285 @@
...
+
+func userResourceCreateMsGraph(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client).Users.MsClient
+
+	upn := d.Get("user_principal_name").(string)
+	mailNickName := d.Get("mail_nickname").(string)
+
+	// default mail nickname to the first part of the UPN (matches the portal)
+	if mailNickName == "" {
+		mailNickName = strings.Split(upn, "@")[0]
```
tombuildsstuff, line 28: whilst there should be one @ here, is there a crash here?

internal/services/users/user_resource_msgraph.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if v, ok := d.GetOk("mobile_phone"); ok {
+		properties.MobilePhone = utils.String(v.(string))
+	} else if v, ok := d.GetOk("mobile"); ok {
+		properties.MobilePhone = utils.String(v.(string))
+	}
+
+	user, _, err := client.Create(ctx, properties)
+	if err != nil {
+		return tf.ErrorDiagF(err, "Creating properties %q", upn)
```
tombuildsstuff, line 106:
```suggestion
		return tf.ErrorDiagF(err, "Creating user %q", upn)
```

internal/services/users/user_resource_msgraph.go
```diff
@@ -0,0 +1,285 @@
...
+
+func userResourceDeleteMsGraph(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client).Users.MsClient
+
+	_, status, err := client.Get(ctx, d.Id())
+	if err != nil {
+		if status == http.StatusNotFound {
+			log.Printf("[DEBUG] User with Object ID %q already deleted", d.Id())
+			return nil
+		}
```
tombuildsstuff, line 274: (as above) this indicate a bug so we should remove this

internal/services/serviceprincipals/service_principal_resource_msgraph.go
```diff
@@ -0,0 +1,119 @@
...
+
+func servicePrincipalResourceDeleteMsGraph(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client).ServicePrincipals.MsClient
+
+	_, status, err := client.Get(ctx, d.Id())
+	if err != nil {
+		if status == http.StatusNotFound {
+			log.Printf("[DEBUG] Service Principal with Object ID %q already deleted", d.Id())
+			return nil
+		}
```
tombuildsstuff, line 108: (as above) can we remove this? this hides a bug

internal/services/serviceprincipals/service_principal_password_resource_msgraph.go
```diff
@@ -0,0 +1,160 @@
...
+	credential, err := helpers.PasswordCredentialForResource(d)
+	if err != nil {
+		attr := ""
+		if kerr, ok := err.(helpers.CredentialError); ok {
+			attr = kerr.Attr()
+		}
+		return tf.ErrorDiagPathF(err, attr, "Generating password credentials for service principal with object ID %q", objectId)
+	}
+
+	id := parse.NewCredentialID(objectId, "password", *credential.KeyId)
```
tombuildsstuff, line 32: there's a panic here?

internal/services/serviceprincipals/service_principal_password_resource_msgraph.go
```diff
@@ -0,0 +1,160 @@
...
+
+	id, err := parse.PasswordID(d.Id())
+	if err != nil {
+		return tf.ErrorDiagPathF(err, "id", "Parsing password credential with ID %q", d.Id())
+	}
+
+	app, status, err := client.Get(ctx, id.ObjectId)
+	if err != nil {
+		if status == http.StatusNotFound {
+			return tf.ErrorDiagPathF(nil, "service_principal_id", "Service principal with object ID %q was not found", id.ObjectId)
```
tombuildsstuff, line 81: in which case the password won't exist, so shouldn't this be removing it from the state?

internal/services/serviceprincipals/service_principal_password_resource_msgraph.go
```diff
@@ -0,0 +1,160 @@
...
+
+	tf.LockByName(servicePrincipalResourceName, id.ObjectId)
+	defer tf.UnlockByName(servicePrincipalResourceName, id.ObjectId)
+
+	app, status, err := client.Get(ctx, id.ObjectId)
+	if err != nil {
+		if status == http.StatusNotFound {
+			log.Printf("[DEBUG] Service Principal with Object ID %q was not found - removing from state!", id.ObjectId)
+			return nil
+		}
```
tombuildsstuff, line 138: (as above) this is hiding a bug which should be in the refresh/read - can we remove this

internal/services/serviceprincipals/service_principal_certificate_resource_msgraph.go
```diff
@@ -0,0 +1,164 @@
...
+	credential, err := helpers.KeyCredentialForResource(d)
+	if err != nil {
+		attr := ""
+		if kerr, ok := err.(helpers.CredentialError); ok {
+			attr = kerr.Attr()
+		}
+		return tf.ErrorDiagPathF(err, attr, "Generating certificate credentials for service principal with object ID %q", objectId)
+	}
+
+	id := parse.NewCredentialID(objectId, "certificate", *credential.KeyId)
```
tombuildsstuff, line 32: there's a potential crash here?

internal/services/serviceprincipals/service_principal_certificate_resource_msgraph.go
```diff
@@ -0,0 +1,164 @@
...
+
+	id, err := parse.CertificateID(d.Id())
+	if err != nil {
+		return tf.ErrorDiagPathF(err, "id", "Parsing certificate credential with ID %q", d.Id())
+	}
+
+	app, status, err := client.Get(ctx, id.ObjectId)
+	if err != nil {
+		if status == http.StatusNotFound {
+			return tf.ErrorDiagPathF(nil, "service_principal_id", "Service principal with object ID %q was not found", id.ObjectId)
```
tombuildsstuff, line 81: (as.. below?) if the service principals gone, the certificate won't exist - so shouldn't this be removing it from the state?

internal/services/serviceprincipals/service_principal_certificate_resource_msgraph.go
```diff
@@ -0,0 +1,164 @@
...
+
+	tf.LockByName(servicePrincipalResourceName, id.ObjectId)
+	defer tf.UnlockByName(servicePrincipalResourceName, id.ObjectId)
+
+	app, status, err := client.Get(ctx, id.ObjectId)
+	if err != nil {
+		if status == http.StatusNotFound {
+			log.Printf("[DEBUG] Service Principal with Object ID %q was not found - removing from state!", id.ObjectId)
+			return nil
+		}
```
tombuildsstuff, line 142: (as elsewhere) this hides a bug within the Read/refresh where we're not removing this from the state - so can we remove this?

tombuildsstuff (changes requested):
hey @manicminer 

I've taken a look through and left some comments inline - but on the whole this is looking pretty good 👍 

My main questions here are around the upgrade to 2.0 (and adding some means for users to switch to the newer fields prior to 2.0) before they're removed, so this isn't a breaking config change - so I think it'd be worth introducing those renamed fields in a 1.x release before we do the 2.0 release to remove the old fields. Whilst that does add a little complexity, it would enable a migration path for users rather than having a bunch of config errors they can't fix at 2.0, which seems worth the trade-off - that said - this is likely outside of the scope of this initial PR (e.g. a follow up 1.x release) - due to this being spread across other resources.

I've also taken a look through Hamilton and left a few comments, but that's also mostly looking good - most of the concerns there are around error handling/rate-limiting, at this time [we're using the default HTTP Client](https://github.com/manicminer/hamilton/blob/aa88cc46fa64ba21d278e16389e0f0dc86bc9a16/msgraph/client.go#L72) but that won't handle 409's or retrying on 500's - so we may want to look into using something like https://github.com/hashicorp/go-retryablehttp (which [would allow us to handle that for 429's and 500's](https://github.com/hashicorp/go-retryablehttp/blob/991b9d0a42d13014e3689dd49a94c02be01f4237/client_test.go#L538-L542)) to handle that, presumably [wired up here](https://github.com/hashicorp/terraform-provider-azuread/blob/222b824e811514a535cb4d09d96a3979e1f44481/internal/common/client_options.go#L32-L40) - WDYT?

That said, on the whole this is looking pretty good - if we can fix up the comments then this should otherwise be good, nice work 👍 

Thanks!

vendor/github.com/manicminer/hamilton/environments/published.go
```diff
@@ -0,0 +1,106 @@
+package environments
+
+type ApiAppId string
+
+// PublishedApis is a map containing Application IDs for well known APIs published by Microsoft.
+// They can be used to acquire access tokens, but are primarily described here for easy inclusion in
+// application manifests and service principal assignments.
+var PublishedApis = map[string]ApiAppId{
```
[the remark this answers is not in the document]
manicminer: I've been told these are the same in all clouds

vendor/github.com/manicminer/hamilton/auth/clientcredentials.go
```diff
@@ -0,0 +1,319 @@
...
+
+	// Scopes specifies a list of requested permission scopes (used for v2 tokens)
+	Scopes []string
+
+	// TokenURL is the clientCredentialsToken endpoint. Typically you can use the AzureADEndpoint
+	// function to obtain this value, but it may change for non-public clouds.
+	TokenURL string
+
+	// Expires optionally specifies how long the clientCredentialsToken is valid for.
+	Expires time.Duration
```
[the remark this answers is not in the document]
manicminer: This actually ended up not being used, so I've removed the field

internal/provider/provider.go
```diff
@@ -89,14 +91,15 @@ func AzureADProvider() *schema.Provider {
...
 				Description: "The Hostname which should be used for the Azure Metadata Service.",
 			},
 
 			"environment": {
 				Type:        schema.TypeString,
 				Required:    true,
-				DefaultFunc: schema.EnvDefaultFunc("ARM_ENVIRONMENT", "public"),
-				Description: "The Cloud Environment which should be used. Possible values are `public`, `usgovernment`, `german`, and `china`. Defaults to `public`.",
+				DefaultFunc: schema.EnvDefaultFunc("ARM_ENVIRONMENT", "global"),
+				Description: "The cloud environment which should be used. Possible values are `global`, `usgovernment`, `dod`, `germany`, and `china`. Defaults to `global`.",
```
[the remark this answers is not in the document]
manicminer: We do still accept `public`, I've added it to the description string for clarity

internal/helpers/msgraph/replication.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+	}).WaitForStateContext(ctx)
+}
+
+func WaitForListRemove(ctx context.Context, item string, f func() ([]string, error)) (interface{}, error) {
+	return (&resource.StateChangeConf{
+		Pending:                   []string{"Found"},
+		Target:                    []string{"NotFound"},
+		Timeout:                   5 * time.Minute,
+		MinTimeout:                1 * time.Second,
```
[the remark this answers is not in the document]
manicminer: AFAICT the SDK implements exponential backoff here, so the refresh interval should rapidly rise if a replication delay occurs? So far in testing, the API has been consistent very quickly (unlike the old API) so this is here more as a safety measure.

internal/helpers/msgraph/application.go
```diff
@@ -0,0 +1,309 @@
...
+		if _, err := client.Update(ctx, properties); err != nil {
+			return fmt.Errorf("disabling App Roles for Application with object ID %q: %+v", *application.ID, err)
+		}
+	}
+
+	// then set the new roles
+	properties := msgraph.Application{
+		ID:       application.ID,
+		AppRoles: newRoles,
+	}
```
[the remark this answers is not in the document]
manicminer: The value set in the config should be reflected here (users can disable roles/scopes)

internal/services/applications/application_app_role_resource_msgraph.go
```diff
@@ -0,0 +1,194 @@
...
+
+	app, status, err := client.Get(ctx, id.ObjectId)
+	if err != nil {
+		if status == http.StatusNotFound {
+			return tf.ErrorDiagPathF(nil, "application_object_id", "Application with object ID %q was not found", id.ObjectId)
+		}
+		return tf.ErrorDiagPathF(err, "application_object_id", "Retrieving Application with object ID %q", id.ObjectId)
+	}
+
+	if d.IsNewResource() {
```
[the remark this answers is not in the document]
manicminer: In 2.0 I'm planning to make `role_id` mandatory as this is the only way to identify a role (same applies to oauth2_permissions), I'll split up the CreateUpdate func then as it will be a lot easier to support delta updates?

internal/services/applications/application_data_source.go
```diff
@@ -97,8 +95,9 @@ func applicationDataSource() *schema.Resource {
 
 			// TODO: v2.0 move this to `sign_in_audience` property and support other values
 			"available_to_other_tenants": {
-				Type:     schema.TypeBool,
-				Computed: true,
+				Type:       schema.TypeBool,
+				Computed:   true,
+				Deprecated: "[NOTE] This attribute will be replaced by a new property `sign_in_audience` in version 2.0 of the AzureAD provider",
```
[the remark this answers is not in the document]
manicminer: Sure thing, for now it's with only two values that can be translated into the old boolean field

internal/services/applications/application_data_source_msgraph.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		switch {
+		case result == nil || len(*result) == 0:
+			return tf.ErrorDiagF(fmt.Errorf("No applications found matching filter: %q", filter), "Application not found")
+		case len(*result) > 1:
+			return tf.ErrorDiagF(fmt.Errorf("Found multiple applications matching filter: %q", filter), "Multiple applications found")
+		}
+
+		app = &(*result)[0]
```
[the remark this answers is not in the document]
manicminer: Since `result` is a *[]Application rather than []*Application, the compiler throws an invalid operation error

internal/services/applications/application_data_source_msgraph.go
```diff
@@ -0,0 +1,127 @@
...
+	if v := app.IsFallbackPublicClient; v != nil && *v {
+		appType = "native"
+	} else {
+		appType = "webapp/api"
+	}
+	tf.Set(d, "type", appType)
+
+	if app.Api != nil {
+		tf.Set(d, "oauth2_permissions", helpers.ApplicationFlattenOAuth2Permissions(app.Api.OAuth2PermissionScopes))
+	}
```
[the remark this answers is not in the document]
manicminer: I've opted to work around this in the Read func for now, since the schema in the next version will mirror this with nested blocks and the flatten funcs will change then.

internal/services/applications/application_resource_test.go
```diff
@@ -51,7 +52,7 @@ func TestAccApplication_basicDeprecated(t *testing.T) {
 				check.That(data.ResourceName).Key("display_name").HasValue(fmt.Sprintf("acctest-APP-%d", data.RandomInteger)),
 			),
 		},
-		data.ImportStep(),
+		//data.ImportStep(),
```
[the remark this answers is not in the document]
manicminer: It is indeed 👍

internal/services/domains/client/client.go
```diff
@@ -2,19 +2,23 @@ package client
...
 
 type Client struct {
 	AadClient *graphrbac.DomainsClient
+	MsClient  *msgraph.DomainsClient
 }
 
 func NewClient(o *common.ClientOptions) *Client {
 	aadClient := graphrbac.NewDomainsClientWithBaseURI(o.AadGraphEndpoint, o.TenantID)
-	o.ConfigureClient(&aadClient.Client, o.AadGraphAuthorizer)
+	msClient := msgraph.NewDomainsClient(o.TenantID)
```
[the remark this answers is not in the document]
manicminer: I ended up setting this [in the ClientOptions.ConfigureClient() method](https://github.com/hashicorp/terraform-provider-azuread/blob/5d84cf4d72d6791ae0a8ca2dfa2c7917cb083896/internal/common/client_options.go#L34) rather than expand all the constructors to support args that would invariably have the same values for most users

internal/services/groups/group_resource_msgraph.go
```diff
@@ -0,0 +1,244 @@
...
+		return tf.ErrorDiagF(err, "Failed to generate mailNickname")
+	}
+
+	properties := msgraph.Group{
+		DisplayName:  utils.String(displayName),
+		MailNickname: utils.String(mailNickname),
+
+		// API only supports creation of security groups
+		SecurityEnabled: utils.Bool(true),
+		MailEnabled:     utils.Bool(false),
```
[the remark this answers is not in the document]
manicminer: Sadly yes, msgraph returns a 400 with "The service does not currently support writes of mail-enabled groups" if you try to set MailEnabled on create or update :(

vendor/github.com/manicminer/hamilton/msgraph/client.go
```diff
@@ -0,0 +1,412 @@
...
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, status, nil, err
+	}
+
+	var o odata.OData
+
+	// Check for json content before looking for odata metadata
+	contentType := strings.ToLower(resp.Header.Get("Content-Type"))
+	if strings.HasPrefix(contentType, "application/json") {
```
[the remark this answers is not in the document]
manicminer: Happy to make it case insensitive but I think we should only match from the start of the value?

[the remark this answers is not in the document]
manicminer: Oh, it already is effectively case insensitive since `contentType` is forcibly lowercased :)

vendor/github.com/manicminer/hamilton/msgraph/client.go
```diff
@@ -0,0 +1,412 @@
...
+
+	// Perform the request
+	resp, status, o, err := c.performRequest(req, input)
+	if err != nil {
+		return nil, status, o, err
+	}
+
+	// Check for json content before handling pagination
+	contentType := strings.ToLower(resp.Header.Get("Content-Type"))
+	if strings.HasPrefix(contentType, "application/json") {
```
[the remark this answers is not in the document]
manicminer: Yeah, perhaps if a third occurrence arises? At the moment it's only in two places

internal/services/serviceprincipals/client_config_data_source_msgraph.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/hashicorp/terraform-provider-azuread/internal/tf"
+)
+
+func clientConfigDataSourceReadMsGraph(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client)
+
+	objectId := client.Claims.ObjectId
+	if client.Claims.ObjectId == "" {
+		return tf.ErrorDiagPathF(errors.New("oid claim in access token is empty"), "object_id", "Could not determine object ID of authenticated principal")
+	}
```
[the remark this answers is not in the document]
manicminer: Good point, whilst technically we can proceed without this (aside from this data source, currently), it's not a great indicator if we can't determine this upfront

internal/services/users/user_resource_msgraph.go
manicminer: AFAICT, if the "@" is absent, or even for empty strings, indexing the zeroth item always seems to succeed

internal/services/serviceprincipals/service_principal_password_resource_msgraph.go
manicminer: I thought about this, if for some reason the service_principal_id either gets mismatched or there's a 404 for some other reason, the primary copy of the password is in the state, and not auto-removing it could save users from a lot of rotation?

internal/services/serviceprincipals/service_principal_certificate_resource_msgraph.go
manicminer: Similarly to passwords, I feel a bit hesitant about removing certificates from state, wouldn't it be safer to have the user run `tf state rm`, or fix their config?

vendor/github.com/manicminer/hamilton/msgraph/applications.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	if application.Owners == nil {
+		return status, errors.New("cannot update application with nil Owners")
+	}
+	for _, owner := range *application.Owners {
+		// don't fail if an owner already exists
+		checkOwnerAlreadyExists := func(resp *http.Response, o *odata.OData) bool {
+			if resp.StatusCode == http.StatusBadRequest {
+				if o.Error != nil {
+					re := regexp.MustCompile("One or more added object references already exist")
```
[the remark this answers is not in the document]
manicminer: Great idea, have done that for the odata errors already being used

internal/helpers/msgraph/application.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// first disable any existing roles
+	if app.AppRoles != nil && len(*app.AppRoles) > 0 {
+		properties := msgraph.Application{
+			ID:       application.ID,
+			AppRoles: app.AppRoles,
+		}
+
+		for _, role := range *properties.AppRoles {
+			*role.IsEnabled = false
```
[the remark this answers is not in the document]
manicminer: I'm not certain why, but changing this causes the pointer to not be updated. Will have to dig a little more.

internal/helpers/msgraph/credentials.go
```diff
@@ -0,0 +1,160 @@
...
+		EndDateTime: &endDate,
+		KeyId:       utils.String(keyId),
+		SecretText:  utils.String(value),
+	}
+
+	if v, ok := d.GetOk("display_name"); ok {
+		credential.DisplayName = utils.String(v.(string))
+	} else if v, ok := d.GetOk("description"); ok { // TODO: remove in v2.0
+		credential.DisplayName = utils.String(v.(string))
+	}
```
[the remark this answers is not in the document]
manicminer: Perhaps, although it's also needed on the read operation to help reduce diff noise, particularly if people switch back and forth between the two APIs.

vendor/github.com/manicminer/hamilton/msgraph/client.go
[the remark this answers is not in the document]
manicminer: So... looking at this closer with a view to tidying it a bit, I discovered that my previously committed retry logic and some other optimisations got lost during a refactor, presumably I forgot to git pull at some point. That's now fixed and as a consequence did in fact tidy this up a bit :)

manicminer:
@tombuildsstuff Thanks again for the mega review! I've worked through most of it and introduced the new properties for Applications. At the moment it's using a lot of if-else blocks to handle the deprecated properties but this will be migrated to expand funcs in 2.0 (the flatten funcs are already in since they're suitable now). I still need to add the documentation and fix up a couple of tests around the `oauth2_permissions` / `api.0.oauth2_permissikon_scopes` properties, but I think it's working well so far if you're able to take a second look?

I also haven't merged the updates to Hamilton yet, will ensure that's done and tagged prior to merge.

On closer inspection of the graph client in Hamilton, I noticed that I somehow stripped out the retry logic for 429s and 5xx - that's now re-added (but not merged) and shows up in this PR. I looked at go-retryablehttp and I think it could work well, but as best I can tell it's not accepted by the oauth2 package so we'd need to shim that somehow?

internal/helpers/msgraph/application.go
```diff
@@ -0,0 +1,386 @@
...
+	owners, _, err := client.ListOwners(ctx, *application.ID)
+	if err != nil {
+		return fmt.Errorf("retrieving owners for Application with object ID %q: %+v", *application.ID, err)
+	}
+
+	existingOwners := *owners
+	ownersForRemoval := utils.Difference(existingOwners, desiredOwners)
+	ownersToAdd := utils.Difference(desiredOwners, existingOwners)
+
+	if ownersForRemoval != nil {
```
Yanson, line 328: Since you can't delete the last owner, this will fail if you are changing a single owner. Recommend adding owners first and moving this block underneath the `ownersToAdd` block so there is always at least one owner.

vendor/github.com/manicminer/hamilton/auth/msi.go
```diff
@@ -0,0 +1,148 @@
...
+	"net/http"
+	"net/url"
+	"strconv"
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+const (
+	msiDefaultApiVersion = "2018-02-01"
```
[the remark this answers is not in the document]
manicminer: Short answer, yes, either with the default endpoint or with the CloudShell-specific one `http://localhost:50342/oauth2/token`. It (cloudshell) is pretty buggy however and frequently one of the token requests in a typical TF apply will time out.

internal/helpers/msgraph/application.go
manicminer: @Yanson Great catch, thanks. I've updated so that owners for groups and applications are both added before old ones are removed.

Yanson: Thanks for updating. Is there any way to add a test to prevent this regressing?

> ... and have no additional permissions

Unfortunately in our organisation we are not permitted such elevated permission and may only mange the groups we created (and own).

Also, I don't think it's related to this PR, but the msgraph API - I was unable to remove the second-to-last owner if the last owner is a Service Principal. If you know anything about that, it would be great to address.

manicminer: > Thanks for updating. Is there any way to add a test to prevent this regressing?

Not easily since we typically use Global Admin to separate concerns between permissions of our test account and the accounts it manages during testing, but I'll look into it.

> Also, I don't think it's related to this PR, but the msgraph API - I was unable to remove the second-to-last owner if the last owner is a Service Principal. If you know anything about that, it would be great to address.

Interesting, I haven't been bitten by this but will do some testing 👍

manicminer:
Rebased after merging #426

internal/services/serviceprincipals/service_principal_certificate_resource_msgraph.go
tombuildsstuff: nah that seems reasonable, this would show up as 'gone' in the plan - so that's the expected behaviour here

internal/services/serviceprincipals/client_config_data_source_msgraph.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/hashicorp/terraform-provider-azuread/internal/clients"
+	"github.com/hashicorp/terraform-provider-azuread/internal/tf"
+)
+
+func clientConfigDataSourceReadMsGraph(_ context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client)
+	objectId := client.Claims.ObjectId
+	d.SetId(fmt.Sprintf("%s-%s-%s", client.TenantID, client.ClientID, objectId))
```
tombuildsstuff, line 17: Might be worth making the ID here just the Tenant ID, since the Client ID and Object ID would likely change through other invocations. That's allow users who are using different accounts (e.g. CI and local) against the same tenant to show changes and not force-new resources if the values/ID changes?

internal/services/applications/application_resource.go
```diff
@@ -57,21 +53,101 @@ func applicationResource() *schema.Resource {
...
+				},
+			},
+
+			// TODO: v2.0 consider another computed typemap attribute `app_role_ids` for easier consumption
 			"app_role": {
 				Type:       schema.TypeSet,
 				Optional:   true,
 				Computed:   true,
 				ConfigMode: schema.SchemaConfigModeAttr,
+				//Set:        tf.SetFuncHashId, // TODO: v2.0 enable this SchemaSetFunc when id is no longer Computed
```
tombuildsstuff, line 145: we probably don't want to do this since it means changes to fields other than ID won't be tracked?

internal/services/applications/application_resource.go
```diff
@@ -57,21 +53,101 @@ func applicationResource() *schema.Resource {
...
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						// TODO: v2.0 also consider another computed typemap attribute `oauth2_permission_scope_ids` for easier consumption
+						"oauth2_permission_scope": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Computed: true,
+							Set:      tf.SetFuncHashId,
```
tombuildsstuff, line 73: (as below) the set function only using the ID will only track changes to the ID, which likely isn't what we want?

internal/services/applications/application_password_resource.go
```diff
@@ -30,7 +25,74 @@ func applicationPasswordResource() *schema.Resource {
...
+			},
+
+			"value": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				Sensitive:    true,
+				Deprecated:   "[NOTE] In version 2.0 of the AzureAD provider, this attribute will become read-only as it will no longer be possible to specify a password value. It will be generated by Azure Active Directory and persisted to state for reuse in your Terraform configuration.",
+				ValidateFunc: validation.StringLenBetween(1, 863), // Encrypted secret cannot be empty and can be at most 1024 bytes.
+			},
```
tombuildsstuff, line 69: we could conditionally change the behaviour of this field in the Beta to be either Required & ForceNew or Computed? e.g.

```suggestion
			"value": func() *schema.Schema {
				if msGraphEnabled {
					return &schema.Schema{
						Type:         schema.TypeString,
						Computed:     true,
						Sensitive:    true,
					}
				}
				
				return &schema.Schema{
					Type:         schema.TypeString,
					Required:     true,
					ForceNew:     true,
					Sensitive:    true,
					Deprecated:   "[NOTE] In version 2.0 of the AzureAD provider, this attribute will become read-only as it will no longer be possible to specify a password value. It will be generated by Azure Active Directory and persisted to state for reuse in your Terraform configuration.",
					ValidateFunc: validation.StringLenBetween(1, 863), // Encrypted secret cannot be empty and can be at most 1024 bytes.
				}
			}
			}(),
```

whilst not ideal it seems fine with a flag in the documentation during the beta?

internal/services/applications/application_resource_msgraph.go
```diff
@@ -0,0 +1,688 @@
...
+	if claims := flattenApplicationOptionalClaim(in.AccessToken); len(claims) > 0 {
+		optionalClaims["access_token"] = claims
+	}
+	if claims := flattenApplicationOptionalClaim(in.IdToken); len(claims) > 0 {
+		optionalClaims["id_token"] = claims
+	}
+	// TODO: v2.0 enable this
+	//if claims := flattenApplicationOptionalClaim(in.Saml2Token); len(claims) > 0 {
+	//	optionalClaims["saml2_token"] = claims
+	//}
```
tombuildsstuff, line 614: these fields need to be set in all circumstances

internal/services/applications/application_resource_msgraph.go
```diff
@@ -0,0 +1,688 @@
...
+			"additional_properties": []string{},
+		}
+
+		if claim.Source != nil {
+			optionalClaim["source"] = *claim.Source
+		}
+
+		if claim.AdditionalProperties != nil && len(*claim.AdditionalProperties) > 0 {
+			optionalClaim["additional_properties"] = *claim.AdditionalProperties
+		}
```
tombuildsstuff, line 644: same here

internal/services/applications/application_resource_msgraph.go
```diff
@@ -0,0 +1,688 @@
...
+	if in == nil {
+		return []map[string]interface{}{}
+	}
+
+	result := make([]map[string]interface{}, 0)
+	for _, requiredResourceAccess := range *in {
+		resource := make(map[string]interface{})
+		if requiredResourceAccess.ResourceAppId != nil {
+			resource["resource_app_id"] = *requiredResourceAccess.ResourceAppId
+		}
```
tombuildsstuff, line 662: and here

internal/services/applications/application_resource_msgraph.go
```diff
@@ -0,0 +1,688 @@
...
+func flattenApplicationResourceAccess(in *[]msgraph.ResourceAccess) []interface{} {
+	if in == nil {
+		return []interface{}{}
+	}
+
+	accesses := make([]interface{}, 0)
+	for _, resourceAccess := range *in {
+		access := make(map[string]interface{})
+		if resourceAccess.ID != nil {
+			access["id"] = *resourceAccess.ID
```
tombuildsstuff, line 681: and here

internal/services/applications/application_oauth2_permission_resource_aadgraph.go
```diff
@@ -0,0 +1,198 @@
...
+	if d.IsNewResource() {
+		newPermissions, err = aadgraph.OAuth2PermissionAdd(app.Oauth2Permissions, &permission)
+		if err != nil {
+			if _, ok := err.(*aadgraph.AlreadyExistsError); ok {
+				return tf.ImportAsExistsDiag("azuread_application_oauth2_permission", id.String())
+			}
+			return tf.ErrorDiagF(err, "Failed to add OAuth2 Permission")
+		}
+	} else {
+		if existing, _ := aadgraph.OAuth2PermissionFindById(app, id.PermissionId); existing == nil {
```
tombuildsstuff, line 80: we should check the error here too?

internal/services/applications/application_oauth2_permission_resource_msgraph.go
```diff
@@ -0,0 +1,206 @@
...
+			app.Api = &msgraph.ApplicationApi{}
+		}
+		if app.Api.AppendOAuth2PermissionScope(scope) != nil {
+			if _, ok := err.(*grapherrors.AlreadyExistsError); ok {
+				return tf.ImportAsExistsDiag("azuread_application_oauth2_permission", id.String())
+			}
+			return tf.ErrorDiagF(err, "Failed to add OAuth2 Permission")
+		}
+	} else {
+		if existing, _ := helpers.OAuth2PermissionFindById(app, id.PermissionId); existing == nil {
```
tombuildsstuff, line 82: same here?

tombuildsstuff (changes requested):
hey @manicminer 

Thanks for pushing those changes - I've taken another look through and if we can fix those up (so that the `value` field is exposed for the password resources, so those can use the Beta?) and the tests pass this should otherwise be good to go 👍 

Thanks!

internal/services/serviceprincipals/client_config_data_source_msgraph.go
manicminer: Granted you can end up with diff flapping if you run Terraform with different auth principals, but this is also commonly used to assign object ownership in other resources and if that ends up mismatched due to running with a different principal then Terraform can lock itself out of other resources.

internal/services/applications/application_resource.go
manicminer: Absolutely. I misunderstood the semantics of changes to set members, this doesn't do what I had expected 👍

vendor/github.com/manicminer/hamilton/environments/apis.go
```diff
@@ -0,0 +1,97 @@
...
+	MsGraphCliName  ApiCliName = "ms-graph"
+)
+
+// API represent an API configuration for Microsoft Graph or Azure Active Directory Graph.
+type Api struct {
+	// The Application ID for the API.
+	AppId ApiAppId
+
+	// The Azure CLI codename for the API. Used with `az account get-access-token`.
+	CliName ApiCliName
```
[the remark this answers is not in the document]
manicminer: Ah thanks good to bear in mind. This needs further work anyway since the "codename" is defined in the `environments` package but isn't yet consumed by the `auth` package, which currently just uses an iota to toggle either API for CLI authentication.

internal/services/applications/application_password_resource.go
manicminer: Agreed, this makes sense 👍

manicminer: I think this comes back to one of the initial poblems we had, in that the provider and resource schemas are declared before the provider is configured. Not sure if there's a way to defer this until after it's been determined whether the beta is enabled?

manicminer: @tombuildsstuff This has gotten really fun.

In addition to the `value` field, the `key_id` (`keyId`) field is also no longer allowed to be user-supplied. However, despite [documentation to the contrary](https://docs.microsoft.com/en-us/graph/api/application-addpassword?view=graph-rest-beta&tabs=http), the remaining fields `display_name`, `start_date` and `end_date` are completely ignored by the API. It assigns null, now and now+2yrs respectively for each of those.

So I've opted to reject those properties at apply time when MS Graph is used, with a view to making them all Computed in v2.0. I'll raise the issue of the other fields being ignored and then hopefully we can make them Optional in future. WDYT?

vendor/github.com/manicminer/hamilton/auth/cache.go
```diff
@@ -0,0 +1,36 @@
...
+	mutex  sync.Mutex
+	token  *oauth2.Token
+}
+
+// Token returns the current token if it's still valid, else will acquire a new token
+func (c *cachedAuthorizer) Token() (*oauth2.Token, error) {
+	if c.token == nil || !c.token.Valid() {
+		c.mutex.Lock()
+		defer c.mutex.Unlock()
+		token, err := c.source.Token()
```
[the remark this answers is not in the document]
manicminer: Circling back to this, I see what you mean. I've used RWMutex to good effect elsewhere, do you have any thoughts on this?

```go
func (c *cachedAuthorizer) Token() (*oauth2.Token, error) {
	c.mutex.RLock()
	valid := c.token != nil && c.token.Valid()
	c.mutex.RUnlock()

	if !valid {
		c.mutex.Lock()
		defer c.mutex.Unlock()
		token, err := c.source.Token()
		if err != nil {
			return nil, err
		}
		c.token = token
	}

	return c.token, nil
}
```

internal/tf/import.go
```diff
@@ -34,3 +34,13 @@ func ValidateResourceIDPriorToImportThen(idParser ResourceIDValidator, importer
 		},
 	}
 }
+
+func ImportNotSupported() *schema.ResourceImporter {
```
tombuildsstuff, line 38: (as above) this isn't necessary - Terraform supports this natively if we remove this

vendor/github.com/manicminer/hamilton/msgraph/groups.go
```diff
@@ -0,0 +1,451 @@
...
+	return &data.Id, status, nil
+}
+
+// AddOwners adds a new owner to a Group.
+// First populate the Owners field of the Group using the AppendOwner method of the model, then call this method.
+func (c *GroupsClient) AddOwners(ctx context.Context, group *Group) (int, error) {
+	var status int
+	for _, owner := range *group.Owners {
+		// don't fail if an owner already exists
+		checkOwnerAlreadyExists := func(resp *http.Response, o *odata.OData) bool {
```
tombuildsstuff, line 379: wouldn't this indicate a bug?

vendor/github.com/manicminer/hamilton/msgraph/groups.go
```diff
@@ -0,0 +1,451 @@
...
+		// check for ownership before attempting deletion
+		if _, status, err := c.GetOwner(ctx, id, ownerId); err != nil {
+			if status == http.StatusNotFound {
+				continue
+			}
+			return status, err
+		}
+
+		// despite the above check, sometimes owners are just gone
+		checkOwnerGone := func(resp *http.Response, o *odata.OData) bool {
```
tombuildsstuff, line 428: (as above) wouldn't this indicate a bug?

vendor/github.com/manicminer/hamilton/odata/odata.go
```diff
@@ -0,0 +1,124 @@
...
+		return err
+	}
+	*o = OData(o2)
+
+	// Look for errors in the "error" and "odata.error" fields
+	var e map[string]json.RawMessage
+	if err := json.Unmarshal(data, &e); err != nil {
+		return err
+	}
+	for _, k := range []string{"error", "odata.error"} {
```
tombuildsstuff, line 46: can this also be in `@odata.error`?

tombuildsstuff (approved):
a few minor comments but this otherwise LGTM 👍

vendor/github.com/manicminer/hamilton/msgraph/groups.go
manicminer: This mostly helps to work around API inconsistency where newly created groups (and other objects) sometimes inherit an owner but sometimes don't

vendor/github.com/manicminer/hamilton/msgraph/groups.go
manicminer: Similarly sometimes the read operation on a group indicates the presence of an owner when it's actually already been removed, so you end up with a GET showing an owner, but DELETEing that owner gets you a 403.

vendor/github.com/manicminer/hamilton/odata/odata.go
manicminer: I haven't seen that but maybe? AFAICT `@odata.error` isn't in the odata standard, although I think i'd probably expect it to have astring value if it showed up anywhere.

docs/resources/application.md
```diff
@@ -135,18 +145,53 @@ The following arguments are supported:
...
+* `display_name` - (Required) Display name for the app role that appears during app role assignment and in consent experiences.
+* `enabled` - (Optional) Determines if the app role is enabled: Defaults to `true`.
 * `id` - The unique identifier of the app role. This attribute is computed and cannot be specified manually in this block. If you need to specify a custom `id`, it's recommended to use the [azuread_application_app_role](application_app_role.html) resource.
-* `is_enabled` - (Optional) Determines if the app role is enabled: Defaults to `true`.
-* `value` - (Optional) Specifies the value of the roles claim that the application should expect in the authentication and access tokens.
+* `value` - (Optional) The value that is used for the `roles` claim in ID tokens and OAuth 2.0 access tokens that are authenticating an assigned service or user principal.
+
+~> In version 2.0 of the provider, the `id` property will become mandatory. For more information, see the [Upgrade Guide for v2.0](../guides/microsoft-graph.html).
```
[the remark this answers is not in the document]
manicminer: no reason why not :D

docs/resources/application_oauth2_permission_scope.md
```diff
@@ -0,0 +1,58 @@
...
+* `type` - (Required) Whether this delegated permission should be considered safe for non-admin users to consent to on behalf of themselves, or whether an administrator should be required for consent to the permissions. Defaults to `User`. Possible values are `User` or `Admin`.
+* `user_consent_description` - (Optional) Delegated permission description that appears in the end user consent experience, intended to be read by a user consenting on their own behalf.
+* `user_consent_display_name` - (Optional) Display name for the delegated permission that appears in the end user consent experience.
+* `value` - (Required) The value that is used for the `scp` claim in OAuth 2.0 access tokens.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+*No additional attributes are exported*
```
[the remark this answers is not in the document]
manicminer: I have tried to avoid mentioning the resource ID too much as it's a frequent source of confusion - instead there are contextual attributes like `scope_id` in this case.

manicminer:
## Test Results

### Applications - AAD Graph

Screenshot 2021-05-19 at 20 50 55

### Applications - MS Graph

Screenshot 2021-05-19 at 21 00 44

### Domains - AAD Graph

Screenshot 2021-05-19 at 21 01 45

### Domains - MS Graph

Screenshot 2021-05-19 at 21 02 23

### Groups - AAD Graph

Screenshot 2021-05-19 at 21 02 55

### Groups - MS Graph

Screenshot 2021-05-19 at 21 03 24

### Service Principals - AAD Graph

Screenshot 2021-05-19 at 21 04 10

### Service Principals - MS Graph

Screenshot 2021-05-19 at 21 04 42

### Users - AAD Graph

Screenshot 2021-05-19 at 21 05 17

### Users - MS Graph

Screenshot 2021-05-19 at 21 05 46

ken5scal:
Great work!

simongottschlag:
😁🎉🚀💐🥇⭐️🙌🥳

JonZeolla:
Awesome work!

bergmeister:
It seems as part of this (commit [here](https://github.com/hashicorp/terraform-provider-azuread/commit/4bfce613d6260852e8f4d99bc54f8e49dc285a79#diff-20e1b0b3a19616f5586877bf83247b078abe90d7249d02222ba79751db540857)), the properties of many azuread resources like e.g. `azuread_group` changed from `display_name` to `name`. Was this breaking change intentional as the PR title seems to indicate the new MSGraph functionality is off by default for now?

Issue #280, opened 2020-06-16
Add support for "mail" lookup


trajano:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The following also works so if we had a mail option for the data "azuread_user" this would solve the issue. It should throw an error if it detects that the query will return more than one result.

az ad user list --filter "mail eq 'email@guest.com'"

### New or Affected Resource(s)

* data azuread_user

### Potential Terraform Configuration

```hcl
data "azuread_user" "YYYY" {
  mail = "YYYY@ZZZZ.com"
}
```

### References

* #275

manicminer:
Hi @trajano, I'm looking at adding this and have an untested implementation. However, we're currently unable to test this due to the `mail` field being (a) readonly and (b) only populated for users with an O365 license. I'll look further into this.

trajano:
I'm actually just using the Terraform azuread to help manage my Office 365 installation (rather than an Azure system)

trajano:
Our workaround for now

```
data "azuread_user" "ContractorDevelopers" {
  for_each = merge(
    { for e in csvdecode(file("data/ContractorDevelopers.csv")) : e.Email => "${replace(e.Email, "@", "_")}#EXT#@${var.tenant_domain}" }
  )
  user_principal_name = each.value
}
```

Where the CSV file contains e-mail addresses only. no other fields.

PR #374, opened 2020-12-16
user: support the `mail` property

ghostinushanka:
With this Terraform operators will be able to also set/modify the primary email address for the users.

Change explicitly depends on acceptance of https://github.com/Azure/azure-rest-api-specs/pull/12127 which will make the Mail field available for updates.

### DIFF
--- a/internal/services/aadgraph/user_resource.go
+++ b/internal/services/aadgraph/user_resource.go
@@ -85,8 +85,10 @@ func userResource() *schema.Resource {
 			},
 
 			"mail": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validate.StringIsEmailAddress,
 			},
 
 			"onpremises_sam_account_name": {
@@ -201,7 +203,13 @@ func userResourceCreate(d *schema.ResourceData, meta interface{}) error {
 	ctx := meta.(*clients.AadClient).StopContext
 
 	upn := d.Get("user_principal_name").(string)
+	mail := d.Get("mail").(string)
 	mailNickName := d.Get("mail_nickname").(string)
+
+	//default mail to the upn if not set to a different address
+	if mail == "" {
+		mail = upn
+	}
 
 	//default mail nickname to the first part of the UPN (matches the portal)
 	if mailNickName == "" {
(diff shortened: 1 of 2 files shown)


manicminer:
Hi @ghostinushanka, thanks for this suggestion. Unfortunately the `mail` field is readonly and only populated/valid for Office 365 tenants. It's not possible to set via the legacy API - although it is documented and included in the SDK (for create operations only), an error is thrown if you do try to set it. There is an issue open #280 tracking this.

Regrettably this will not work and we won't be able to merge this. Since we have a lot of work in progress to migrate to the newer MS Graph API - where setting this field might be more feasible - I'm going to close this. Thanks again for submitting.

ghostinushanka:
Hello @manicminer, you have me at a loss here. I do know that the SDK currently supports the create-only, that is why I have referred the pull request of the colleague of mine, where we have added support for the update operation as well.

You mention that the field is supported by the "Office 365 tenants" and in the #280 you mention "O365 license" - that is the confusing part for me. 
Internally we are using the modified provider with the modified version of the SDK right now and it is working, and, I have tested mail modification with "Azure AD Free" tenant created from the landing page of azure services and "Enterprise P2" we have.
Could you please point me to the part of the documentation explaining the differences you have mentioned? Lack of information/understanding on my end comes to mind and would like to know more. Thank you.

manicminer:
Hi @ghostinushanka

Sorry for the confusion from my choice of phrasing. I was referring to the Azure AD SKUs as per https://azure.microsoft.com/en-us/pricing/details/active-directory/

I've dug up an old branch where I was testing setting the `mail` property, and you're correct it is now accepted by the API. I had previously tried setting it for users in a Free tenant and in an O365 tenant, and it was rejected by the API (this wasn't documented by Microsoft). However that seems to have changed for the better.

I'm happy to reopen this, on the basis that we could look to merge it when/if the upstream PR is merged and released.

For some additional context, there's ongoing work to move away from the Azure AD Graph API to the Microsoft Graph API. Generally speaking, we're not looking to merge in new features unless they translate well during the API migration - this would qualify if the swagger/SDK can be fixed up.

manicminer:
Hi @ghostinushanka, it looks like the swagger change was merged but hasn't made it into the SDK as yet.

As AzureAD v1.5.0 has now been released with initial Microsoft Graph support, features using the Azure SDK are now frozen. Whilst you cannot currently update the `mail` field for users, since for this release we are maintaining compatibility, this will become a writable field in version 2.0.

It's unfortunate the SDK change didn't make it sooner, but as we will no longer be using the Azure SDK going forward I'm going to close this PR. You can expect to see this fixed in v2.0 using the new API. Thanks again for the effort :)

Issue #323, opened 2020-09-24
Announcement - Upcoming changes in AzureAD v2.0 and Microsoft Graph Support


manicminer:
Our goal for the AzureAD provider is to transition to using [Microsoft Graph](https://developer.microsoft.com/en-us/graph) for all resources, and eventually drop support for the deprecated [Azure Active Directory Graph](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-graph-api) API. This involves building out SDK support and working out a strategy for migrating existing configurations. We are not planning any new features using the old API.

The provider recently had a v1.0 release which resolved some outstanding issues with the currently implemented resources. Some groundwork was also laid to enable us to implement multiple services, similarly to AzureRM. We'll be able to support resources backed by either API, in the same provider version.

## Proposal

Whilst we work to achieve feature parity when using Microsoft Graph, we'll introduce partial support using a feature toggle. This will be implemented either as provider configuration, as an environment variable, or both, and will enable operators to switch to Microsoft Graph for the resources that support it. This will look something like:

```hcl
provider "azuread" {
  enable_msgraph = true
}
```

Toggling with an environment variable would look something like `AAD_PROVIDER_ENABLE_MSGRAPH=true`.

Once the provider has feature parity using MS Graph, a major release v2.0 will be readied and we'll switch to using MS Graph by default.

## Benefits

- The Azure Active Directory Graph is deprecated and will at some point be switched off.
- Microsoft Graph models resources much closer to their current implementation than Azure Active Directory Graph, which has been, to a degree, feature frozen and unable to maintain compatibility in some cases.
- Lots more resources will be available to implement in the provider, for example
    - fully featured applications
    - expanded service principal configuration
    - guest users
    - directory roles
    - directory schema extensions
    - dynamic group membership
    - entitlement management
    - privileged identity management
    - B2C
    - Office 365 features such as Mail, Calendars, OneDrive, SharePoint, Teams

## Caveats

- Existing configurations may require modification in places to work seamlessly with the next major version (this is fairly common for any major release).
- It's not clear at this time whether we'll be able to continue supporting Azure CLI authentication, although we will strive to do so.

## Plan

- [ ] Investigate feature toggle by provider config and/or environment variable
- [ ] Build SDK support (progressing)
- [ ] Get feature parity with current resources
- [ ] Work out and document a migration guide
- [ ] Major release
- Additional features and resources 🚀 

---

[...]

PR #378, opened 2021-01-13
Draft:Add manager

KqLLL:


manicminer:
Hi @KqLLL, thanks for this PR.

In some cases like this, where the SDK doesn't support the functionality we need, we might add a shim to the provider to get the coverage we need, or to fix SDK handling bugs. We could feasibly do that here, however the PR is seeking to add a feature against the deprecated API which we are in the process of replacing - see #323 for details. This API is also purported to be switched off entirely by Microsoft in June, so releasing new features against it is unfortunately fighting against the tide at this point.

As such, I've opened a feature request for this in #381 and we'll seek to add this functionality using Microsoft Graph in the near future. Please upvote that issue by adding a thumbsup reaction as we'll be ranking feature requests by popularity. We would also welcome a contribution if you are up for implementing this using the new API - once we have released the next version of the provider containing the pieces needed.

Thanks again, unfortunately we won't be able to merge this at present and so accordingly I'm going to those this PR.

PR #389, opened 2021-01-22
Update `examples/application` name to display_name

gkman:
Updates `examples/application/main.tf` from `name` which is depreciated to `display_name`


manicminer (approved):
Thanks @gkman, this LGTM :+1:

Issue #307, opened 2020-08-20
Support guest user invitations


manicminer:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support Guest user invitations for AAD tenants.

### New or Affected Resource(s)

* azuread_guest_user

### References

* #297

romainDavaze:
Hello, I'd like to add some comments on this.

Please note that the feature implemented in #401 only allows to create user resources with the user type as `Guest`. This is not considered as a user invitation.
In fact, when you look at your Azure AD users, you will see that the `User Type` value is `Guest` as expected but the `Creation Type` remains empty (we need it to be set at `Invitation`). This means that the API allows you to create guest users but they are not really considered as such because it does not trigger the invitation process.

It is problematic because some features aren't working with this approach. We originally implemented this feature because we needed to invite users in order to use the External Identities feature (it is a prerequisite for this to work).

For us the only way to create guest users properly is to use the `/invitations` graph's API endpoint but this needs to be implemented in the go sdk before we can use it here. See https://github.com/Azure/azure-sdk-for-go/issues/4086

manicminer:
Hi @romainDavaze, thanks for noting that. Supporting the `/invitations` API in MS Graph is on our roadmap (for context see #323) and will likely land in v2.0. I've reopened this issue to track it.

Can I ask which features don't work, apart from the invitation process itself? I'd like to figure out whether it's worth supporting setting the userType property versus only supporting guests via the invitations API.

romainDavaze:
Thank you for your response, I'll keep an eye on this issue then.

I've only encountered the issue with external identities so far because I only use guest users for this. Because this feature needs you to invite users to your AAD prior to be able to use an external IDP, I guess that just by creating a user resource, we're not creating a guest user from Azure's point of view, or at least not completely. Maybe it has something to do with the `Source` property when you look at the user profile. When you do it via Terraform, it is empty whereas `Invited User` is mentioned when you do it from the portal.
So for me, any feature that requires a guest user will not work this way.

[...]

Issue #41, opened 2019-02-12
azuread_group: Support for AzureAD B2B user invitations (guests)


daniel-simpson:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Building on [this comment](https://github.com/terraform-providers/terraform-provider-azuread/issues/36#issuecomment-462155004), I would like to be able to add guest users to a group using AzureAD B2B.

My use case for this is a clean separation of concerns between user authentication (handled in a separate tenant) and authorization (RBAC on groups).  There are Microsoft Graph APIs available for:
- [Get User in Tenant](https://docs.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0)
- [Add user to tenant (via invitation)](https://docs.microsoft.com/en-us/graph/api/invitation-post?view=graph-rest-1.0)
- [Delete user from tenant](https://docs.microsoft.com/en-us/graph/api/user-delete?view=graph-rest-1.0)

The main part I can't get my head around is whether or not this would require an `azuread_guest` resource type or if we can infer it using the first snippet below.

If we don't need a new `azuread_guest` resource type, the side effect of extra guest users in this tenant is mitigated by the fact that once removed from all groups, the invited guest's permissions are effectively zero, though I'm aware that in Terraform we don't like things to be implicit 😄 

### New or Affected Resource(s)

* azuread_group
* (Optional) azuread_guest

### Potential Terraform Configuration

```hcl
resource "azuread_group" "mygroup" {
  name = "MyGroup"
  # owners = ["foo.bar@example.com"]
  # members = ["foo.bar2@example.com"]
  guests = ["externalAadUser@test.onmicrosoft.com"]
}
```

OR

```hcl
resource "azuread_guest" "guest_invite" {
  email = "test@mail.com"
}

resource "azuread_group" "mygroup" {
  name = "MyGroup"
  members = ["${azuread_guest.guest_invite.id}"]
}
```

### References

* Depends on #8 (merged)
* Original comment on #36

a138076:
Would be very interested in this functionality!

daniel-simpson:
[...]

PR #401, opened 2021-02-15
added support for guest users

lde:
Allow creation of guests users in azure active directory.

Refers issues #307 and #41

### DIFF
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -43,6 +43,7 @@ The following arguments are supported:
 * `surname` - (Optional) The user's surname (family name or last name).
 * `usage_location` - (Optional) The usage location of the User. Required for users that will be assigned licenses due to legal requirement to check for availability of services in countries. The usage location is a two letter country code (ISO standard 3166). Examples include: `NO`, `JP`, and `GB`. Cannot be reset to null once set. 
 * `user_principal_name` - (Required) The User Principal Name of the User.
+* `user_type` - (Optional) - The user type in Directory. Valid values are `Guest` and `Member`. Defaults to `Member`.
 
 ## Attributes Reference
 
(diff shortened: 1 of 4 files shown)


internal/services/users/user_resource.go
```diff
@@ -195,6 +195,12 @@ func userResource() *schema.Resource {
 				Computed:    true,
 				Description: "The primary cellular telephone number for the user.",
 			},
+			"user_type": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Computed:         true,
```
katbyte, line 201: should we default to member instead of computed here?

katbyte (changes requested):
Hey @lde - thanks for the PR. one question i've left inline about default/computed and other then that we need to update the docs too

internal/services/users/user_resource.go
lde: This field isn't mandatory in api, I won't override default behaviour set by Microsoft if user set no value.
Is ok for your ?

lde:
Hi,
I've already updated the documentation by adding the property in user.md. Do yo see more to to ?

I've replied to your inline comment. 

Thanks

internal/services/users/user_resource.go
manicminer: I think it would be more predictable to have a default here - guest users should probably be explicitly configured as such and the intention is clearer if we default to a regular user.

docs/resources/user.md
```diff
@@ -43,6 +43,7 @@ The following arguments are supported:
 * `surname` - (Optional) The user's surname (family name or last name).
 * `usage_location` - (Optional) The usage location of the User. Required for users that will be assigned licenses due to legal requirement to check for availability of services in countries. The usage location is a two letter country code (ISO standard 3166). Examples include: `NO`, `JP`, and `GB`. Cannot be reset to null once set. 
 * `user_principal_name` - (Required) The User Principal Name of the User.
+* `user_type` - (Optional) - The user type in Directory. Valid values are `Guest` and `Member`.
```
manicminer, line 46:
```suggestion
* `user_type` - (Optional) - The user type in the directory. Valid values are `Guest` and `Member`. Defaults to `Member`.
```

manicminer (changes requested):
Thanks for the PR @lde, I have some other minor comments mostly just formatting related. This tests well and if we can adjust the schema per @katbyte's suggestion and the comments below, this should be good to merge!

manicminer (approved):

manicminer:
Fixes #41
Fixes #307

manicminer:
Test results

Screenshot 2021-02-23 at 11 59 59

PR #404, opened 2021-02-22
Documentation fix for azuread_group

jhawthorn22:
Company Administrator and User Account Administrator roles don't exist anymore. It should be Groups Administrator now.


manicminer (approved):
Thanks for this documentation fix @jhawthorn22, much appreciated!

PR #406, opened 2021-02-24
data.azuread_user: export the user_type attribute

manicminer:
Also add documentation note to explain the problem with being forced to set a password for guest users.

### DIFF
--- a/internal/services/users/user_data_source.go
+++ b/internal/services/users/user_data_source.go
@@ -158,8 +158,9 @@ func userData() *schema.Resource {
 			},
 
 			"user_type": {
-				Type:     schema.TypeString,
-				Computed: true,
+				Type:        schema.TypeString,
+				Computed:    true,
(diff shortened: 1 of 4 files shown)


internal/services/users/user_data_source.go
```diff
@@ -156,6 +156,11 @@ func userData() *schema.Resource {
 				Computed:    true,
 				Description: "The primary cellular telephone number for the user.",
 			},
+
+			"user_type": {
+				Type:     schema.TypeString,
+				Computed: true,
```
jackofallops, line 162: The other schema items have the description field, is it worth including here for completeness?

jackofallops (approved):
Thanks @manicminer - just one question inline, otherwise LGTM 👍

internal/services/users/user_data_source.go
manicminer: Sure, can do 👍

PR #411, opened 2021-03-04
Add a client_certificate implementation to create-for-rbac example module

manicminer:


jackofallops (approved):
Thanks @manicminer - LGTM 👍

PR #412, opened 2021-03-08
fix gomodule export path & update to Go 1.16

jackofallops:
Fixes go.mod post org move.

Bumps Go version to 1.16 in line with AzureRM


manicminer (approved):
Awesome, thanks @jackofallops! LGTM 🚀

PR #413, opened 2021-03-08
Revert support for setting `user_type` to `Guest`

manicminer:
- Adding users in this way bypasses the invitations API and creates broken accounts
- No invitation email is sent, which is sort of expected but not great UX
- The guest user can browse to the inviting tenant in the Portal, but cannot see it, or any linked subscriptions, in their global filter list
- AAD roles assigned to such a user do not work
- RBAC roles assigned to such a user do not work
- Subscriptions do not work

We will reserve this functionality for MS Graph when we have access to the invitations API which processes guest accounts correctly.

Example:

Screenshot 2021-03-08 at 16 22 20


jackofallops (approved):
Thanks @manicminer - LGTM 👍

PR #415, opened 2021-03-10
📝 Fix wrong parameter in `azuread_application` resource

JCZuurmond:
The `name` in the `azuread_application` resource mentioned with the `azuread_service_principal` docs should be `display_name`.


manicminer (approved):
Hi @JCZuurmond, many thanks for this correction, this LGTM!

PR #426, opened 2021-04-21
Support URNs properly

### LINKED ISSUE #425: Application identifier_uris no longer supports URNs and string values that are not considered a URL
### Community Note

terraform: v0.13.4
azuread_provider  ≥  1.2.0

### Affected Resource(s)

* `azuread_application `

```hcl
resource "azuread_application" "application" {
  name    = "hello_world"
  homepage     ="http://helloworld"
  identifier_uris = ["urn:uuid:41055901-ff51-46ee-bcf8-c37d0c3e484b"]
  reply_urls     = ["https://helloworld/acs/reply"]
  available_to_other_tenants = false
  oauth2_allow_implicit_flow = false
  type  = "webapp/api"
}
```

### Debug Output

```
Error: URL has no host

  on ../src/tf/modules/tf-azuread-saml/main.tf line 11, in resource "azuread_application" "application":
  11:   identifier_uris            = var.identifier_uris
```

### Expected Behavior

Both AzureAD UI and the Microsoft Graph happily accept a string for the EntityID/Identifier_uris. Any value (providing it is unique across the tenant) should be accepted and was up until and including version 1.1.1. 

### Actual Behavior

When using something other than a url format, terraform throws an error of `URL has no host` and it doesnt even tell you  what host/application has the issue. 

### Steps to Reproduce

1. `terraform apply`

### References

I believe the issue is shown via this line of code: https://github.com/hashicorp/terraform-provider-azuread/blob/v1.2.2/internal/validate/url.go#L57-L63 and is caused by this: https://github.com/hashicorp/terraform-provider-azuread/blob/v1.2.2/internal/validate/url.go#L46 

We currently use this provider to setup and create app registrations and enterprise apps used for SAML. We use this for both internal and external apps and cannot force a 3rd party to switch from a random string or URN to a url. 

I've spoken to @manicminer via slack about this issue and he should have some understanding of it.

buffer51:
Fixes #425 by refactoring the URL validation to support both URLs & URNs.

I looked for Go packages that do URN validation, and found [voicera/gooseberry urn#TryParseString](https://pkg.go.dev/github.com/voicera/gooseberry@v0.0.0-20181223025147-dc233900870c/urn#TryParseString) and [leodido/go-urn#Parse](https://pkg.go.dev/github.com/leodido/go-urn#Parse). Neither seemed worth adding as a dependency, so I used the simple validation logic from the former.

### DIFF
--- a/internal/services/applications/application_resource.go
+++ b/internal/services/applications/application_resource.go
@@ -139,7 +139,7 @@ func applicationResource() *schema.Resource {
 				Type:             schema.TypeString,
 				Optional:         true,
 				Computed:         true,
-				ValidateDiagFunc: validate.URLIsHTTPOrHTTPS,
+				ValidateDiagFunc: validate.IsHTTPOrHTTPSURL,
 			},
 
 			"identifier_uris": {
@@ -148,7 +148,7 @@ func applicationResource() *schema.Resource {
 				Computed: true,
 				Elem: &schema.Schema{
 					Type:             schema.TypeString,
-					ValidateDiagFunc: validate.URLIsAppURI,
+					ValidateDiagFunc: validate.IsAppURI,
 				},
 			},
 
@@ -156,7 +156,7 @@ func applicationResource() *schema.Resource {
 			"logout_url": {
 				Type:             schema.TypeString,
 				Optional:         true,
-				ValidateDiagFunc: validate.URLIsHTTPOrHTTPS,
+				ValidateDiagFunc: validate.IsHTTPOrHTTPSURL,
 			},
 
 			// TODO: v2.0 put this in an `implicit_grant` block and rename to `access_token_issuance_enabled`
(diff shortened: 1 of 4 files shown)


buffer51:
I mistakenly had assumed URNs have 3 parts, but it's 3 or more parts. If we wanted to be stricter we could use a URN regex to make sure it's totally valid per the RFC (Google suggests a few), but we can't actually validate the parts (we'd need a library of known URN namespaces, which feels overkill).

Let me know what you think is more appropriate.

manicminer:
@buffer51 Thanks for updating. A simple string split is perfectly fine, we don't want to be overly strict and I don't think it's necessary to wield a regex 🙂

buffer51:
Brilliant @manicminer 😄 If I may, what is the release cadence for this project?

manicminer:
That's difficult to characterize as we're working through the Microsoft Graph transition at the moment, but there is hopefully a release forthcoming in the next week or so.

manicminer (approved):
Thanks again @buffer51 and @MarkDordoy, this LGTM 👍

PR #429, opened 2021-05-04
Minor Documentation Correction: git clone URI should be a valid URI

rikkuness:
git clone URI is missing protocol prefix


manicminer (approved):
Thanks for the fixup @rikkuness, this LGTM 👍

PR #434, opened 2021-05-14
Migrate from GitHub HashiBot lock behavior to GitHub Actions

bflad:
There is not an equivalent for the `release_commenter` behavior yet to fully remove the configuration (unless you are okay without it for now, then happy to just remove the whole thing 👍 ).


manicminer (approved):
Awesome, thanks @bflad! 👍

PR #439, opened 2021-05-21
Bugfix: suppress spurious deprecation notice for `metadata_host` provider field even when not set

### LINKED ISSUE #437: The `metadata_host` provider attribute is deprecated and will be removed in version 2.0
hey all,

Since updating to version 1.5.0, we have started seeing the following warnings: 

```
 Warning: Deprecated Attribute
│
│   with provider["registry.terraform.io/hashicorp/azuread"],
│   on _generated_global_providers.tf line 26, in provider "azuread":
│   26: provider "azuread" {
│
│ The `metadata_host` provider attribute is deprecated and will be removed in
│ version 2.0
```

https://github.com/hashicorp/terraform-provider-azuread/blob/e9f389f73d419c94cb9bf036f38878877b0439db/internal/provider/provider.go#L90

We are not defining the metadata_host attribute on the provider block or setting its corresponding environment variable. 

It is not something that's bugging us massively, but it would be nice to not have it showing up in our pipelines until v2 is released.

manicminer:
Provider schema validation raises a warning diagnostic for deprecated fields even if a property isn't present in configuration (or it's corresponding env var is also not set). This removes the Deprecation field from `metadata_host` and updates the documentation.

Closes #437


katbyte (approved):
LGTM 🚀

PR #448, opened 2021-06-02
microsoft-graph.md: Simplify PowerShell command

bergmeister:
Double quotes are for interpolated strings, for constant strings, single quotes should be used, see https://github.com/PowerShell/PSScriptAnalyzer/blob/master/RuleDocumentation/AvoidUsingDoubleQuotesForConstantString.md
In this instance, not even a single quote is required as PowerShell knows that the `env` PSDrive only takes strings and converts the integer to a string for you


manicminer (approved):
Thanks @bergmeister, LGTM!

Issue #453, opened 2021-06-07
azuread_group: cannot create group with more than 20 owners or members


manicminer:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v0.15.5
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v1.5.0
```

### Affected Resource(s)

* `azuread_group`

### Terraform Configuration Files

```hcl
provider "azuread" {
  use_microsoft_graph = true
}

data "azuread_domains" "test" {
  only_initial = true
}

resource "azuread_user" "test" {
  count = 25

  user_principal_name = "testing-relationships-${count.index}@${data.azuread_domains.test.domains.0.domain_name}"
  display_name        = "Testing Relationships ${count.index}"
  password            = "Qwer5678!@#"
}

resource "azuread_group" "test" {
  display_name = "Testing Relationships"
  owners       = azuread_user.test.*.object_id
  members      = azuread_user.test.*.object_id
}
```

### Error Output

```
╷
│ Error: Creating group "Testing Relationships"
│
│   with azuread_group.test,
│   on test.tf line 17, in resource "azuread_group" "test":
│   17: resource "azuread_group" "test" {
│
│ GroupsClient.BaseClient.Post(): unexpected status 400 with OData error: Request_BadRequest: Exceeded changed links limit (20)
╵
```

### Expected Behavior

The group should be created with the specified owners and/or members

### Actual Behavior

The group is not created at all

### Steps to Reproduce

1. `terraform apply` ensuring the group does not yet exist
2. Observe error

### References

* Originally reported in https://github.com/manicminer/hamilton/issues/47

PR #454, opened 2021-06-07
Bugfix: Enable setting >20 group owners or members at create time

manicminer:
When creating a group having more than 20 owners or members, the group creation fails. This mitigates the problem by not including the owners or members in the initial Create request, then adding both the specified owners and members afterwards.

To ensure the group can continue to be managed in cases where write access to the group is dependent on the authenticated principal being an owner (which might ultimately be via group memberships), the caller's object ID is specified as the sole owner when the group is created, and then removed afterwards if that same principal object ID was not included in the list of owners in the configuration.

Fixes: #453
Related: https://github.com/manicminer/hamilton/issues/47


manicminer:
Test result for `TestAccGroup_manyMembersAndOwners`

Screenshot 2021-06-08 at 01 44 50

katbyte (approved):
LGTM 🐈

Issue #442, opened 2021-05-24
oauth2_permission_scope requires field id


Theuno:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
**Terraform v0.14.9**
provider registry.terraform.io/hashicorp/**azuread v1.5.0**

### Affected Resource(s)

* `azuread_application `

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    azuread = {
      source = "hashicorp/azuread"
      version = "1.5.0"
    }
  }
}

provider "azuread" {
  # Configuration options
}

data "azuread_client_config" "current" {}

resource "azuread_application" "bugDemo1" {
  display_name = "bugDemo1"
  prevent_duplicate_names = true

  api {
    oauth2_permission_scope {
      admin_consent_display_name = "Allow User Impersonation"
      // id                         = "000-000-000"
      type                       = "Admin"
      user_consent_description   = "Enables the client application to call the Contacts API on your behalf"
      user_consent_display_name  = "Allow User Impersonation"
      value                      = "access_as_user"
    }
  }
}
```

### Expected Behavior

As the ID is not known before creation of the oauth2_permission_scope, it is not expected that this is a *required* parameter in this context.

### Actual Behavior

Can't create a scope without the id setting, using a generated id doesn't work.
Using 1.4.0 of the azuread plugin, with `oauth2_permissions` the id field is not required.

### Steps to Reproduce

1. `terraform apply` without an id.

### Important Factoids

n/a

### References

n/a

manicminer:
Hi @Theuno, thanks for raising this issue!

We made this change in order to overcome a conflict between Terraform schema limitations and the specific API semantics. Previously we'd generate a UUID and send that in the API request, however in the new `oauth2_permission_scope` block you'll need to specify a UUID. It can be any valid UUID, as long as it's unique inside your _tenant_.

[...]

PR #457, opened 2021-06-08
Improve docs around scope IDs for applications

manicminer:
Also update deprecated properties in examples

Resolves: #442


jackofallops (approved):
LGTM 👍

Issue #451, opened 2021-06-07
New Resource for Authorized client applications for Oauth 2 permission scope


drdamour:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

kinda surprised i didn't see another request for this yet, so maybe i'm missing it in the docs or the name in another request.  This is assigning permissions scopes to specific authorized clients.  The main benefit being supporting Azure.Identity's visual studio provider to work.

This is the area here
![image](https://user-images.githubusercontent.com/1514496/121071443-2f21ea00-c78d-11eb-9ae9-81e3409b6786.png)

### New or Affected Resource(s)

* azuread_application_oauth2_permission_authorized_client

### Potential Terraform Configuration

```hcl

resource "azuread_application" "example" {
  name = "example"
}

resource "azuread_application_oauth2_permission_scope" "example1" {
  application_object_id      = azuread_application.example.id
  admin_consent_description  = "Administer the application 1"
  admin_consent_display_name = "Administer 1"
  enabled                    = true
  type                       = "User"
  user_consent_description   = "Administer the application 1"
  user_consent_display_name  = "Administer 1"
  value                      = "administer1"
}

resource "azuread_application_oauth2_permission_scope" "example2" {
  application_object_id      = azuread_application.example.id
  admin_consent_description  = "Administer the application 2"
  admin_consent_display_name = "Administer2"
  enabled                    = true
  type                       = "User"
  user_consent_description   = "Administer the application 2"
  user_consent_display_name  = "Administer2"
  value                      = "administer2"
}

azuread_application_oauth2_permission_authorized_client "example"
{
    client_id = "872cd9fa-d31f-45e0-9eab-6e460a02d1f1"
    scope_ids = [
       azuread_application_oauth2_permission_scope.example1.id,
       azuread_application_oauth2_permission_scope.example2.id,
    ]
}
```

assuming the api follows the gui which lets you assign many at one:
![image](https://user-images.githubusercontent.com/1514496/121072259-33023c00-c78e-11eb-8157-e24fce96dc2e.png)

[...]

Issue #263, opened 2020-06-01
Add support for pre authorized applications in an Azure AD app registration


TechWatching:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add support for pre authorized applications in an Azure AD app registration. Applications in Azure AD can expose an API with a scope and pre authorize some existing applications on that scope. This generates this kind of json in the application mantifest.

```json
	"preAuthorizedApplications": [
		{
			"appId": "5e3ce6c0-2arf-4285-8s23b-75ee78787346",
			"permissionIds": [
				"434b0aae-77b3-481a-b78f-4dfds29b"
			]
		},
		{
			"appId": "1fec8e78-bce4-4aaf-dsq12b-5451cc387264",
			"permissionIds": [
				"434b0aae-77b3-481a-b78f-4dfds29b"
			]
		}
```

### New or Affected Resource(s)

* azuread_application
* azuread_serviceprincipal

daniel-chambers:
I'd like to comment that I think this should be a separate resource to `azuread_application`. In my case, I have two applications, a SPA and an API. The SPA app depends on the API as a required resource. But the API depends on the SPA as a pre-authorized application. 
Having pre-authorized applications be a separate resource would break this cycle.

PR #472, opened 2021-06-30
New resource: application_pre_authorized

manicminer:
This is implemented as a standalone resource as it references oauth2 permission scope IDs on the same application, that must exist, and be enabled, prior to setting the `preAuthorizedApplications` property.

Note: PR currently targets `v2` branch, but will change to `main` once #461 is merged

Closes: #263
Closes: #451

### DIFF
--- a/docs/resources/pre_authorized_application.md
+++ b/docs/resources/pre_authorized_application.md
@@ -1,0 +1,71 @@
+---
+subcategory: "Applications"
+---
+
+# Resource: azuread_pre_authorized_application
+
+Manages client applications that are pre-authorized with the specified permissions to access an application's APIs without requiring user consent.
+
+## Example Usage
+
+```terraform
+resource "azuread_application" "authorized" {
+  display_name = "example-authorized-app"
+}
+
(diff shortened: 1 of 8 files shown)


docs/resources/pre_authorized_application.md
```diff
@@ -0,0 +1,71 @@
...
+      id                         = "2d5e07ca-664d-4d9b-ad61-ec07fd215213"
+      type                       = "User"
+      user_consent_description   = "Access the application"
+      user_consent_display_name  = "Access"
+      value                      = "user_impersonation"
+    }
+  }
+}
+
+resource "azuread_pre_authorized_application" "example" {
```
katbyte, line 42: should this be renamed to
```suggestion
resource "azuread_application_pre_authorized" "example" {
```

to sort better with application?

katbyte (approved):
aside from one comment LGTM 🍰

docs/resources/pre_authorized_application.md
manicminer: Good call, will do 👍

Issue #469, opened 2021-06-25
azuread_application - block for all platform configurations.


JordanW94:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I've just recently started using the azuread_application resource and I'm in need of accessing Mobile and desktop applications platform configuration. It seems the web block is the only available platform for Terraform to configure.

### New or Affected Resource(s)

azuread_application

### Potential Terraform Configuration

```hcl
resource "azuread_application" "azuread" {
  display_name     = "namehere"
  owners = [object_id]
  sign_in_audience = "AzureADMyOrg"
  prevent_duplicate_names = true

  mobile_desktop_platform {
    redirect_uri = ["com.terraform.app://oauth", "http://localhost:300/"]
  }
}

```

Would this be possible?

### References
https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/application

* #0000

manicminer:
Hi @JordanW94, thanks for the feature request. You are correct, at this time the provider can only configure some properties of the web block for application. This is due to API limitations, which we are currently working to move beyond by transitioning to a newer API (see #323 for details).

This work is ongoing at the moment and we hope to be able to release full support for different platform settings for applications in the near future.

JordanW94:
Thanks for the quick reply @manicminer, I thought that would be case. I look forward to seeing the new features!

Issue #462, opened 2021-06-15
App registration should accept on_premise_security_identifier as additional property


dvdvorle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

```
Terraform v1.0.0
on windows_amd64
+ provider registry.terraform.io/hashicorp/azuread v1.5.1
```

### Terraform Configuration Files

```hcl
resource "azuread_application" "application" {
  display_name            = var.name
  group_membership_claims = "SecurityGroup"

  # Onderstaande is niet mogelijk omdat de property on_premise_security_identifier niet ondersteund wordt.
  # Daarom hieronder maar met een script

  optional_claims {
    access_token {
      name                  = "groups"
      source                = null
      essential             = false
      additional_properties = ["on_premise_security_identifier", "emit_as_roles"]
    }
  }
}
```

### Expected Behavior
In portal terms:
![image](https://user-images.githubusercontent.com/175623/122073814-020eb080-cdf9-11eb-9022-58ea15530f86.png)

In app registration manifest terms:
```
{
    ... snip ...
    "optionalClaims": {
        "accessToken": [
	    {
	        "additionalProperties": [
		    "on_premise_security_identifier",
		    "emit_as_roles"
		],
		"essential": false,
		"name": "groups",
		"source": null
	    }
    ]
    ... snip ...
}
```

### Actual Behavior

│ Error: expected optional_claims.0.access_token.0.additional_properties.0 to be one of [dns_domain_and_sam_account_name emit_as_roles include_externally_authenticated_upn include_externally_authenticated_upn_without_hash netbios_domain_and_sam_account_name sam_account_name use_guid], got on_premise_security_identifier

### Steps to Reproduce

1. `terraform apply`

Issue #436, opened 2021-05-21
Add support for publicClient


czmirek:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Please add support for the `publicClient` property [described in the MS Graph docs](https://docs.microsoft.com/en-us/graph/api/resources/application?view=graph-rest-1.0). This should be possible after moving this provider to MS Graph.

```json
{
   "publicClient": {
      "redirectUris": [ "http://example.com" ]
   }
}
```

Right now, only the web redirectUris are supported.

```json
{
    "web": {
        "redirectUris": [ "http://example.com"],
    }
}
```

### New or Affected Resource(s)

* azuread_application

### Potential Terraform Configuration

```hcl
resource "azuread_application" "app" {
  display_name = "My application"
  public_client {
     redirect_uris = [ "http://example.com" ]
  }
}
```

manicminer:
Thanks for requesting @czmirek, we'll look to add this in v2.0 as it requires an incompatible schema change.

Issue #320, opened 2020-09-18
Add azuread_application support for termsOfServiceUrl/privacyStatementUrl


JulienRavet:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Please add support for "infos" properties of Azure AD applications
```
"info": {
        "termsOfServiceUrl": null,
        "supportUrl": null,
        "privacyStatementUrl": null,
        "marketingUrl": null,
        "logoUrl": null
    },
```

### Affected Resource(s)
* azuread_application

manicminer:
Thanks @JulienRavet for requesting these attributes. These depend on upcoming support for Microsoft Graph and we'll look at adding them once we have that.

JulienRavet:
Hi @manicminer 

The application logo seems to be available in the graph API (even if it's not one one the requested in the issue title :wink:)
https://docs.microsoft.com/fr-fr/previous-versions/azure/ad/graph/api/entity-and-complex-type-reference#application-entity 

mainLogo | Edm.Stream | POST, GET, PATCH | The main logo for the application.Notes: not nullable

PR #474, opened 2021-06-30
Application improvements

manicminer:
- Case insensitive matching on UUIDs for passwords and certificates
- Role/scope validation also checks IDs now that users must set it themselves
- Extensive whole-resource validation when `sign_in_audience` is changed to include personal accounts, to avoid breakage
- Updates Hamilton to v0.20.0 for fixes, features and improved odata query support

### New Applications Features
- Support additional properties in the `api` block
  - `accept_mapped_claims`
  - `known_client_applications`
  - `requested_access_token_version`
- Export the `disabled_by_microsoft` attribute
- Support the `logo_url`, `marketing_url`, `privacy_statement_url`, `support_url` and `terms_of_service_url` properties
- Support the `device_only_auth_enabled` property
- Support the `oauth2_post_response_required` property
- Export the `publisher_domain` attribute (this is readonly, awaiting feedback on how we can set this)
- Support the `public_block` block
- Support the `single_page_application` block
- Export two computed typemap attributes: `app_role_ids` and `oauth2_permission_scope_ids` to alleviate typeset handling with config boilerplate

Closes: #188
Closes: #244
Closes: #286
Closes: #320 
Closes: #436
Closes: #462
Closes: #469
Closes: #477


internal/services/applications/application_data_source.go
```diff
@@ -42,6 +43,12 @@ func applicationDataSource() *schema.Resource {
 				ValidateDiagFunc: validate.UUID,
 			},
 
+			"disabled_by_microsoft_status": {
```
katbyte, line 46: i don't think we need status?
```suggestion
			"disabled_by_microsoft": {
```

internal/services/applications/application_data_source.go
```diff
@@ -56,6 +63,21 @@ func applicationDataSource() *schema.Resource {
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"accept_mapped_claims": {
```
katbyte, line 66: should we use this for consistency with other bools?
```suggestion
						"mapped_claims_enabled": {
```

internal/services/applications/application_resource.go
```diff
@@ -225,6 +274,51 @@ func applicationResource() *schema.Resource {
 				},
 			},
 
+			"info": {
```
katbyte, line 277: info seems superfluous - should we just move these all to the top level isntead?

internal/services/applications/application_resource.go
```diff
@@ -225,6 +274,51 @@ func applicationResource() *schema.Resource {
...
+							Description: "URL of the application's terms of service statement",
+							Type:        schema.TypeString,
+							Optional:    true,
+						},
+					},
+				},
+			},
+
+			"oauth2_post_response_required": {
+				Description: "Specifies whether, as part of OAuth 2.0 token requests, Azure AD allows POST requests, as opposed to GET requests.",
```
katbyte, line 317: this says "post allowed" vs the property name "response_required"

internal/services/applications/application_resource.go
```diff
@@ -248,6 +342,25 @@ func applicationResource() *schema.Resource {
 				},
 			},
 
+			"public_client": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"redirect_uris": {
```
katbyte, line 351: makes sense to pull this out to a top level property `public_client_redirect_uris`

internal/services/applications/application_resource.go
```diff
@@ -304,6 +417,25 @@ func applicationResource() *schema.Resource {
 				}, false),
 			},
 
+			"single_page_application": {
```
katbyte, line 420: and here? or will there be other properties there

katbyte (approved):
aside from a few comments on schema this looks good!

internal/services/applications/application_resource.go
manicminer: Agreed, that makes sense

internal/services/applications/application_resource.go
manicminer: This is a tricky one since the docs are ambiguous on the behaviour. I'll try to dig up the actual meaning and make it clearer one way or the other.

internal/services/applications/application_resource.go
manicminer: Yeah I'm sort of anticipating additional properties over time in both the `public_client` and `single_page_application` blocks, so was trying to preempt future deprecations here

internal/services/applications/application_resource.go
manicminer: It seems like this description, which was pulled from [application manifest docs](https://docs.microsoft.com/en-gb/azure/active-directory/develop/reference-app-manifest?WT.mc_id=Portal-Microsoft_AAD_RegisteredApps#oauth2requiredpostresponse-attribute), is the only documentation available on this one. Will follow up separately and try to find out more.

manicminer:
@katbyte I've added some additional validation to try and avoid applying bad application configurations; could you take another look before merging? Thanks! :)

manicminer:
This is unfortunately blocked due to a plugin SDK issue around block handling

manicminer:
Test results

Screenshot 2021-07-04 at 19 20 08

internal/services/applications/application_data_source.go
```diff
@@ -252,6 +349,35 @@ func applicationDataSource() *schema.Resource {
 				Computed:    true,
 			},
 
+			"single_page_application": {
```
katbyte, line 352: if this is a list shouldn't it be plural?
```suggestion
			"single_page_applications": {
```

internal/services/applications/application_resource.go
katbyte: any update on this one?

katbyte (approved):
aside from a couple comments LGTM 🍰

internal/services/applications/application_data_source.go
manicminer: This block is intended to contain the settings for the application when it's functioning as a single page app (the API container is just `spa` but I felt it needed expanding)

internal/services/applications/application_resource.go
manicminer: Not as yet, have reached out to get some clarity on this field

manicminer: I've raised this separately as #491 against the 2.0.0 milestone in order to unblock the PR stack :)

Issue #484, opened 2021-07-19
azuread_application v2.0 Error: missing expected [ when attempting to read state created by v1.6.0


conzyor34:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
$ terraform -v
Terraform v1.0.0
on linux_amd64
+ provider registry.localhost.com/local/azuread v2.0.0
+ provider registry.terraform.io/hashicorp/azuread v1.6.0
```

The v2.0 `azuread` provider was compiled from `main`.

### Affected Resource(s)

* `azuread_application`

#### Before

```hcl
terraform {
  required_providers {
    azuread = {
      source = "hashicorp/azuread"
      version = ">= 1.6.0"
    }     
  }
}

resource "azuread_application" "app_registration" {
  display_name = "test-app-registration"
  identifier_uris = ["https://website.company.com/"]
  sign_in_audience = "AzureADMyOrg"

  web {
    homepage_url = "https://website.company.com"
    redirect_uris = [
      "http://localhost/auth/redirect",
      "https://website.company.com/auth/redirect"
    ]

    implicit_grant {
      access_token_issuance_enabled = false
    }
  }
}
```

#### After Upgrade to v2.0

```hcl
terraform {
  required_providers {
    azuread = {
      source = "registry.localhost.com/local/azuread"
      version = ">= 2.0.0"
    }
  }
}

resource "azuread_application" "app_registration" {
  display_name = "test-app-registration"
  identifier_uris = ["https://website.company.com/"]
  sign_in_audience = "AzureADMyOrg"

  web {
    homepage_url = "https://website.company.com"
    redirect_uris = [
      "http://localhost/auth/redirect",
      "https://website.company.com/auth/redirect"
    ]

    implicit_grant {
      access_token_issuance_enabled = false
    }
  }
}
```

### Debug Output

https://gist.github.com/conzyor34/bba0541e90f1fce4fc657668f10478fc

### Expected Behavior

`No changes. Your infrastructure matches the configuration.`

### Actual Behavior

```
Error: missing expected [

  with azuread_application.app_registration,
[...]

PR #487, opened 2021-07-20
Add state migration for azuread_application to handle group_membership_claims type change

manicminer:
- Migrate `group_membership_claims` in `azuread_application` from a TypeString to a TypeSet
- Organize other state migrations into `migrations/` directory for each service

Fixes: #484

### DIFF
--- a/internal/services/applications/application_password_resource.go
+++ b/internal/services/applications/application_password_resource.go
@@ -3,7 +3,6 @@ package applications
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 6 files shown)


koikonom (approved):
LGTM

conzyor34:
Probably should change the log message in `migrations/application_resource.go`? It's changing `group_membership_claims` not `ID`.

`log.Println("[DEBUG] Migrating ID from v0 to v1 format")`

PR #490, opened 2021-07-21
Logging and inline client cert authentication

manicminer:
- Update Hamilton to v0.23.1
  - New method signatures throughout to add `odata.Query{}`
- Bump plugin SDK version to work with Terraform 1.0.2
- Implement request and response logging via middleware functions in `msgraph.Client{}`
- New feature: client certificate authentication by specifying an inline certificate


jackofallops (approved):
LGTM 🚀

Issue #369, opened 2020-12-02
data (all azuread): should be case insensitive


Tbohunek:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Much like #246 please make `data` lookups for all `azuread` resources case insensitive, meaning make `name`, `names`, `display_name`, `user_principal_name`, `user_principal_names` and such case-insensitive.

Keep `resource` case sensitive to avoid breaking change.

> **_Personal Note_** If accepted, this approach should be adopted for all future `data` resources on all `providers`.

### New or Affected Resource(s)

* azuread_user
* azuread_users
* azuread_groups
* azuread_application
* azuread_service_principal

### Potential Terraform Configuration

**Technically, no change is needed.** Since `data` lookup is case-sensitive, everyone already provides case-sensitive value, whose behavior won't change.

However, if someone needs to `data` a resource that has been made case-sensitive by someone else (like Azure AD Admin), he could dramatically simplify his code. Consider following setup with `env="dev"`, `project="jupyter"` and `azuread_application named Jupyter-Dev`:

This does not work:
```
data "azuread_application" "jupyter" {
  name = "${var.project}-${var.env}"
}
```
> **_Error_**: DisplayName for Applications matching "displayName eq 'jupyter-dev'" does not match("Jupyter-Dev"!="jupyter-dev")
> on dev/jupyter/main.tf line 37, in data "azuread_application" "jupyter":
  37: data "azuread_application" "jupyter" {

Workaround:
```
locals {
  env_case = {
    dev = "Dev"
    prod = "Prod"
  }
  project_case = {
    jupyter = "Jupyter"
    saturn = "Saturn"
    moon = "Moon"
  }
}
data "azuread_application" "jupyter" {
  name = "${local.project_case[var.project]}-${local.env_case[var.env]}"
}
```

### References

* #246

manicminer:
[...]

PR #492, opened 2021-07-21
data.azuread_application: match display_name case insensitively

manicminer:
- Also match some more UUIDs case insensitively as they should be

Fixes: #369


koikonom (approved):
LGTM

PR #495, opened 2021-07-22
Add state migration for azuread_application to handle public_client attribute name change

conzyor34:
- Migrate `public_client` attribute to `fallback_public_client_enabled` attribute


conzyor34:
Fix for #495

conzyor34:
Cancelling this PR because I messed this up with all the commits. Just learned about squash merges and will try again.

Issue #494, opened 2021-07-22
azuread_application needs schema migration for public_client attribute: Bool in v1.6.0 to List in v2.0.0


conzyor34:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

When upgrading provider version to v2.0.0, state created by v.1.6.0 fails to be read properly because the `public_client` attribute has changed types.

### Terraform (and AzureAD Provider) Version

```
$ terraform -v
Terraform v1.0.0
on linux_amd64
+ provider registry.localhost.com/local/azuread v2.0.0
+ provider registry.terraform.io/hashicorp/azuread v1.6.0
```

The v2.0 `azuread` provider was compiled from `main`.

### Affected Resource(s)

* `azuread_application`

#### Before

```hcl
terraform {
  required_providers {
    azuread = {
      source = "hashicorp/azuread"
      version = ">= 1.6.0"
    }     
  }
}

resource "azuread_application" "app_registration" {
  display_name = "test-app-registration"
  identifier_uris = ["https://website.company.com/"]
  sign_in_audience = "AzureADMyOrg"

  web {
    homepage_url = "https://website.company.com"
    redirect_uris = [
      "http://localhost/auth/redirect",
      "https://website.company.com/auth/redirect"
    ]

    implicit_grant {
      access_token_issuance_enabled = false
    }
  }
}
```

#### After Upgrade to v2.0

```hcl
terraform {
  required_providers {
    azuread = {
      source = "registry.localhost.com/local/azuread"
      version = ">= 2.0.0"
    }
  }
}

resource "azuread_application" "app_registration" {
  display_name = "test-app-registration"
  identifier_uris = ["https://website.company.com/"]
  sign_in_audience = "AzureADMyOrg"

  web {
    homepage_url = "https://website.company.com"
    redirect_uris = [
      "http://localhost/auth/redirect",
      "https://website.company.com/auth/redirect"
    ]

    implicit_grant {
      access_token_issuance_enabled = false
    }
  }
}
```

### Expected Behavior

Not an error.

### Actual Behavior

```
Error: missing expected [

  with azuread_application.app_registration,
[...]

PR #497, opened 2021-07-22
Add migration for public_client attribute

conzyor34:
- Add a migration for the `public_client` attribute to rename it to `fallback_public_client_enabled`


conzyor34:
This fixes #494

manicminer (approved):
Hi @conzyor34, many thanks for this. I amended it a little to only set `fallback_public_client_enabled` if it was not already present in state - for provider versions where this was introduced, it is computed and set to the same value as `public_client` (and vice versa). It will only need to be added to state where someone has upgraded from an earlier version directly to 2.0, else the value of `public_client` can be safely discarded.

Issue #498, opened 2021-07-22
azuread_application_password re-creates resource on provider v1.6.0 to v2.0.0 upgrade


conzyor34:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v1.0.0
on linux_amd64
+ provider registry.localhost.localdomain/localhost/azuread v2.0.0
+ provider registry.terraform.io/hashicorp/azuread v1.6.0
```

### Affected Resource(s)

* `azuread_application_password`

### Before (v1.6.0)

```hcl
terraform {
 required_providers {
   azuread = {
     source = "hashicorp/azuread"
     version = ">= 1.6.0"
   }     
 }
}

resource "azuread_application" "app_registration" {
  display_name = "test-app-registration"
  identifier_uris = ["https://website.company.com"]
  sign_in_audience = "AzureADMyOrg"
  
  api {
    oauth2_permission_scope {
      admin_consent_description  = "Allow the application to access test-app-registration on behalf of the signed-in user."
      admin_consent_display_name = "Access test-app-registration"
      enabled                    = true
      id                         = "fa449e9c-d495-45ec-914f-9d14b12b17f2"
      type                       = "User"
      user_consent_description   = "Allow the application to access test-app-registration on your behalf."
      user_consent_display_name  = "Access test-app-registration"
      value                      = "user_impersonation"
    }
  }

  web {
    homepage_url = "https://website.company.com"
    redirect_uris = [
      "http://localhost/auth/redirect",
      "https://website.company.com/auth/redirect"
    ]

    implicit_grant {
      access_token_issuance_enabled = false
    }
  }
}

resource "azuread_application_password" "app_password" {
  application_object_id = azuread_application.app_registration.id
  display_name = "TestCredential"
}
```

### After (v2.0.0 upgrade)

```hcl
terraform {
  required_providers {
    azuread = {
      source = "registry.localhost.localdomain/localhost/azuread"
      version = ">= 2.0.0"
    }
  }
}

<-- snip everything else the same -->
```

### Expected Behavior

`application_password` resource should not be replaced.

[...]

PR #499, opened 2021-07-23
When reading, use decoded CustomKeyIdentifier as display_name if DisplayName is nil

conzyor34:
Fix for #498 

- Decode CustomKeyIdentifier used by prior provider versions and use it as display_name if DisplayName is nil

### DIFF
--- a/internal/services/applications/application_password_resource.go
+++ b/internal/services/applications/application_password_resource.go
@@ -2,6 +2,7 @@ package applications
 
 import (
 	"context"
+	b64 "encoding/base64"
 	"errors"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/services/applications/application_password_resource.go
```diff
@@ -201,7 +202,17 @@ func applicationPasswordResourceRead(ctx context.Context, d *schema.ResourceData
 	}
 
 	tf.Set(d, "application_object_id", id.ObjectId)
-	tf.Set(d, "display_name", credential.DisplayName)
+
+	if credential.DisplayName != nil {
+		tf.Set(d, "display_name", credential.DisplayName)
+	} else {
```
manicminer, line 208: This will cause a crash if both `displayName` and `customKeyIdentifier` are null - which is the case for new password resources where `display_name` is omitted

```suggestion
	} else if credential.CustomKeyIdentifier != nil {
```

manicminer (approved):
Hi @conzyor34, thanks for this fix. I added a nil check for `CustomKeyIdentifier` since with new passwords created with MS Graph, both `DisplayName` and `CustomKeyIdentifier` can potentially be nil which causes a crash.

PR #500, opened 2021-07-23
azuread_service_principal_password: fix for unintentional forcenew

manicminer:
Source display_name from `CustomKeyIdentifier` field as per legacy API to prevent unintentional ForceNew after upgrade

(Same fix as implemented in #499 for `azuread_application_password`)


katbyte (approved):
LGTM 🤖

Issue #512, opened 2021-08-10
Feature Request: datasource all_users_datasource.go 


Threpio:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

For work I am currently treating our AAD setup as the single source of truth for the user base/permissions and group ownership. I would like to be able to fetch all of the users in the AAD to utilise the permissions already defined when I create other non AAD setup users (for example: Pagerduty). 

I would like for there to be a datasource list of users that can be iterated through to include their basic information such as userPrincipleName, email, phoneNumber and groups they are a part of.

This functionality of listing is available through the CLI and API. I would be able to use local_exec to obtain this information but this is not best practices and I would much prefer to allow terraform to handle this information for me. 

### New or Affected Resource(s)

To be created:
```
internal/services/users/all_users_datasource.go
internal/services/users/all_users_datasource_test.go
```

### Potential Terraform Configuration

```hcl
data "azuread_all_users" "users" {}
```

### References

* #0000

manicminer:
Hi @Threpio, thanks for requesting this!

Have you considered using the `azuread_groups` data source? It has a `members` attribute, so you could use a high level group containing all the users you wish to manage and discover their object IDs that way?

Threpio:
@manicminer - I have created a working branch on my fork and confirmed it works for my needs.

Thanks for your speedy comment though. I want to have a catch all as-well and want to move away from a global group in our AAD setup in the future. 

I just need to write some tests and I will create the PR.

Am I to create the documentation for this data source aswell? I know some other providers have automated documentation and don't want to duplicate my effort. (Happy to do it otherwise)

Threpio:
@manicminer - We did consider developing it as an extension as the azuread_users datasource but figured that the majority of the times that we use that datasource we DO NOT want a result that is not queried to be returned. So for us it means that if we mis-configure the datasource then it can look to be correct even when not filtered. (Happy to be overrule in this for the greater project)

 - It could be deigned as a `all_users` flag against the `azuread_users` datasource but I can't see how the `only-one-of` functionality for those inputs would work additionally with a boolean flag?

PR #513, opened 2021-08-11
add all_users_data_source.go and tests

Threpio:
For issue: https://github.com/hashicorp/terraform-provider-azuread/issues/512

This datasource is to be used to retrieve all users within the AAD (not restricted to groups) and so is a catchall. 

The datasource does not currently return the `mail_nicknames` like some of the other generalised data_sources.

The tests are minimal and I understand if other tests need to be put into place but I am not sure of the scenarios that could be explored. Without more input params or config I do not believe that there is more to tests.

Please let me know if anything needs to be changed/refactored as this is my first PR against this repo.


manicminer:
Hi @Threpio, thanks for this PR!

Quoting your comment on the issue (https://github.com/hashicorp/terraform-provider-azuread/issues/512#issuecomment-896540013):

> @manicminer - We did consider developing it as an extension as the azuread_users datasource but figured that the majority of the times that we use that datasource we DO NOT want a result that is not queried/filtered to be returned. So for us it means that if we mis-configure the datasource then it can look to be correct even when not filtered. (Happy to be overrule in this for the greater project)
> 
>     * It could be deigned as a `all_users` flag against the `azuread_users` datasource but I can't see how the `ExactlyOneOf` functionality for those inputs would work additionally with a boolean flag?

I do think this would be better implemented in the existing `azuread_users` data source, assuming no blockers arise from trying that. I agree it would be sensible to gate this with a boolean field. For the schema, I think we can add `ExactlyOneOf` to the boolean field as well and include it in the set of mutually exclusive fields? In the event that it is set to `false`, and the other fields are not specified, we'd likely have to check this in a CustomizeDiff function, e.g. `if show_all_users is false and mail_nicknames, object_ids, user_principal_names are all empty then raise an error`

Threpio:
@manicminer Thanks for the input. I will endeavour to implement it as that flag instead.

Would you like me to abandon this PR and branch or implement it here?

Threpio:
@manicminer - Changes have been pushed and I believe this works. I cannot work out how to test for it an error is thrown. Do you have any examples/documentation for what I could do to test that an error is thrown?

Threpio:
@manicminer - Sorry for the ping. I have attempted to fix the formatting and pass the linting stage but are still failing.

Any chance you could take a quick look and point out where I have mucked up? I have run the different fmt'ing commands with no changes. Feel free to push to the branch.

manicminer:
@Threpio No problem at all, it looks like the linter is complaining about line 222 of users_data_source_test.go - it looks like there's a tab indenting this line however Terraform configs should be indented by 2 spaces.

I'm going to queue this for v2.1.0 since we're currently working on some outstanding bugs and getting v2.0.0 ready for release. As it's possible the same files could change in the meantime, once 2.0.0 is out I'll circle back and give a proper review.

Threpio:
@manicminer I have fixed the merge conflicts - Apologies for the delay.

manicminer (approved):
Thanks @Threpio, this is looking good. I pushed some small changes:

- I renamed `show_all_users` to `return_all_users`
- We can use `ConflictsWith` between `ignore_missing` and `return_all_users`, and move the block that makes the API call to the if-else block, and then there's no need to do a ny apply-time validation :)

Many thanks for the contribution!

Test results:
Screenshot 2021-09-02 at 10 41 41

Issue #516, opened 2021-08-13
Feature Request: <show_all_groups = true> groups_datasource.go


Threpio:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Similar to #512 

For work I am treating the AAD setup as the single source of truth for user base/permissions and group permissions.

I would like to be able to fetch a list of all groups as well as fetch all groups on a wildcard filter (if possible).

The functionality of fetching all groups is similar to the all_users source where we are currently passing through a query set of params and so passing an empty set would return all groups. 

### New or Affected Resource(s)

AFFECTED - Changed to include a flag for 'all groups':

```
internal/services/groups/groups_datasource.go
internal/services/groups/groups_datasource_test.go
```

### Potential Terraform Configuration

```hcl
data "azuread_groups" "groups" {
  show_all_groups = true
}
```

### References

Sibling Issue: 
#512

PR #520, opened 2021-08-14
Feature/groups_datasource show_all_flag

Threpio:
For issue: #516 

This flag allows for terraform to interact with a list of currently existing groups within AAD for use in other environments (EG. Pagerduty).

It returns the same as other flags/inputs for the datasource and checks to ensure that a number greater than 0 are returned (otherwise an error is thrown in line with logic elsewhere)


internal/services/groups/groups_data_source.go
```diff
@@ -62,13 +69,27 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
...
 	if v, ok := d.GetOk("display_names"); ok {
 		displayNames = v.([]interface{})
 	}
 
-	if len(displayNames) > 0 {
+	if returnAll {
+		result, _, err := client.List(ctx, odata.Query{})
+		if err != nil {
+			return tf.ErrorDiagF(err, "Could not retrieve groups")
+		}
```
manicminer, line 83: We try to check errors separately so it's more obvious when they occur

internal/services/groups/groups_data_source.go
```diff
@@ -62,13 +69,27 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+		if result == nil {
+			return tf.ErrorDiagF(errors.New("API returned nil result"), "Bad API Response")
+		}
```
manicminer, line 86: We should check for nil pointers before dereferencing as doing `*result` when `result` is nil would cause a crash

internal/services/groups/groups_data_source.go
```diff
@@ -62,13 +69,27 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+		if len(*result) == 0 {
+			return tf.ErrorDiagPathF(err, "return_all", "No groups found")
+		}
+
+		groups = append(groups, *result...)
```
manicminer, line 91: This notation is a useful shortcut to iterating with a for loop

manicminer (approved):
Thanks again for this @Threpio! I pushed some changes but I noted them inline below in case it's useful to you.

Test results
Screenshot 2021-09-02 at 11 10 10

internal/services/groups/groups_data_source.go
Threpio: Didn't know this one - Thank you for that - I will read through all these comments :)

Issue #534, opened 2021-08-31
panic: interface conversion: interface {} is nil, not map[string]interface {}


dimbleby:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

terraform 1.0.2, azuread 2.0.1

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

I have a pretty large configuration that I probably mostly can't share, but if you have questions about specifics I'll do my best.

### Panic Output

https://gist.github.com/dimbleby/0d239ce9c5074abb537c9eed07961fe3

### Expected Behavior

No panic

### Actual Behavior

Panic

### Steps to Reproduce

Currently unclear, sorry.  I will try to figure out whether this reproduces reliably and, if so, what the difference is between working and not-working cases - and report back.

manicminer:
Hi @dimbleby, thanks for reporting this crash! I've opened a PR to mitigate this (and some other cases) with additional nil checking to prevent this.

Given your configuration, we could possibly narrow this down to the exact resource, but the linked mitigation should be enough to fix this and some other similar crash scenarios.

PR #536, opened 2021-08-31
Nil checking to prevent crashes in block expand/flatten handling for applications

manicminer:
Fixes: #534


jackofallops (approved):
LGTM 👍

Issue #537, opened 2021-08-31
Error: azuread_user + random_password


sergiomcalzada:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
terraform -v
Terraform v1.0.4 on linux_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.0.1
+ provider registry.terraform.io/hashicorp/random v3.1.0
```
### Affected Resource(s)

* `azuread_user`

### Terraform Configuration Files

```hcl
resource "random_password" "user" {
  length  = 32
  
  # lower   = true
  # upper   = true
  # number  = true
  # #special = true

  # min_lower   = 1
  # min_upper   = 1
  # min_numeric = 1
  # #min_special = 1  

  # #override_special = "_%@"
}

resource "azuread_user" "user1" {
  user_principal_name = "a@test.com"
  display_name        = "a@test.com"
  mail_nickname       = "a@test.com"

  password              = random_password.user.result
  force_password_change = true

  depends_on =[random_password.user]
}

resource "azuread_user" "user2" {
  user_principal_name = "a@test.com"
  display_name        = "a@test.com"
  mail_nickname       = "a@test.com"

  password              = "SecretP@sswd99!"
  force_password_change = true
}
```

### Debug Output

```
 Error: `password` is required when creating a new user

   with azuread_user.user1,
   on main.tf line 17, in resource "azuread_user" "user1":
   17: resource "azuread_user" "user1" {

```

### Expected Behavior

Resource `user1` is created without errors

### Actual Behavior

An error about empty password.

Resource `user2 `with hardcoded password works properly

### Steps to Reproduce

1. `terraform plan` or `terraform apply`

manicminer:
Hi @sergiomcalzada, thanks for reporting this. I've identified the source of this bug and will try to issue a fix asap. Thanks!

PR #543, opened 2021-09-01
Bugfix: password validation at apply-time …

manicminer:
… because it's unknown at plan time when interpolated

Closes: #537


koikonom (approved):
LGTM

Issue #136, opened 2019-08-15
azuread_user: Support for "PasswordNeverExpire" option


lzadjsf:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Times to times there is a need to set password never expire for Azure AD user account as it is a service account. It need to applications that still doesn't support latest authentication and authorisation mechanisms. Say, services which requires for LDAP bindings through Azure Active Directory Domain Services.

It can be easily done by PowerShell command:
```
Set-MsolUser -UserPrincipalName <user ID> -PasswordNeverExpires $true
```
By the way, there is a need to perform that kind of task through TF scripts during automatic deployment of service.

### New or Affected Resource(s)

* azuread_user

### Potential Terraform Configuration
```hcl
resource "azuread_user" "example" {
  user_principal_name = "jdo@hashicorp.com"
  display_name        = "J. Doe"
  mail_nickname       = "jdoe"
  password            = "SecretP@sswd99!"
  password_expiration  ="True|False"
}
```

### References

* https://docs.microsoft.com/en-GB/office365/admin/add-users/set-password-to-never-expire?view=o365-worldwide
* https://docs.microsoft.com/en-us/azure/active-directory/authentication/concept-sspr-policy

lzadjsf:
Hello,

Any news?

katbyte:
Hi @lzadjsf,

Thank you for opening this issue so we can track this feature. As that property is currently not exposed but the SDK this is not on our current roadmap to implement anytime soon.

katbyte:
This might be possible via the additional properties map.. 🤔

dbourcet-wiz:
@katbyte What do you mean by the additionnal properties map? It does interest me.

manicminer:
This is supported by the Azure AD Graph API. I have logged an issue requesting SDK support.

https://github.com/Azure/azure-sdk-for-go/issues/9851

alexwilcox9:
[...]

PR #550, opened 2021-09-03
Support for password policies

alexwilcox9:
Closes: https://github.com/hashicorp/terraform-provider-azuread/issues/136

@manicminer is this the kind of thing you had in mind? 
One thing I haven't checked is if both are true will graph always return them in that order or should we just check if the string contains `DisablePasswordExpiration`/`DisableStrongPassword` rather than going for an exact match

### DIFF
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -37,6 +37,8 @@ The following arguments are supported:
 * `consent_provided_for_minor` - (Optional) Whether consent has been obtained for minors. Supported values are `Granted`, `Denied` and `NotRequired`. Omit this property or specify a blank string to unset.
 * `country` - (Optional) The country/region in which the user is located, e.g. `US` or `UK`.
 * `department` - (Optional) The name for the department in which the user works.
+* `disable_password_expiration` - (Optional) Whether the users password is exempt from expiring. Defaults to `false`.
+* `disable_strong_password` - (Optional) Whether the user is allowed weaker passwords than the default policy to be specified. Defaults to `false`.
 * `display_name` - (Required) The name to display in the address book for the user.
 * `employee_id` - (Optional) The employee identifier assigned to the user by the organisation.
 * `fax_number` - (Optional) The fax number of the user.
--- a/internal/services/users/user_resource.go
+++ b/internal/services/users/user_resource.go
@@ -380,16 +380,16 @@ func userResourceCreate(ctx context.Context, d *schema.ResourceData, meta interf
 		mailNickName = strings.Split(upn, "@")[0]
 	}
 
-	passwordPolicies := utils.NullableString("")
-	disable_strong_password := d.Get("disable_strong_password").(bool)
-	disable_password_expiration := d.Get("disable_password_expiration").(bool)
+	var passwordPolicies string
+	disableStrongPassword := d.Get("disable_strong_password").(bool)
+	disablePasswordExpiration := d.Get("disable_password_expiration").(bool)
 
-	if disable_strong_password && (!disable_password_expiration) {
-		passwordPolicies = utils.NullableString("DisableStrongPassword")
-	} else if (!disable_strong_password) && disable_password_expiration {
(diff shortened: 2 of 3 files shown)


manicminer:
@alexwilcox9 This looks good! I think we could probably make them top-level properties like `force_password_change`, there isn't much to gain by putting them in a block?

alexwilcox9:
@manicminer I just love blocks! That's fair enough I've rearranged it a bit now, also not sure what to call the two functions as they're not really expand/flatten functions so suggestions are welcome!

manicminer:
I like blocks too, it's just that they introduce complexity when interacting with the TF plugin SDK :)

To be honest I don't think we need flatten/expand functions in this case, we can do it inline in the create/update funcs without much boilerplate. The only thing I would change with the parsing of the API response for `passwordPolicies` is I would split it into a slice on the comma, and do a case insensitive (space trimmed) comparison with each value, to account for the API reordering or changing them.

alexwilcox9:
Agreed, splitting it into a slice makes more sense. I've scrapped the functions and moved the contents into the create/update/read functions

alexwilcox9:
I've hit a new issue now that I'm hoping you can help advise on.

Currently I set `passwordPolicies` to a pointer to an empty string if neither options are specified however that lead to me getting the following
```
        UsersClient.BaseClient.Post(): unexpected status 400 with OData error:
        Request_BadRequest: Invalid value specified for property 'passwordPolicies'
        of resource 'User'.
```

So I then had the idea to set `passwordPolicies` to nil if neither option were selected which did initially solved the problem but seems to have broken the update function should you go from having either option set to neither of them as sending nil doesn't update the password policies
```
    testcase.go:60: Step 5/6 error: After applying this test step, the plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # azuread_user.test will be updated in-place
          ~ resource "azuread_user" "test" {
              ~ disable_password_expiration = true -> false
              ~ disable_strong_password     = true -> false
                id                          = "b6df9c72-649e-41d3-91fb-fec9461d596e"
                # (16 unchanged attributes hidden)
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
--- FAIL: TestAccUser_update (30.32s)
```

So I can't work out what value you set if you want to turn both options off..

I hope that makes sense! Any ideas?

manicminer:
Hi @alexwilcox9, this is a fairly common issue which can be resolved, but requires a change in the SDK (hamilton). For values that should be unset by sending `null` instead of a blank string, there is a custom type `StringNullWhenEmpty` which should be used instead of a regular `string`.

I've tried to catch as many of these as I can based on documentation, but it doesn't always indicate in the docs. You can fix it in Hamilton and it should solve this.

The line to fix is https://github.com/manicminer/hamilton/blob/main/msgraph/models.go#L1126

It is currently:
```go
	PasswordPolicies                *string                  `json:"passwordPolicies,omitempty"`
```

but it should be:
```go
	PasswordPolicies                *StringNullWhenEmpty     `json:"passwordPolicies,omitempty"`

```

Give me a shout on Slack if I can assist with making this change in your dev environment. I often add a [go.mod override](https://golang.org/ref/mod#go-mod-file-replace) and then vendor from a local copy, and then PR the change in Hamilton once i'm confident it works.

alexwilcox9:
Thanks for the tip and help getting my environment set up, I've created a PR in the SDK repo for this change
https://github.com/manicminer/hamilton/pull/96

alexwilcox9:
Just thinking... whilst I'm at it should I update the user data resource to include these new attributes?

manicminer (changes requested):
This looks and works great, just some minor suggestions on variable naming but otherwise this LGTM 👍

manicminer (approved):
LGTM 🚀 

Screenshot 2021-09-09 at 23 39 09

Issue #551, opened 2021-09-03
Filter for azuread_groups security_enabled


timja:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azuread_groups

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.

data "azuread_groups" "groups" {
  display_names = ["group-a", "group-b"]
  security_enabled = true
}
```

Without this I get:
```
More than one group found with display name: group-name
```

because we have a security group and a mail group with the same name.

This has happened since the azuread provider was released in 2.0

### References

* #0000

PR #552, opened 2021-09-05
Feature/issue 551 azuread_groups security_enabled flag

Threpio:
Fixes [#551](https://github.com/hashicorp/terraform-provider-azuread/issues/551)

Added a `security_enabled=true` flag to `azuread_groups` datasource. It currently does not implement any logic when `security_enabled=false` which some people might want in the future.

This checks to ensure that atleast 1 result is returned - If one non security_enabled group is found but then dropped then an error suggesting the flag will be returned.

The test functionality for this and the return_all flag needs to be revisited by someone that understands this testing framwork more than me. There are single cases

### DIFF
--- a/docs/data-sources/groups.md
+++ b/docs/data-sources/groups.md
@@ -30,6 +30,15 @@ data "azuread_groups" "allGroups" {
 }
 ```
 
+*Look up all security groups*
+```terraform
+data "azuread_groups" "allGroups" {
+  return_all       = true
+  security_enabled = true
+}
+```
+
+
 ## Argument Reference
 
 The following arguments are supported:
@@ -38,8 +47,11 @@ The following arguments are supported:
 * `object_ids` - (Optional) The object IDs of the groups.
 * `return_all` - (Optional) A flag to denote if all groups should be fetched and returned.
 * `security_enabled` - (Optional) A flag to denote if only `security_enabled=true` groups should be returned.
+* `mail_enabled` - (Optional) A flag to denote if only `mail_enabled=true` groups should be returned.
 
 ~> One of `display_names`, `object_ids` or `return_all` should be specified. Either of the first two _may_ be specified as an empty list, in which case no results will be returned.
+
+~> `security_enabled` and `mail_enabled` flags work with `return_all` and `display_names` but not `object_ids`
 
 ## Attributes Reference
 
--- a/docs/data-sources/groups.md
+++ b/docs/data-sources/groups.md
@@ -37,6 +37,7 @@ The following arguments are supported:
 * `display_names` - (Optional) The display names of the groups.
 * `object_ids` - (Optional) The object IDs of the groups.
 * `return_all` - (Optional) A flag to denote if all groups should be fetched and returned.
+* `security_enabled` - (Optional) A flag to denote if only `security_enabled=true` groups should be returned.
 
 ~> One of `display_names`, `object_ids` or `return_all` should be specified. Either of the first two _may_ be specified as an empty list, in which case no results will be returned.
 
--- a/docs/data-sources/groups.md
+++ b/docs/data-sources/groups.md
@@ -30,7 +30,7 @@ data "azuread_groups" "allGroups" {
 }
 ```
 
-*Look up all security groups*
+*Look up all security-enabled groups*
 ```terraform
 data "azuread_groups" "allGroups" {
   return_all       = true
--- a/docs/data-sources/groups.md
+++ b/docs/data-sources/groups.md
@@ -51,7 +51,7 @@ The following arguments are supported:
 
 ~> One of `display_names`, `object_ids` or `return_all` should be specified. Either of the first two _may_ be specified as an empty list, in which case no results will be returned.
 
-~> `security_enabled` and `mail_enabled` flags work with `return_all` and `display_names` but not `object_ids`
+-> `security_enabled` and `mail_enabled` flags work with `return_all` and `display_names` but not `object_ids`
 
 ## Attributes Reference
 
--- a/docs/data-sources/groups.md
+++ b/docs/data-sources/groups.md
@@ -44,10 +44,10 @@ data "azuread_groups" "allGroups" {
 The following arguments are supported:
 
 * `display_names` - (Optional) The display names of the groups.
+* `mail_enabled` - (Optional) A flag to denote if only `mail_enabled=true` groups should be returned.
 * `object_ids` - (Optional) The object IDs of the groups.
 * `return_all` - (Optional) A flag to denote if all groups should be fetched and returned.
 * `security_enabled` - (Optional) A flag to denote if only `security_enabled=true` groups should be returned.
-* `mail_enabled` - (Optional) A flag to denote if only `mail_enabled=true` groups should be returned.
 
 ~> One of `display_names`, `object_ids` or `return_all` should be specified. Either of the first two _may_ be specified as an empty list, in which case no results will be returned.
 
--- a/internal/services/groups/groups_data_source.go
+++ b/internal/services/groups/groups_data_source.go
@@ -161,6 +161,10 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 			newDisplayNames = append(newDisplayNames, *group.DisplayName)
 		}
 	}
+	// Check if securityEnabled has caused the number of returned groups to be 0
+	if len(newObjectIds) == 0 && securityEnabled{
+		return tf.ErrorDiagF(errors.New("No groups found with 'security_enabled = true'"), "Unexpected Number of groups returned")
+	}
 
 	h := sha1.New()
 	if _, err := h.Write([]byte(strings.Join(newDisplayNames, "-"))); err != nil {
--- a/internal/services/groups/groups_data_source.go
+++ b/internal/services/groups/groups_data_source.go
@@ -112,10 +112,8 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 
 			if securityEnabled {
 				groups = append(groups, *result...)
-
 			} else if count > 1 {
 				return tf.ErrorDiagPathF(err, "display_names", "More than one group found with display name: %q", displayName)
-
 			} else if count == 0 {
 				return tf.ErrorDiagPathF(err, "display_names", "No group found with display name: %q", displayName)
 			}
--- a/internal/services/groups/groups_data_source.go
+++ b/internal/services/groups/groups_data_source.go
@@ -103,27 +103,7 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 		if len(*result) == 0 {
 			return tf.ErrorDiagPathF(err, "return_all", "No groups found")
 		}
-		if securityEnabled && !mailEnabled {
-			for _, r := range *result {
-				if *r.SecurityEnabled {
-					groups = append(groups, r)
-				}
-			}
-		} else if mailEnabled && !securityEnabled {
-			for _, r := range *result {
-				if *r.MailEnabled {
-					groups = append(groups, r)
-				}
-			}
-		} else if mailEnabled && securityEnabled {
-			for _, r := range *result {
-				if *r.MailEnabled && *r.SecurityEnabled {
-					groups = append(groups, r)
-				}
-			}
-		} else {
-			groups = append(groups, (*result)[0])
-		}
+		groups = filterResults(securityEnabled, mailEnabled, result)
 
 	} else if len(displayNames) > 0 {
 		expectedCount = len(displayNames)
@@ -144,26 +124,8 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 				return tf.ErrorDiagPathF(err, "display_names", "More than one group found with display name: %q", displayName)
 			} else if count == 0 {
 				return tf.ErrorDiagPathF(err, "display_names", "No group found with display name: %q", displayName)
-			} else if securityEnabled && !mailEnabled {
-				for _, r := range *result {
-					if *r.SecurityEnabled {
-						groups = append(groups, r)
-					}
-				}
-			} else if mailEnabled && !securityEnabled {
-				for _, r := range *result {
-					if *r.MailEnabled {
-						groups = append(groups, r)
-					}
-				}
-			} else if mailEnabled && securityEnabled {
-				for _, r := range *result {
-					if *r.MailEnabled && *r.SecurityEnabled {
-						groups = append(groups, r)
-					}
-				}
 			} else {
-				groups = append(groups, (*result)[0])
+				groups = filterResults(securityEnabled, mailEnabled, result)
 			}
 		}
 	} else if objectIds, ok := d.Get("object_ids").([]interface{}); ok && len(objectIds) > 0 {
@@ -182,7 +144,7 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 		}
 	}
 
-	if !returnAll && !securityEnabled && len(groups) != expectedCount {
+	if !returnAll && !securityEnabled && !mailEnabled && len(groups) != expectedCount {
 		return tf.ErrorDiagF(fmt.Errorf("Expected: %d, Actual: %d", expectedCount, len(groups)), "Unexpected number of groups returned")
 	}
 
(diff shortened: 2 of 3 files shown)


docs/data-sources/groups.md
```diff
@@ -37,6 +37,7 @@ The following arguments are supported:
 * `display_names` - (Optional) The display names of the groups.
 * `object_ids` - (Optional) The object IDs of the groups.
 * `return_all` - (Optional) A flag to denote if all groups should be fetched and returned.
+* `security_enabled` - (Optional) A flag to denote if only `security_enabled=true` groups should be returned.
```
manicminer, line 40:
```suggestion
* `security_enabled` - (Optional) Whether the groups are security-enabled.
```

internal/services/groups/groups_data_source.go
```diff
@@ -59,6 +59,12 @@ func groupsDataSource() *schema.Resource {
 				Optional:     true,
 				ExactlyOneOf: []string{"display_names", "object_ids", "return_all"},
 			},
+
+			"security_enabled": {
+				Description: "Retrieve only groups that are security_enabled groups",
```
manicminer, line 64:
```suggestion
				Description: "Whether the groups are security-enabled",
```

internal/services/groups/groups_data_source.go
```diff
@@ -59,6 +59,12 @@ func groupsDataSource() *schema.Resource {
...
+				Type:        schema.TypeBool,
+				Optional:    true,
```
manicminer, line 66: I think this should be mutually exclusive with the `return_all` property?

```suggestion
				Optional:    true,
				ConflictsWith: []string{"return_all"},
```

internal/services/groups/groups_data_source.go
```diff
@@ -59,6 +59,12 @@ func groupsDataSource() *schema.Resource {
 				Optional:     true,
 				ExactlyOneOf: []string{"display_names", "object_ids", "return_all"},
```
manicminer, line 60:
```suggestion
				ConflictsWith: []string{"security_enabled"},
				ExactlyOneOf: []string{"display_names", "object_ids", "return_all"},
```

internal/services/groups/groups_data_source.go
```diff
@@ -102,7 +109,10 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 			}
 
 			count := len(*result)
-			if count > 1 {
+
+			if securityEnabled {
+				groups = append(groups, *result...)
+			} else if count > 1 {
```
manicminer, line 115: This looks like it will just append the results without checking the security-enabled status of the discovered group?

internal/services/groups/groups_data_source_test.go
```diff
@@ -121,3 +135,12 @@ data "azuread_groups" "test" {
 }
 `
 }
+
+func (GroupsDataSource) securityEnabled() string {
+	return `
+data "azuread_groups" "test" {
+  return_all       = true
+  security_enabled = true
```
manicminer, line 143: These should be mutually exclusive since `return_all` implies all groups? Alternatively we could update the docs to reflect that `return_all` excludes `display_names` and `object_ids` but can be used with `security_enabled` to return all security-enabled groups.

internal/services/groups/groups_data_source.go
```diff
@@ -59,6 +59,12 @@ func groupsDataSource() *schema.Resource {
 				Optional:     true,
 				ExactlyOneOf: []string{"display_names", "object_ids", "return_all"},
 			},
+
+			"security_enabled": {
+				Description: "Retrieve only groups that are security_enabled groups",
+				Type:        schema.TypeBool,
+				Optional:    true,
+			},
```
manicminer, line 67: This should also be computed since it will otherwise default to false, which would imply only non-security-enabled groups are returned by default.

```suggestion
				Computed: true,
			},
```

manicminer (changes requested):
Hi @Threpio, thanks for this PR, this is off to a great start!

If we're adding a `security_enabled` flag, could we also add a `mail_enabled` flag? M365 groups can be secuyrity-enabled so it's probably desired to be able to specify both `security_enabled = true` and `mail_enabled = false`.

For the implementation, I think it would be better to first build up a filter based on `security_enabled` (and `mail_enabled`) and include that in the main if-else block. We'd also need to explicitly check the results since the filter is useful for server-side filtering but we also need client-side checks.

internal/services/groups/groups_data_source.go
Threpio: I think it should be possible to query for all groups that are security_enabled ?
 
I know it is a complication with regards to logic but the ability to just filter out 365 groups could be a useful tool.

Threpio:
Would you like me to do the mail_enabled flag logic in this PR aswell or to have it's own?

docs/data-sources/groups.md
Threpio: I like the previous wording - It matches the boolean flag wording of `return_all` - But I understand if the english is a bit crazy.

manicminer: That's fine, I was thinking to align the wording more with the `azuread_group` DS but this is sufficiently different 👍

internal/services/groups/groups_data_source.go
manicminer: Yeah the more I've thought about it, I agree it should be possible to specify `return_all` and filter out mail and security groups.

manicminer:
> Would you like me to do the mail_enabled flag logic in this PR aswell or to have it's own?

In this PR would be great since they are very related - and noting that M365 groups can be security enabled, having the mail_enabled property will be important.

It would be good to also add the `types` property to this and the singular azuread_group data sources (with similar logic to how the `supports_services` field works in the azuread_domains DS), but that can wait for another PR.

> Also I have noted that I have missed the documentation for logic of the conflict between specified object_ids and this flag - I am right in thinking they would conflict right?

Correct, `object_ids` would be exclusive to security_enabled/mail_enabled, but `display_names` will be complementary/compatible.

Threpio:
Just to check - is mail_enabled and security_enabled conflict? Can a group retain both of these flags to true

manicminer:
> Just to check - is mail_enabled and security_enabled conflict? Can a group retain both of these flags to true

A group (depending on its type) can be either mail_enabled / security_enabled or both.

internal/services/groups/groups_data_source.go
Threpio: This has been changed and the logic moved into the results fetching

internal/services/groups/groups_data_source_test.go
Threpio: I think I agree with the alternative. Otherwise we can look to apply a number of filters for other fields at the same time.

internal/services/groups/groups_data_source.go
Threpio: Filter-out or filter in?

internal/services/groups/groups_data_source.go
```diff
@@ -87,8 +105,7 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 		if len(*result) == 0 {
 			return tf.ErrorDiagPathF(err, "return_all", "No groups found")
 		}
-
-		groups = append(groups, *result...)
+		groups = filterResults(securityEnabled, mailEnabled, result)
```
manicminer, line 108: Could we compose odata filters for this too, to minimise the number of unrelated results going over the wire? e.g. `securityEnabled eq true` and/or `mailEnabled eq true`

internal/services/groups/groups_data_source.go
```diff
@@ -102,13 +119,15 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 			}
 
 			count := len(*result)
-			if count > 1 {
+
+			//Could we make this a switch case?
```
manicminer, line 123: If think this is fine as an if-else block for now. Maybe one more case would make it eligible IMO

internal/services/groups/groups_data_source.go
```diff
@@ -102,13 +119,15 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+			if !mailEnabled && !securityEnabled && count > 1 {
```
manicminer, line 124: Assuming we use odata filters, I think this would be better left as `if count > 1`?

internal/services/groups/groups_data_source.go
```diff
@@ -126,7 +145,7 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 		}
 	}
 
-	if !returnAll && len(groups) != expectedCount {
+	if !returnAll && !securityEnabled && !mailEnabled && len(groups) != expectedCount {
```
manicminer, line 148: As above, with server-side filtering, this can be left as-is

internal/services/groups/groups_data_source.go
```diff
@@ -139,10 +158,13 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
...
 			return tf.ErrorDiagF(errors.New("API returned group with nil displayName"), "Bad API response")
 		}
-
 		newObjectIds = append(newObjectIds, *group.ID)
 		newDisplayNames = append(newDisplayNames, *group.DisplayName)
 	}
+	// Check if securityEnabled/mailEnabled has caused the number of returned groups to be 0
+	if len(newObjectIds) == 0 && (securityEnabled || mailEnabled) {
+		return tf.ErrorDiagF(errors.New("No groups found with a filter flag applied"), "Unexpected Number of groups returned")
+	}
```
manicminer, line 167: Not sure what scenario this is intended to catch as this would already have errored in the main if-else block above?

internal/services/groups/groups_data_source.go
```diff
@@ -156,3 +178,29 @@ func groupsDataSourceRead(ctx context.Context, d *schema.ResourceData, meta inte
 
 	return nil
 }
+
+// Extracting the flag functionality to prevent duplication
+func filterResults(securityEnabled, mailEnabled bool, results *[]msgraph.Group) (groups []msgraph.Group) {
+	if securityEnabled && !mailEnabled {
```
manicminer, line 184: We'll need a nil check for `results` here or we'll get a crash when dereferencing

internal/services/groups/groups_data_source_test.go
manicminer: Sounds good to me!

internal/services/groups/groups_data_source_test.go
```diff
@@ -70,6 +70,34 @@ func TestAccGroupsDataSource_returnAll(t *testing.T) {
...
+	data.DataSourceTest(t, []resource.TestStep{
+		{
+			Config: GroupsDataSource{}.mailEnabled(),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).Key("display_names.#").Exists(),
+				check.That(data.ResourceName).Key("object_ids.#").Exists(),
+			),
+		},
+	})
+}
```
manicminer, line 99: These are good, can we also add a test for security_enabled + mail_enabled to make sure these work together as expected?

internal/services/groups/groups_data_source_test.go
```diff
@@ -121,3 +149,21 @@ data "azuread_groups" "test" {
...
+  security_enabled = true
+}
+`
+}
+
+func (GroupsDataSource) mailEnabled() string {
+	return `
+data "azuread_groups" "test" {
+  return_all   = true
+  main_enabled = true
```
manicminer, line 166: typo?

```suggestion
  mail_enabled = true
```

docs/data-sources/groups.md
```diff
@@ -30,16 +30,29 @@ data "azuread_groups" "allGroups" {
 }
 ```
 
+*Look up all security groups*
```
manicminer, line 33: I think it's worth this subtle distinction due to the aforementioned possibility that M365 groups can act as security groups

```suggestion
*Look up all security-enabled groups*
```

docs/data-sources/groups.md
```diff
@@ -30,16 +30,29 @@ data "azuread_groups" "allGroups" {
...
+
 ## Argument Reference
 
 The following arguments are supported:
 
 * `display_names` - (Optional) The display names of the groups.
 * `object_ids` - (Optional) The object IDs of the groups.
 * `return_all` - (Optional) A flag to denote if all groups should be fetched and returned.
+* `security_enabled` - (Optional) A flag to denote if only `security_enabled=true` groups should be returned.
+* `mail_enabled` - (Optional) A flag to denote if only `mail_enabled=true` groups should be returned.
```
manicminer, line 50: Could we move this up a few lines to maintain alphabetic sorting?

docs/data-sources/groups.md
```diff
@@ -30,16 +30,29 @@ data "azuread_groups" "allGroups" {
...
 
 ~> One of `display_names`, `object_ids` or `return_all` should be specified. Either of the first two _may_ be specified as an empty list, in which case no results will be returned.
 
+~> `security_enabled` and `mail_enabled` flags work with `return_all` and `display_names` but not `object_ids`
```
manicminer, line 54: This should be an info box (`~>` makes a red box)

```suggestion
-> `security_enabled` and `mail_enabled` flags work with `return_all` and `display_names` but not `object_ids`
```

manicminer (changes requested):
@Threpio Thanks for the updates, this is looking mostly good. Just a few comments on the implementation, mainly the use of odata filters to get server-side filtering (we'll still keep the client-side filtering too). If you can take a look at these, this should be good to merge!

docs/data-sources/groups.md
Threpio: I had no idea that there were info boxes - Apologies

internal/services/groups/groups_data_source.go
Threpio: Yes - Using server side filtering would allow that to return correctly.

If we search for a group with displayname xyz-2 but there is a group xyz-23 does it return both groups or just the one with the exact matching display name? (I assume this is Hamilton and their server doing the clever querying.

internal/services/groups/groups_data_source.go
manicminer: Sorry filter in, as in I should have written "filter for" :D

internal/services/groups/groups_data_source.go
Threpio: I may be being stupid and not understanding odata querying - But would this not lead to something like:

`			if securityEnabled && mailEnabled{
				query := odata.Query{
					Filter: fmt.Sprintf("displayName eq '%s', securityEnabled eq true, mailEnabled eq true", displayName),
				}
			} else if securityEnabled {
				query := odata.Query{
					Filter: fmt.Sprintf("displayName eq '%s', securityEnabled eq true", displayName),
				}
			} else if mailEnabled {
				query := odata.Query{
					Filter: fmt.Sprintf("displayName eq '%s', mailEnabled eq true", displayName),
				}
			} else {
				query := odata.Query{
					Filter: fmt.Sprintf("displayName eq '%s'", displayName),
				}
			}`

internal/services/groups/groups_data_source.go
manicminer: There are several operators and functions. `eq` is a whole value check. there's also a `startsWith` function which does what you'd expect ([docs](https://docs.microsoft.com/en-us/graph/query-parameters#filter-parameter))

internal/services/groups/groups_data_source.go
Threpio: I suppose what we should think about aswell is the oppertunity for future work - Perhaps people will want to specify `mail_enabled=false` and only get non-mail_enabled groups

manicminer: Right, that's sort of what I was trying to get at with the inclusion of both properties. Someone may want to specify `security_enabled=true` and `mail_enabled=false` to get security groups that aren't mail enabled (e.g. effectively only get classic security groups and not potential M365 groups)

manicminer: (You should be able to use d.GetOk() to test if the property was specified versus just having it's default (unset) value)

Threpio:
Sorry @manicminer - My priorities at work have been changed and I will no longer be able to finish up this PR. Do you want me to close it or?

manicminer:
@Threpio No problem, thanks for your work on it and for letting us know. Go ahead and close it, I'm sure it will be picked up again at some point :)

PR #556, opened 2021-09-07
[docs] Fix typo in Service Principal/Client Secret doc

nexxai:
The page references using a "Client Certificate" but it should read "Client Secret"


manicminer (approved):
Hi @nexxai, good spot and thanks for the fix!

Issue #568, opened 2021-09-11
Conditional Access - Bad Request when only specifying sign in frequency options in session controls block


alexwilcox9:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
```
terraform_version | Terraform v1.0.6 on linux_amd64 + provider registry.terraform.io/hashicorp/azuread v2.2.1
```

### Affected Resource(s)

* `azuread_conditional_access_policy`

### Terraform Configuration Files

```hcl
resource "azuread_conditional_access_policy" "test" {
  display_name = "acctest-CONPOLICY-1"
  state        = "disabled"

  conditions {
    applications {
      included_applications = ["All"]
      excluded_applications = ["00000004-0000-0ff1-ce00-000000000000"]
    }
    users {
      included_users = ["All"]
      excluded_users = ["GuestsOrExternalUsers"]
    }
    client_app_types = ["all"]
    locations {
      included_locations = ["All"]
      excluded_locations = ["AllTrusted"]
    }
    platforms {
      included_platforms = ["android"]
      excluded_platforms = ["iOS"]
    }
    sign_in_risk_levels = ["medium"]
    user_risk_levels    = ["medium"]
  }

  grant_controls {
    operator          = "OR"
    built_in_controls = ["mfa"]
  }

  session_controls {
    sign_in_frequency                         = 10
    sign_in_frequency_period                  = "hours"
  }
}
```

### Output
```
        Error: Could not create conditional access policy
        
          with azuread_conditional_access_policy.test,
          on terraform_plugin_test.tf line 2, in resource "azuread_conditional_access_policy" "test":
           2: resource "azuread_conditional_access_policy" "test" {
        
        ConditionalAccessPolicyClient.BaseClient.Post(): unexpected status 400 with
        OData error: BadRequest: The server could not process the request because it
        is malformed or incorrect.: 1007: Incoming ConditionalAccessPolicy object is
        null or does not match the schema of ConditionalAccessPolicy type. For
        examples, please see API documentation at
[...]

PR #569, opened 2021-09-11
Bugfix CA Policies

alexwilcox9:
Fixes https://github.com/hashicorp/terraform-provider-azuread/issues/568

Whilst testing this I initially started by commenting out the setting of `ApplicationEnforcedRestrictions` and `CloudAppSecurity` in `expandConditionalAccessSessionControls` which bizarrely started causing panics in `flattenConditionalAccessSessionControls`

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xed80d0]

goroutine 691 [running]:
github.com/hashicorp/terraform-provider-azuread/internal/services/conditionalaccess.flattenConditionalAccessSessionControls(...)
        /home/alexwilcox/go/src/github.com/hashicorp/terraform-provider-azuread/internal/services/conditionalaccess/conditional_access_policy_resource.go:573
github.com/hashicorp/terraform-provider-azuread/internal/services/conditionalaccess.conditionalAccessPolicyResourceRead(0x13afeb8, 0xc000d0acc0, 0xc0009aa400, 0x10322a0, 0xc00089a1c0, 0xc000992d70, 0xc00075dae0, 0x0)
        /home/alexwilcox/go/src/github.com/hashicorp/terraform-provider-azuread/internal/services/conditionalaccess/conditional_access_policy_resource.go:425 +0x6d0
github.com/hashicorp/terraform-provider-azuread/internal/services/conditionalaccess.conditionalAccessPolicyResourceCreate(0x13afeb8, 0xc000d0acc0, 0xc0009aa400, 0x10322a0, 0xc00089a1c0, 0xc000992d60, 0x9059ea, 0xc00075cb60)
```

I believe I've solved this by adding some nil checks in that function

I found that not setting `ApplicationEnforcedRestrictions` and `CloudAppSecurity` at all in `expandConditionalAccessSessionControls` even if they weren't set in the terraform configuration got it working so I've now rearranged the logic in that function so that if the attributes aren't set in the terraform we don't instantiate those structs

I've tried updating the `session_controls` block in the complete test to the following and all are now passing

```
  session_controls {
    application_enforced_restrictions_enabled = true
    cloud_app_security_policy                 = "monitorOnly"
    sign_in_frequency                         = 10
    sign_in_frequency_period                  = "hours"
  }
```
```
  session_controls {
    cloud_app_security_policy                 = "monitorOnly"
    sign_in_frequency                         = 10
    sign_in_frequency_period                  = "hours"
  }
```
```
  session_controls {
    sign_in_frequency                         = 10
    sign_in_frequency_period                  = "hours"
  }
```
```
  session_controls {
    application_enforced_restrictions_enabled = true
    cloud_app_security_policy                 = "monitorOnly"
  }
```

Also when running all these tests I did hit the update issue that named locations were getting (and were solved by the refresh func) but I wasn't sure how best to go about adding this as I see in named locations you just compare each value in turn and there's a lot of values to check in this resource! If that is the way I'm happy to crack on, just wanted to check first

```
=== CONT  TestAccConditionalAccessPolicy_update
    testcase.go:60: Step 3/4 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # azuread_conditional_access_policy.test will be updated in-place
          ~ resource "azuread_conditional_access_policy" "test" {
                id           = "84812c78-43fe-4338-b738-9c037623f6d3"
                # (2 unchanged attributes hidden)
        
              ~ conditions {
                  ~ client_app_types    = [
                      - "browser",
                      + "all",
                    ]
                  ~ sign_in_risk_levels = [
                      + "medium",
                    ]
                  ~ user_risk_levels    = [
                      + "medium",
                    ]
        
                  ~ applications {
                      ~ excluded_applications = [
                          + "00000004-0000-0ff1-ce00-000000000000",
                        ]
                        # (2 unchanged attributes hidden)
                    }
        
                  ~ locations {
                      ~ excluded_locations = [
                          + "AllTrusted",
                        ]
                        # (1 unchanged attribute hidden)
                    }
        
                  + platforms {
                      + excluded_platforms = [
                          + "iOS",
                        ]
                      + included_platforms = [
                          + "android",
                        ]
                    }
        
                    # (1 unchanged block hidden)
                }
        
              ~ grant_controls {
                  ~ built_in_controls             = [
                      - "block",
                      + "mfa",
                    ]
                    # (3 unchanged attributes hidden)
                }
        
              + session_controls {
                  + application_enforced_restrictions_enabled = true
                  + cloud_app_security_policy                 = "monitorOnly"
                  + sign_in_frequency                         = 10
                  + sign_in_frequency_period                  = "hours"
                }
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
--- PASS: TestAccConditionalAccessPolicy_complete (45.17s)
--- PASS: TestAccConditionalAccessPolicy_basic (46.88s)
--- FAIL: TestAccConditionalAccessPolicy_update (48.58s)
FAIL
FAIL    github.com/hashicorp/terraform-provider-azuread/internal/services/conditionalaccess     48.591s
```

### DIFF
--- a/internal/services/conditionalaccess/conditional_access_policy_resource.go
+++ b/internal/services/conditionalaccess/conditional_access_policy_resource.go
@@ -568,22 +568,31 @@ func flattenConditionalAccessSessionControls(in *msgraph.ConditionalAccessSessio
 		return []interface{}{}
 	}
 
-	result := make(map[string]interface{})
-
+	applicationEnforceRestrictions := false
 	if in.ApplicationEnforcedRestrictions != nil {
-		result["application_enforced_restrictions_enabled"] = in.ApplicationEnforcedRestrictions.IsEnabled
+		applicationEnforceRestrictions = *in.ApplicationEnforcedRestrictions.IsEnabled
 	}
 
-	if in.CloudAppSecurity != nil {
-		result["cloud_app_security_policy"] = in.CloudAppSecurity.CloudAppSecurityType
+	cloudAppSecurity := ""
+	if in.CloudAppSecurity != nil && in.CloudAppSecurity.CloudAppSecurityType != nil {
+		cloudAppSecurity = *in.CloudAppSecurity.CloudAppSecurityType
 	}
 
-	if in.SignInFrequency != nil {
-		result["sign_in_frequency"] = in.SignInFrequency.Value
-		result["sign_in_frequency_period"] = in.SignInFrequency.Type
+	signInFrequency := 0
+	signInFrequencyPeriod := ""
+	if in.SignInFrequency != nil && in.SignInFrequency.Value != nil && in.SignInFrequency.Type != nil {
+		signInFrequency = int(*in.SignInFrequency.Value)
+		signInFrequencyPeriod = *in.SignInFrequency.Type
 	}
 
-	return []interface{}{result}
+	return []interface{}{
+		map[string]interface{}{
+			"application_enforced_restrictions_enabled": applicationEnforceRestrictions,
+			"cloud_app_security_policy":                 cloudAppSecurity,
+			"sign_in_frequency":                         signInFrequency,
+			"sign_in_frequency_period":                  signInFrequencyPeriod,
+		},
+	}
 }
 
 func expandConditionalAccessConditionSet(in []interface{}) *msgraph.ConditionalAccessConditionSet {
@@ -724,20 +733,20 @@ func expandConditionalAccessSessionControls(in []interface{}) *msgraph.Condition
 	signInFrequency := config["sign_in_frequency"].(int)
 
 	var result msgraph.ConditionalAccessSessionControls
-
+	// The API doesn't accept boolean false values, we must instead omit them
 	if applicationEnforcedRestrictions {
 		result.ApplicationEnforcedRestrictions = &msgraph.ApplicationEnforcedRestrictionsSessionControl{
 			IsEnabled: utils.Bool(applicationEnforcedRestrictions),
 		}
 	}
-
+	// The API doesn't accept boolean false values, we must instead omit related fields
 	if cloudAppSecurity != "" {
 		result.CloudAppSecurity = &msgraph.CloudAppSecurityControl{
 			IsEnabled:            utils.Bool(true),
 			CloudAppSecurityType: utils.String(cloudAppSecurity),
 		}
 	}
-
+	// The API doesn't accept boolean false values, we must instead omit related fields
 	if signInFrequency > 0 {
 		result.SignInFrequency = &msgraph.SignInFrequencySessionControl{
 			IsEnabled: utils.Bool(true),
--- a/internal/services/conditionalaccess/conditional_access_policy_resource.go
+++ b/internal/services/conditionalaccess/conditional_access_policy_resource.go
@@ -29,6 +29,8 @@ func conditionalAccessPolicyResource() *schema.Resource {
 		UpdateContext: conditionalAccessPolicyResourceUpdate,
 		DeleteContext: conditionalAccessPolicyResourceDelete,
 
+		CustomizeDiff: conditionalAccessPolicyCustomizeDiff,
+
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(5 * time.Minute),
 			Read:   schema.DefaultTimeout(5 * time.Minute),
@@ -339,9 +341,10 @@ func conditionalAccessPolicyResource() *schema.Resource {
 				},
 			},
 			"session_controls": {
-				Type:     schema.TypeList,
-				Optional: true,
-				MaxItems: 1,
+				Type:             schema.TypeList,
+				Optional:         true,
+				MaxItems:         1,
+				DiffSuppressFunc: conditionalAccessPolicyDiffSuppress,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"application_enforced_restrictions_enabled": {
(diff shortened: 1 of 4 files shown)


kaovd:
Nice catch, do you think we need a bunch of tests to make sure we've got 100% coverage or this will be good?

alexwilcox9:
@kaovd I'm not sure really, there are so many different combinations we could produce tests for there has to be a balance and the real world will always produce something we didn't plan for.
Will see what @manicminer thinks when they're back

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -568,14 +568,22 @@ func flattenConditionalAccessSessionControls(in *msgraph.ConditionalAccessSessio
...
+	if in.CloudAppSecurity != nil {
+		result["cloud_app_security_policy"] = in.CloudAppSecurity.CloudAppSecurityType
+	}
+
+	if in.SignInFrequency != nil {
+		result["sign_in_frequency"] = in.SignInFrequency.Value
+		result["sign_in_frequency_period"] = in.SignInFrequency.Type
 	}
+
+	return []interface{}{result}
```
manicminer, line 586: This makes sense as the API doesn't seem to be setting these fields at all when they are "disabled". However, since we always want to set values in state for nested properties, even when they are the default values, can we rather do something like this to ensure that values are always set?

```go
func flattenConditionalAccessSessionControls(in *msgraph.ConditionalAccessSessionControls) []interface{} {
	if in == nil {
		return []interface{}{}
	}

	applicationEnforceRestrictions := false
	if in.ApplicationEnforcedRestrictions != nil {
		applicationEnforceRestrictions = *in.ApplicationEnforcedRestrictions.IsEnabled
	}

	cloudAppSecurity := ""
	if in.CloudAppSecurity != nil && in.CloudAppSecurity.CloudAppSecurityType != nil {
		cloudAppSecurity = *in.CloudAppSecurity.CloudAppSecurityType
	}

	signInFrequency := 0
	signInFrequencyPeriod := ""
	if in.SignInFrequency != nil && in.SignInFrequency.Value != nil && in.SignInFrequency.Type != nil {
		signInFrequency = int(*in.SignInFrequency.Value)
		signInFrequencyPeriod = *in.SignInFrequency.Type
	}

	return []interface{}{
		map[string]interface{}{
			"application_enforced_restrictions_enabled": applicationEnforceRestrictions,
			"cloud_app_security_policy":                 cloudAppSecurity,
			"sign_in_frequency":                         signInFrequency,
			"sign_in_frequency_period":                  signInFrequencyPeriod,
		},
	}
}
```

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -715,19 +723,27 @@ func expandConditionalAccessSessionControls(in []interface{}) *msgraph.Condition
 	cloudAppSecurity := config["cloud_app_security_policy"].(string)
 	signInFrequency := config["sign_in_frequency"].(int)
 
-	result := msgraph.ConditionalAccessSessionControls{
-		ApplicationEnforcedRestrictions: &msgraph.ApplicationEnforcedRestrictionsSessionControl{
+	var result msgraph.ConditionalAccessSessionControls
+
+	if applicationEnforcedRestrictions {
```
manicminer, line 728: Might be good to leave a comment here explaining why we don't set it.

```suggestion
	// The API doesn't accept boolean false values, we must instead omit them
	if applicationEnforcedRestrictions {
```

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -715,19 +723,27 @@ func expandConditionalAccessSessionControls(in []interface{}) *msgraph.Condition
...
+		result.ApplicationEnforcedRestrictions = &msgraph.ApplicationEnforcedRestrictionsSessionControl{
 			IsEnabled: utils.Bool(applicationEnforcedRestrictions),
-		},
-		CloudAppSecurity: &msgraph.CloudAppSecurityControl{
-			IsEnabled:            utils.Bool(cloudAppSecurity != ""),
+		}
+	}
+
+	if cloudAppSecurity != "" {
```
manicminer, line 734:
```suggestion
	// The API doesn't accept boolean false values, we must instead omit related fields
	if cloudAppSecurity != "" {
```

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -715,19 +723,27 @@ func expandConditionalAccessSessionControls(in []interface{}) *msgraph.Condition
...
+		result.CloudAppSecurity = &msgraph.CloudAppSecurityControl{
+			IsEnabled:            utils.Bool(true),
 			CloudAppSecurityType: utils.String(cloudAppSecurity),
-		},
-		SignInFrequency: &msgraph.SignInFrequencySessionControl{
-			IsEnabled: utils.Bool(signInFrequency > 0),
+		}
+	}
+
+	if signInFrequency > 0 {
```
manicminer, line 741:
```suggestion
	// The API doesn't accept boolean false values, we must instead omit related fields
	if signInFrequency > 0 {
```

manicminer (dismissed):
Hi @alexwilcox9 thanks for picking this up. This seems to be down to API bugs where it won't allow boolean values to be false, but insists they be omitted entirely to unset them!

The approach you've taken is good, my only suggestion would be for the flatten funcs, that they always set values even when those values are the default. Suggested approach inline below :)

We also seem to have missed some nil checks in the initial implementation which would account for the crashes you saw. I added nil checks for all nested pointers in my below suggestion.

In terms of tests, I think we're OK for these (API bug causing) cases to not litter the tests with all the different permutations as that will explode exponentially.

internal/services/conditionalaccess/conditional_access_policy_resource.go
alexwilcox9: Makes sense to me, I've added that verbatim

alexwilcox9:
Thanks for the feedback @manicminer I've now added the requested changes
What do you think about the occasionally failing update test? Is it worth building a refresh func in?

manicminer:
Note to reviewer: This change is technically breaking since some Optional fields are now Required, but those fields were required by the API anyway and omitting them would have failed to apply. The updated `expandConditionalAccessSessionControls()` works around some peculiar API behaviors, pasted below from the commit message for visibility.

Adjust the behaviour of the `sessionControls` field
  - The API is very fussy about the field in POST requests
  - Where it has no features enabled, it must be omitted
  - Where a feature in this section is disabled, that feature must be omitted (`isEnabled: false` is not recognized in POST requests)
  - This does not apply to PATCH requests, where `sessionControls` can be omitted when no contained features are enabled, but if any are enabled, _all_ the features must be specified with their respective `isEnabled` field (which can be `false` for PATCH requests).

manicminer:
**Test results**

Screenshot 2021-09-15 at 15 27 59

jackofallops (approved):
This LGTM 🚀

Issue #570, opened 2021-09-13
Keepers field should be documented azuread_service_principal_password


nikolaifa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The field 
* Keepers

should be documented as an argument https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/service_principal_password
and not just within an example

### New or Affected Resource(s)

azuread_service_principal_password

* #0000

manicminer:
Thanks for spotting this omission @nikolaifa! This will be added in the next release.

PR #572, opened 2021-09-13
Document `keepers` property for application and service principal passwords

manicminer:
Closes: #570

### DIFF
--- a/docs/resources/application_password.md
+++ b/docs/resources/application_password.md
@@ -43,7 +43,7 @@ resource "time_rotating" "example" {
 
 resource "azuread_application_password" "example" {
   application_object_id = azuread_application.example.object_id
-  keepers = {
+  rotate_when_changed = {
     rotation = time_rotating.example.id
   }
 }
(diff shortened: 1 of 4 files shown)


katbyte (approved):
LGTM 🩹

katbyte:
thou one might argue that "keepers" is a bad name for this property as how does that equate "rotate password when" 🤔

manicminer:
Yeah it's not the most intuitive name tbf. I took the concept and the naming from the [random provider](https://registry.terraform.io/providers/hashicorp/random/latest/docs#resource-keepers) where it seems to be aiming for a generic approach. Totally happy to rename it to something more specific, though I'm struggling to think of anything good :D

katbyte:
ahh given there is a keeper password manager might be best to change hah, doesn't seem like its a standard term for it? - no reason not to be verbose `rotate_password_when_changed` ?

manicminer:
> ahh given there is a keeper password manager might be best to change hah, doesn't seem like its a standard term for it? - no reason not to be verbose `rotate_password_when_changed` ?

Sounds perfect 👌

manicminer:
@katbyte Have renamed this, do you think we need to support the old property name given that it was undocumented?

katbyte (approved):
like it! thanks for making that change 🙌  LGTM

Issue #418, opened 2021-03-12
Support for Azure App Registration Logo


thecodetinker:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As a Terraform user I would like the ability to configure a Logo on an App Registration for branding purposes. Currently this is possible via PowerShell, but obviously would be great to set this in Terraform.

https://docs.microsoft.com/en-us/powershell/module/azuread/set-azureadapplicationlogo?view=azureadps-2.0

### New or Affected Resource(s)

* azuread_application

### Potential Terraform Configuration

```hcl
resource "azuread_application" "example" {
  name = "MyApp"
  logo = "./logo.png" # or maybe filebase64("./logo.png") depending on implementation
}
```

### References

* #0000

manicminer:
Hi @thecodetinker, thanks for requesting this feature! We will make sure to include this on our roadmap for Applications once we have [released support for Microsoft Graph](https://github.com/hashicorp/terraform-provider-azuread/issues/323).

PR #574, opened 2021-09-14
azuread_application: support the `logo` property

manicminer:
Depends on: https://github.com/manicminer/hamilton/pull/103
Closes: #418

Support for uploading a logo image for an application. Due to API limitations a logo cannot be removed once uploaded, unless it is replaced with another uploaded image.


katbyte (approved):
LGTM 🏗️

PR #597, opened 2021-09-25
Add support for a few additional users attributes

michaljirman:
Add support for a few additional users attributes:  
* employeeType
* costCenter
* division

Depends on Hamilton update:
[PR](https://github.com/manicminer/hamilton/pull/99)


internal/services/users/user_data_source.go
```diff
@@ -122,6 +122,24 @@ func userDataSource() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+
+			"division": {
+				Description: "The name of the division in which the user works.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+
```
manicminer, line 142: I think these should be Computed rather than Optional? Also could we sort these fields broadly in the following precedence: Required, then Optional, then Computed, and each of those sections alphanumerically?

internal/services/users/user_data_source.go
```diff
@@ -359,6 +377,7 @@ func userDataSourceRead(ctx context.Context, d *schema.ResourceData, meta interf
 	tf.Set(d, "department", user.Department)
 	tf.Set(d, "display_name", user.DisplayName)
 	tf.Set(d, "employee_id", user.EmployeeId)
+	tf.Set(d, "employee_type", user.EmployeeType)
```
manicminer, line 380: It looks like `cost_center` and `division` are missing here?

docs/data-sources/user.md
```diff
@@ -47,6 +47,9 @@ The following attributes are exported:
 * `department` - The name for the department in which the user works.
 * `display_name` - The display name of the user.
 * `employee_id` - The employee identifier assigned to the user by the organisation.
+* `employee_type` - (Optional) Captures enterprise worker type. For example, Employee, Contractor, Consultant, or Vendor.
+* `cost_center` - (Optional) The cost center associated with the user.
+* `division` - (Optional) The name of the division in which the user works.
```
manicminer, line 52: As these are attributes, we can drop the `(Optional)` prefix
```suggestion
* `employee_type` - Captures enterprise worker type. For example, Employee, Contractor, Consultant, or Vendor.
* `cost_center` - The cost center associated with the user.
* `division` - The name of the division in which the user works.
```

docs/resources/user.md
```diff
@@ -41,6 +41,9 @@ The following arguments are supported:
 * `disable_strong_password` - (Optional) Whether the user is allowed weaker passwords than the default policy to be specified. Defaults to `false`.
 * `display_name` - (Required) The name to display in the address book for the user.
 * `employee_id` - (Optional) The employee identifier assigned to the user by the organisation.
+* `employee_type` - (Optional) Captures enterprise worker type. For example, Employee, Contractor, Consultant, or Vendor.
+* `cost_center` - (Optional) The cost center associated with the user.
+* `division` - (Optional) The name of the division in which the user works.
```
manicminer, line 46: These should be sorted in alphanumeric order.

internal/services/users/user_resource.go
```diff
@@ -133,6 +133,25 @@ func userResource() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+
+			"division": {
+				Description: "The name of the division in which the user works.",
+				Type:        schema.TypeString,
+				Optional:    true,
+			},
+
```
manicminer, line 154: As above, could we order these fields so that all the Computed fields are first, then the Optional ones, and then the Computed ones, and each section in alphanumeric order?

internal/services/users/user_resource.go
```diff
@@ -426,6 +445,19 @@ func userResourceCreate(ctx context.Context, d *schema.ResourceData, meta interf
...
+	}
+
+	costCenter, costCenterExists := d.GetOk("cost_center")
+	division, divisionExists := d.GetOk("division")
+	if costCenterExists || divisionExists {
+		properties.EmployeeOrgData = &msgraph.EmployeeOrgData{
+			CostCenter: utils.String(costCenter.(string)),
+			Division:   utils.String(division.(string)),
+		}
+	}
```
manicminer, line 459: As these are not Optional+Computed fields, we should set them every time even if they are blank - `d.Get()` will return the default value if a property is not configured.

internal/services/users/user_resource.go
```diff
@@ -514,6 +547,13 @@ func userResourceUpdate(ctx context.Context, d *schema.ResourceData, meta interf
 		properties.OnPremisesImmutableId = utils.String(d.Get("onpremises_immutable_id").(string))
 	}
 
+	if d.HasChange("cost_center") || d.HasChange("division") {
+		properties.EmployeeOrgData = &msgraph.EmployeeOrgData{
+			CostCenter: utils.String(d.Get("cost_center").(string)),
+			Division:   utils.String(d.Get("division").(string)),
+		}
+	}
+
```
manicminer, line 556: As above, when updating we should also set these every time. _Generally_, we only need to use `d.HasChange()` for Optional+Computed fields.

internal/services/users/user_resource_test.go
```diff
@@ -186,6 +186,9 @@ resource "azuread_user" "test" {
   department                 = "acctestUser-%[1]d-Dept"
   display_name               = "acctestUser-%[1]d-DisplayName"
   employee_id                = "%[3]s%[3]s"
+  employee_type              = "Contractor"
+  cost_center              	 = "acctestUser-%[1]d-CostCenter"
+  division                   = "acctestUser-%[1]d-Division"
```
manicminer, line 191: Could we also sort these? :)

manicminer (changes requested):
Hi @michaljirman, thanks for this PR! Overall this looks good, I have a few suggestions inline below on the field handling in the Create/Update funcs, and also some of the ordering in the schema and docs (the linter also picked up on one of these). If we can fix these up, this should be good to merge!

internal/services/users/user_resource_test.go
michaljirman: done

internal/services/users/user_resource.go
michaljirman: done, good point!

internal/services/users/user_resource.go
michaljirman: done

internal/services/users/user_resource.go
michaljirman: reordered - hopefully :-)

docs/resources/user.md
michaljirman: done

docs/data-sources/user.md
michaljirman: done

internal/services/users/user_data_source.go
michaljirman: good point; added at the bottom of the function

internal/services/users/user_data_source.go
michaljirman: done

michaljirman:
> Hi @michaljirman, thanks for this PR! Overall this looks good, I have a few suggestions inline below on the field handling in the Create/Update funcs, and also some of the ordering in the schema and docs (the linter also picked up on one of these). If we can fix these up, this should be good to merge!

Thanks a lot for the review! I have pushed the code changes based on the code review.
Considering this is my first contribution to the official version of this provider, please kindly review it thoroughly. It is more likely I have missed more things than it is common for the other contributors ;-) 

Thanks!

manicminer:
@michaljirman Thanks for making the changes! This mostly LGTM, I had to make a small fix upstream for the `EmployeeType` field which is nullable (doesn't accept blank strings). Once https://github.com/manicminer/hamilton/pull/110 and then #608 are merged, this should also be good to merge 👍

manicminer:
Test results

Screenshot 2021-09-30 at 21 32 22

manicminer (approved):

PR #616, opened 2021-10-06
Workaround for corrupted or missing `@odata.id` for directory objects

manicminer:
We use the `@odata.id` field (`DirectoryObject.ODataId`) to reference objects for `@odata.bind` fields, e.g. when creating applications, groups, service principals, or posting new members for groups or directory roles. Two breaking changes were recently introduced in the API:

1. `@odata.id` stopped being returned for requests with `Accept: odata.metadata=minimal`. Hamilton has a fix for this in v0.32.0.
2. The format of the `@odata.id` value has changed for some tenants. Previously it was always the URI of the object, now in many cases despite the odata.json spec suggesting this should be a URI, it takes the form `directoryObject('0000...')`. This form isn't recognised by other API endpoints like applications, groups etc.

To work around (2), we're manually constructing the OData ID to look like a URI. This is intended to be a temporary fix until the API stabilizes.

Fixes: https://github.com/hashicorp/terraform-provider-azuread/issues/588


internal/services/applications/application_resource.go
```diff
@@ -918,6 +918,13 @@ func applicationResourceCreate(ctx context.Context, d *schema.ResourceData, meta
 	if callerObject == nil {
 		return tf.ErrorDiagF(errors.New("returned callerObject was nil"), "Could not retrieve calling principal object %q", callerId)
 	}
+	// TODO: remove this workaround for https://github.com/hashicorp/terraform-provider-azuread/issues/588
+	//if callerObject.ODataId == nil {
+	//	return tf.ErrorDiagF(errors.New("ODataId was nil"), "Could not retrieve calling principal object %q", callerId)
+	//}
+	callerObject.ODataId = (*odata.Id)(utils.String(fmt.Sprintf("%s/v1.0/%s/directoryObjects/%s",
+		client.BaseClient.Endpoint, client.BaseClient.TenantId, callerId)))
```
manicminer, line 926: Note: this intentionally hardcodes the API version into the URI as that's what the API accepts, regardless of the actual version

brnwn4:
@manicminer  - Does this mean we can still use the 2.5 azurerm provider and expect it to work? Or are we still pending the bug fix release tomorrow?

bardholthe:
We're also having this issue.

RafaelM1994:
We are also impacted by this issue. I hope we can get this new version today.

Issue #327, opened 2020-09-30
Allow Application Tags To Be Configured


AshleyPoole:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Feature request is to allows application tags to be configured via Terraform when creating an applicant, so it works similar to editing the manifest within the Portal UI.

### New or Affected Resource(s)

azuread_application

### Potential Terraform Configuration

```hcl
resource "azuread_application" "example" {
  name                       = "example"
  tags = ["example", "tags", "here"]
}
```

manicminer:
@AshleyPoole thanks for requesting this. As applicaiton tags are only supported in the newer MS Graph API, and not the older AAD Graph API, we'll be able to support this attribute once we have migrated to the new API (see #323 for details).

jdelforno:
Tag everything! :D

parameshjava:
Is there any SLA for this please?

PR #630, opened 2021-10-12
azuread_application: support the `features` block and the `tags` property

manicminer:
Closes: #327


manicminer:
Test results

Screenshot 2021-10-14 at 23 48 07

Screenshot 2021-10-14 at 23 48 14

Issue #652, opened 2021-10-30
azuread_service_principal not properly provisioned before running its dependencies


vp777:
I am trying to create an application with a service principal using the following config:

```
resource "azuread_application" "main" {
  display_name    = "main-reader"
  sign_in_audience = "AzureADMyOrg"
  fallback_public_client_enabled = true
  
  web {
    implicit_grant {
      access_token_issuance_enabled = true
      id_token_issuance_enabled     = true
    }
  }
}

resource "azuread_service_principal" "main" {
  application_id = "${azuread_application.main.application_id}"
  
  provisioner "local-exec" {
       #the following command is used to grant admin consent for certain permissions
	command = "az ad app permission grant --id ${azuread_service_principal.main.application_id} --api 00000003-0000-0000-c000-000000000000 --scope 'User.Read.All'" #this command fails
	#command = "sleep 60;az ad app permission grant --id ${azuread_service_principal.main.application_id} --api 00000003-0000-0000-c000-000000000000 --scope 'User.Read.All'" #this command runs smoothly

	interpreter = ["powershell", "-c"]
  }
}
```

But while trying to provision this configuration i get the error:
```
│ Error running command 'az ad app permission grant --id 602e4230-8e6b-4cdb-8daf-da9a097d2055 --api 00000003-0000-0000-c000-000000000000 --scope 'User.Read.All'': exit status 1. Output:
│ ERROR: Service principal with appId or objectId '602e4230-8e6b-4cdb-8daf-da9a097d2055' doesn't exist. If '602e4230-8e6b-4cdb-8daf-da9a097d2055' is an appId, make sure an associated
│ service principal is created for the app. To create one, run `az ad sp create --id 602e4230-8e6b-4cdb-8daf-da9a097d2055`
```

But if i add a delay, e.g. sleep 60 before trying to grant a permission then it executes correctly.
It looks like the azuread_service_principal returns prematurely. I have also tried using a null_resource with dependency on the service principal but i had the same result.

katbyte:
Hey @vp777 - Sorry your having this issue but i believe it is a problem with the AAD provider not azurerm so as such i am going to transfer it

manicminer:
[...]

PR #656, opened 2021-11-03
Mitigate some replication delays after creating some resources

manicminer:
- For applications, service principals, groups and users
- After creating, make a PATCH request with no changes
- Oftentimes if replication is very delayed, the PATCH request will fail
- The SDK handles retries of the PATCH request until it succeeds or times out

Closes: #652


koikonom (approved):
LGTM

manicminer:
Test results (failures are unrelated)

Screenshot 2021-11-03 at 13 59 21

Screenshot 2021-11-03 at 14 35 11

Screenshot 2021-11-03 at 14 35 40

Screenshot 2021-11-03 at 14 35 53

PR #663, opened 2021-11-16
Bring documentation inline with resource

GeorgeShort:
Currently, oauth2_permission_scope is incorrect as the resource within application_data_source.go is oauth2_permission_scopes. Therefore when performing lookups via oauth2_permission_scope, currently this is failing.


manicminer (approved):
Thanks for the docs fix @GeorgeShort! LGTM 👍

PR #670, opened 2021-11-19
Update Hamilton to v0.35.0

manicminer:
- Removes support for German national cloud


jackofallops (approved):
LGTM 🚀

tombuildsstuff:
FWIW we should call this out in the upgrade notes I guess? Whilst this is a breaking change since it's no longer available that seems like a valid exception here

manicminer:
SGTM, will call this out as a breaking change

PR #672, opened 2021-11-22
Support for Administrative Units and their members

manicminer:
Depends on https://github.com/manicminer/hamilton/pull/124 (requires rebase)

* **New Data Source:** `azuread_administrative_unit`
* **New Resource:** `azuread_administrative_unit`
* **New Resource:** `azuread_administrative_unit_member`


manicminer:
Test results

Screenshot 2021-11-22 at 18 56 31

docs/resources/administrative_unit.md
```diff
@@ -0,0 +1,51 @@
...
+
+The following arguments are supported:
+
+* `description` - (Optional) The description of the administrative unit.
+* `display_name` - (Required) The display name of the administrative unit.
+* `members` - (Optional) A set of object IDs of members who should be present in this administrative unit. Supported object types are Users or Groups.
+
+!> **Warning** Do not use the `members` property at the same time as the [azuread_administrative_unit_member](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/administrative_unit_member) resource for the same administrative unit. Doing so will cause a conflict and administrative unit members will be removed.
+
+* `visibility` - (Optional) Whether the administrative unit _and_ its members are hidden or publicly viewable in the directory. Must be one of: `Hiddenmembership` or `Public`. Defaults to `Public`.
```
katbyte, line 37: is there just two options here? would it make sense to make this a boolean such as `publicly_visible` ?

katbyte (approved):
aside from one small comment LGTM 🏗️

docs/resources/administrative_unit.md
manicminer: I went back and forth on this one - right now there are two documented values but the analogous field for Groups has _three_ values and I'm half expecting this one to go the same way. I'll reach out and see if I can find out.

manicminer: It doesn't look like there is any imminent support for more values here, so I'll reduce it to a boolean 👍

manicminer:
Tests still passing

docs/resources/administrative_unit.md
katbyte: ahh if its going to be a third (valid value that makes sense to expose as sometimes there are ones that make little sense) then yea it makes sense as a string

manicminer: In this case a third value doesn't look imminent. If this feature does evolve in time we can deprecate, I think a boolean is the right call here.

Issue #669, opened 2021-11-19
Conditional Access - add Device Filter conditions to tf file argument syntax


josefismael:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Conditional Access Device filter conditions are now GA, but resource documentation does not include the argument syntax for this condition. CA policies with this condition defined are returned when executing an import, however this block is missing from the result.

### New or Affected Resource(s)

`azuread_conditional_access_policy`

### Potential Terraform Configuration

```hcl
device_filter {
            mode = [
                "Exclude",
            ]
            filter = [
                "Device.isCompliant -eq True -or device.trustType -eq 'ServerAD'",
            ]
        }
```

### References
https://docs.microsoft.com/en-us/azure/active-directory/conditional-access/concept-condition-filters-for-devices#filter-for-devices-graph-api

https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/conditional_access_policy#attributes-reference

* #0000

manicminer:
Thanks for requesting this @josefismael, we'll look to add it as soon as we can. Thanks!

PR #673, opened 2021-11-22
Conditional Access improvements, support for device conditions

manicminer:
Depends on https://github.com/manicminer/hamilton/pull/125 and requires rebase

Improvements for azuread_conditional_access:

- use SDK constants for enum validation
- reflect new API handling of sessionControls property
- make all enums case insensitive since the API treats them as such
- support the `device_filter` block (intentionally within a `devices` block as there are additional properties to come)

Closes: #669


docs/resources/conditional_access_policy.md
```diff
@@ -31,6 +31,13 @@ resource "azuread_conditional_access_policy" "example" {
       excluded_applications = ["00000004-0000-0ff1-ce00-000000000000"]
     }
 
+    devices {
+      device_filter {
```
katbyte, line 35: devices.device_filter seems a tad redundant?
```suggestion
    devices {
      filter {
```

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -172,18 +174,53 @@ func conditionalAccessPolicyResource() *schema.Resource {
...
+							},
+						},
+
+						"devices": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"device_filter": {
```
katbyte, line 199: given devices is just a block with a block inside, would it make sense to just move device_filter up a level and remove the devices block entirely?

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -172,18 +174,53 @@ func conditionalAccessPolicyResource() *schema.Resource {
...
+										Type:     schema.TypeList,
+										Optional: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"mode": {
+													Type:             schema.TypeString,
+													Required:         true,
+													DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 208: how come we are suppressing case and not normalizing it to what is expected/correct?

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -224,33 +261,35 @@ func conditionalAccessPolicyResource() *schema.Resource {
 										Type:     schema.TypeList,
 										Required: true,
 										Elem: &schema.Schema{
-											Type: schema.TypeString,
+											Type:             schema.TypeString,
+											DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 265: ditto here

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -224,33 +261,35 @@ func conditionalAccessPolicyResource() *schema.Resource {
...
 										},
 									},
 
 									"excluded_platforms": {
 										Type:     schema.TypeList,
 										Optional: true,
 										Elem: &schema.Schema{
-											Type: schema.TypeString,
+											Type:             schema.TypeString,
+											DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 283: and here

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -261,31 +300,33 @@ func conditionalAccessPolicyResource() *schema.Resource {
 							Type:     schema.TypeList,
 							Optional: true,
 							Elem: &schema.Schema{
-								Type: schema.TypeString,
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 304: and here

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -261,31 +300,33 @@ func conditionalAccessPolicyResource() *schema.Resource {
...
 							},
 						},
 
 						"user_risk_levels": {
 							Type:     schema.TypeList,
 							Optional: true,
 							Elem: &schema.Schema{
-								Type: schema.TypeString,
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 321: and here

internal/services/conditionalaccess/conditional_access_policy_resource.go
```diff
@@ -307,17 +348,18 @@ func conditionalAccessPolicyResource() *schema.Resource {
 							Type:     schema.TypeList,
 							Required: true,
 							Elem: &schema.Schema{
-								Type: schema.TypeString,
+								Type:             schema.TypeString,
+								DiffSuppressFunc: suppress.CaseDifference,
```
katbyte, line 352: ect

katbyte (changes requested):
Thanks @manicminer - couple comments with my main one being why are we ignoring case? we typically normalize it to the correct value of the API doesn't respect it?

docs/resources/conditional_access_policy.md
manicminer: SGTM, will rename

internal/services/conditionalaccess/conditional_access_policy_resource.go
manicminer: I went with the `device` block as there are a bunch of other properties in here that we'll add as soon as we have worked out togglable beta/preview support, or those properties make it to the stable API (whichever comes first) - but this will happen fairly soon so I didn't want to cause a deprecation a few weeks down the line.

internal/services/conditionalaccess/conditional_access_policy_resource.go
manicminer: Good point - the correct casing isn't always obvious as the docs are inconsistent, but I think you're right it makes sense to normalize these (and make judgement calls for consistency as necessary) 👍

manicminer: I'll do some more testing and see if we get any flapping diffs due to casing, if not I'll pull out the diff suppression

manicminer:
Test results

Screenshot 2021-11-25 at 14 38 52

katbyte (approved):
Thanks @manicminer - LGTM 🙌

PR #677, opened 2021-11-24
Conditional access policy - support persistent browser session control

stawik-mesa:
Improvements for azuread_conditional_access:

- support the persistent_browser_mode field


docs/resources/conditional_access_policy.md
```diff
@@ -148,6 +148,7 @@ The following arguments are supported:
 * `cloud_app_security_policy` - (Optional) Enables cloud app security and specifies the cloud app security policy to use. Possible values are: `blockDownloads`, `mcasConfigured`, `monitorOnly` or `unknownFutureValue`.
 * `sign_in_frequency` - (Optional) Number of days or hours to enforce sign-in frequency. Required when `sign_in_frequency_period` is specified. Due to an API issue, removing this property forces a new resource to be created.
 * `sign_in_frequency_period` - (Optional) The time period to enforce sign-in frequency. Possible values are: `hours` or `days`. Required when `sign_in_frequency_period` is specified. Due to an API issue, removing this property forces a new resource to be created.
+* `persistent_browser_mode` - (Optional) Session control to define whether to persist cookies or not. Possible values are: `always` or `never`.
```
manicminer, line 151: These will need to be ordered to pass documentation checks :)

manicminer (changes requested):
Hi @stawik-mesa, thanks so much for this contribution! It looks like we missed persistent browser session controls when adding this resource, and it'd be great to add this.

This mostly LGTM, I have a few comments below mainly on style/ordering, and a typo in the test config. Could you also look at adding this property in the `sessionControls` / `sessionControlsDisabled` test configs? We have to test these separately at the moment to ensure we have test coverage for an API bug workaround.

If you can take a look at these, and the tests pass, then this should be good to merge. Thanks!

stawik-mesa:
> Hi @stawik-mesa, thanks so much for this contribution! It looks like we missed persistent browser session controls when adding this resource, and it'd be great to add this.
> 
> This mostly LGTM, I have a few comments below mainly on style/ordering, and a typo in the test config. Could you also look at adding this property in the `sessionControls` / `sessionControlsDisabled` test configs? We have to test these separately at the moment to ensure we have test coverage for an API bug workaround.
> 
> If you can take a look at these, and the tests pass, then this should be good to merge. Thanks!

Hi @manicminer,

Thanks for the quick response on my PR. How should I integrate the test?
I haven't seen anything related to the other session control fields? How do you test them?

manicminer:
@stawik-mesa You can run the tests by running `make testacc TEST=./internal/services/conditionalaccess`, however we're also happy to run them.

When I try to run this and enable persistent browser mode, I'm getting a 400 error from the API:

```json
{
  "error": {
    "code": "BadRequest",
    "message": "1032: ConditionalActionPolicy validation failed due to InvalidConditionsForPersistentBrowserSessionMode.",
    "innerError": {
      "date": "2021-11-25T11:59:20",
      "request-id": "551250b2-b961-44fd-ad83-f6b1cf6b32dc",
      "client-request-id": "551250b2-b961-44fd-ad83-f6b1cf6b32dc"
    }
  }
}```

I looked back at earlier PRs when we added Conditional Access support, and I see that persistent browser mode was initially added but removed before the resource was merged. Whilst this wasn't discussed in the original PR< it might be that the API doesn't work for this property.

Since we unfortunately can't merge until setting this property works, I'll dig into it a little further and reach out to the service team to find out why we're getting this error.

**Comment by stawik-mesa** [comment]
is it maybe realted to: https://docs.microsoft.com/en-us/answers/questions/502950/conditional-access-exclude-application.html

**Comment by stawik-mesa** [comment]
@manicminer Locally all tests passed

**Comment by stawik-mesa** [comment]
@manicminer I hopefully fixed the format error on the test source code

**Comment by manicminer** [comment]
@stawik-mesa Thanks for resolving the merge! When running the tests I noticed that changes to `session_controls` were not being picked up and tracked it down to the DiffSuppressFunc (previously added by me!) incorrectly discarding changes to that block, causing it to be omitted from API requests and silently ignored by Terraform, in turn allowing the tests to falsely pass. I believe that's now fixed.

I also merged your additional tests in with the tests for the `session_controls` block, so that's now tested with the other sibling properties and it appears to be working well.

I'll open this up for review from another contributor and we should be able to get this merged for this week's release. Thanks again!

**Comment by manicminer** [comment]
Test results

Screenshot 2021-12-02 at 14 05 53

File: internal/services/conditionalaccess/conditional_access_policy_resource.go [hunk 2]
```diff
@@ -397,6 +397,15 @@ func conditionalAccessPolicyResource() *schema.Resource {
...
 						},
 
+						"persistent_browser_mode": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								msgraph.PersistentBrowserSessionModeAlways,
+								msgraph.PersistentBrowserSessionModeNever,
+							}, false),
+						},
```
katbyte, line 407: this seems to be a boolean so could become a bool? maybe
```suggestion
						"persistent_browser_cookies_enabled": {
							Type:     schema.TypeString,
							Optional: true,
							ValidateFunc: validation.StringInSlice([]string{
								msgraph.PersistentBrowserSessionModeAlways,
								msgraph.PersistentBrowserSessionModeNever,
							}, false),
						},
```

katbyte (approved):
aside from 1 minor comment LGTM 💠

internal/services/conditionalaccess/conditional_access_policy_resource.go
manicminer: It's more of a triple-value setting (there's an underlying bool which we infer the value of) so we'll probably have to stick with enum-like strings

Issue #697, opened 2021-12-10
prevent_duplicate_names prevents aad group from being recreated


josh-barker:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```shell
terraform version
Terraform v1.0.11
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.12.0
```

### Affected Resource(s)

* `azuread_group`

### Steps to Reproduce

Step 1: Create AAD Group

```hcl
resource "azuread_group" "example" {
  display_name = "example"

  prevent_duplicate_names = true
  security_enabled = true

  # assignable_to_role = true
}
```

```shell
terraform apply -auto-approve
...
azuread_group.example: Creating...
azuread_group.example: Creation complete after 1s [id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

Step 2: Modify AAD Group that causes replacement

```hcl
resource "azuread_group" "example" {
  display_name = "example"

  prevent_duplicate_names = true
  security_enabled = true

  assignable_to_role = true # forceNew is true on this property
}
```

```shell
terraform apply -auto-approve
azuread_group.example: Refreshing state... [id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx]
╷
│ Error: An existing "azuread_group" with name "example" (ID: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx") was found and `prevent_duplicate_names` was specified. To be managed via Terraform, this resource needs to be imported into the State. Please see the resource documentation for "azuread_group" for more information.
│ 
│   with azuread_group.example,
│   on main.tf line 5, in resource "azuread_group" "example":
│    5: resource "azuread_group" "example" {
```

### Expected behaviour

The group is updated and replaced without error.

### Important Factoids

This error also can be seen when importing an AAD Group and running a plan.

manicminer:
[...]

PR #702, opened 2021-12-16
fix: allow aad group to be recreated when prevent_duplicate_names is true

josh-barker:
This PR replicates the bug as described in #697 

It also shows a potential way to overcome the bug, but there may be a better way to achieve it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ website/vendor
 !command/test-fixtures/**/.terraform/
 
 .env.sh
+.env
 
 # goenv version file
 .go-version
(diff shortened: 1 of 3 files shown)


manicminer:
Hi @josh-barker, sorry for the delay in reviewing this. Thanks for the implementation, I think this is on the right track but I'm concerned about making decisions based on the RawState as this could have unintended side effects. You're right, there are no Plugin SDK helpers that can expose this right now, but I think the best course would be to send this upstream so I'll work on a PR for https://github.com/hashicorp/terraform-plugin-sdk to expose whether a diff involves a ForceNew - I think this is possible based on the data held in the raw diff.

ccadruvi:
Is there any update on the upstream change or this PR?

manicminer:
Superceded by #980

Issue #703, opened 2021-12-21
DirectoryObjects.BaseClient.Get() 403 during assigning the owners for azuread_application


tsologub:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
```
Terraform v0.12.31
azuread v2.12.0
```

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
provider "azuread" {
  client_id = "id"
  client_secret = "secret"
  tenant_id = "id"
}

provider "azurerm" {
  subscription_id = "id"
  client_id = "id"
  client_secret = "secret"
  tenant_id = "id"
  features {}
  alias = "provider_azurerm"
  skip_provider_registration = true
}

provider "random" {
  alias = "provider_random"
}

provider "time" {
  alias = "provider_time"
}

locals {
  password_create_wait = "3s"
  role_assignment_wait = "3s"
}

resource "azuread_application" "app_tas-graph" {
  display_name               = "tas-graph"
  identifier_uris            = []
  sign_in_audience           = "AzureADMyOrg"
  owners                     = [
    "object_id_from_authenticated_service_principal", 
    "human_object_id_who_is_the_owner_of_authenticated_service_principal", 
    "another_human_object_id_who_is_the_owner_of_authenticated_service_principal", 
    "another_human_object_id_who_is_the_owner_of_authenticated_service_principal", 
    "another_human_object_id_who_is_the_owner_of_authenticated_service_principal"
  ]
  prevent_duplicate_names    = true
  web {
    redirect_uris            = ["https://tas-graph/"]
  }
}

resource "azuread_service_principal" "spn_tas-graph" {
  application_id               = azuread_application.app_tas-graph.application_id
  app_role_assignment_required = true
  tags = [ "some", "tags" ]
  owners                     = [
    "object_id_from_authenticated_service_principal", 
    "human_object_id_who_is_the_owner_of_authenticated_service_principal", 
    "another_human_object_id_who_is_the_owner_of_authenticated_service_principal", 
[...]

PR #713, opened 2022-01-10
Don't call /directoryObjects for applications or service principals

manicminer:
No longer retrieve owner principals when managing applications or service principals. Since we no longer use the `@odata.id` returned by the API, it's not necessary to retrieve owner principals via the `/directoryObjects` endpoint. This can sometimes require additional permissions, which we did not document, and which is undesirable in any case.

Additionally, document additional potentially required permissions for groups as we can't get around this when managing groups, due to necessarily needing to distinguish between different types of owner principals.

Fixes: #703

**Test results**

Screenshot 2022-01-13 at 13 05 57

Screenshot 2022-01-13 at 12 51 05

### DIFF
--- a/docs/resources/group.md
+++ b/docs/resources/group.md
@@ -14,6 +14,8 @@ When authenticated with a service principal, this resource requires one of the f
 
 If using the `assignable_to_role` property, this resource additionally requires one of the following application roles: `RoleManagement.ReadWrite.Directory` or `Directory.ReadWrite.All`
 
+If specifying owners for a group, which are user principals, this resource additionally requires one of the following application roles: `User.Read.All`, `User.ReadWrite.All`, `Directory.Read.All` or `Directory.ReadWrite.All`
+
 When authenticated with a user principal, this resource requires one of the following directory roles: `Groups Administrator`, `User Administrator` or `Global Administrator`
 
 ## Example Usage
--- a/internal/services/applications/application_resource.go
+++ b/internal/services/applications/application_resource.go
@@ -981,10 +981,8 @@ func applicationResourceCreate(ctx context.Context, d *schema.ResourceData, meta
 	if callerObject == nil {
 		return tf.ErrorDiagF(errors.New("returned callerObject was nil"), "Could not retrieve calling principal object %q", callerId)
 	}
-	// TODO: remove this workaround for https://github.com/hashicorp/terraform-provider-azuread/issues/588
-	//if callerObject.ODataId == nil {
-	//	return tf.ErrorDiagF(errors.New("ODataId was nil"), "Could not retrieve calling principal object %q", callerId)
-	//}
+
+	// @odata.id returned by API cannot be relied upon, so construct our own
 	callerObject.ODataId = (*odata.Id)(utils.String(fmt.Sprintf("%s/v1.0/%s/directoryObjects/%s",
 		client.BaseClient.Endpoint, client.BaseClient.TenantId, callerId)))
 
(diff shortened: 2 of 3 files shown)


tsologub:
I do not know yet why, but this doesn't work for me. I receive the same:
```
Error: Could not create application
│
│   with azuread_application.app_taras-playground,
│   on playground.tf line 16, in resource "azuread_application" "app_taras-playground":
│   16: resource "azuread_application" "app_taras-playground" {
│
│ ApplicationsClient.BaseClient.Post(): unexpected status 403 with OData error: Authorization_RequestDenied: Insufficient privileges to complete the operation.
```

Testing with the following declaration:
```hcl
terraform {
  required_providers {
    azuread = {
      source = "registry.terraform.io/hashicorp/azuread"
    }
  }
}

# Has the following permissions: Application.ReadWrite.OwnedBy, User.Read
provider "azuread" {
  client_id     = "id"
  client_secret = "secret"
  tenant_id     = "id"
}

resource "azuread_application" "app_taras-playground" {
  display_name            = "taras-playground"
  owners                  = [
    "4ad0e09f-a951-4946-90a2-dacddea52079", # object-id of the SPN from above
    "3c422d16-cebe-4f1f-a282-2163a8ca3dcf"  # my object-id
  ]
  prevent_duplicate_names = true
  web {
    redirect_uris = ["https://taras-playground/"]
  }
}

resource "azuread_service_principal" "spn_taras-playground" {
  application_id               = azuread_application.app_taras-playground.application_id
  app_role_assignment_required = true
  owners                       = [
    "4ad0e09f-a951-4946-90a2-dacddea52079", # object-id of the SPN from above
    "3c422d16-cebe-4f1f-a282-2163a8ca3dcf"  # my object-id
  ]
}

resource "azuread_service_principal_password" "taras-playground" {
  service_principal_id = azuread_service_principal.spn_taras-playground.id
}
```

tsologub:
Correct me if I am wrong, when we perform [create application](https://docs.microsoft.com/en-us/graph/api/application-post-applications?view=graph-rest-1.0&tabs=http) and pass `owners = [...]` explicitly, the MSFT Graph API will perform [add owner](https://docs.microsoft.com/en-us/graph/api/application-post-owners?view=graph-rest-1.0&tabs=http) behind the scenes for every owner from our `owners` array. And according to the [Add Owner permissions section](https://docs.microsoft.com/en-us/graph/api/application-post-owners?view=graph-rest-1.0&tabs=http#permissions) _Required Application.ReadWrite.OwnedBy_ **and** _Directory.Read.All_. _Note: Application.ReadWrite.OwnedBy will not be sufficient to add another owner._

manicminer:
@tsologub There are two ways the provider assigns owners - when creating it will include up to 20 owners in the POST to /applications to create the app, and when updating (or when there are >20 owners to assign at creation) it will POST to the /applications/GUID/owners endpoint.

When you get the 401 error, does Terraform prefix the error output with "Could not create application", or "Could not add owners to application"? That will indicate which of the two calls are failing.

The docs state that Application.ReadWrite.All is required to create applications too, but this isn't always the case. So IMO it's not clear whether that same role is needed to append owners or not, since many operations on applications do work with Application.ReadWrite.OwnedBy. The effective permissions are also very different when authenticating as a user versus as a service principal.

tsologub:
> When you get the 401 error, does Terraform prefix the error output with "Could not create application", or "Could not add owners to application"?

In my case, I have two owners included. So the `2022-01-11T14:43:15.918+0100 [ERROR] vertex "azuread_application.app_taras-playground" error: Could not create application` first one is failing. 

> The effective permissions are also very different when authenticating as a user versus as a service principal - I'm assuming you're doing the latter as we're talking about app roles?

Yes, I am authenticating with the service principal.

jackofallops (approved):
LGTM 👍

Issue #729, opened 2022-02-05
external_senders_allowed failing even when unset with Service Principal authentication


andyshinn:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
❯ tfe terraform/vht-organization version
Terraform v1.1.4
on darwin_amd64
+ provider registry.terraform.io/hashicorp/archive v2.2.0
+ provider registry.terraform.io/hashicorp/aws v3.59.0
+ provider registry.terraform.io/hashicorp/azuread v2.16.0
```

### Affected Resource(s)

* `azuread_group`

### Terraform Configuration Files

```hcl
resource "azuread_group" "accounts_qa" {
  for_each                   = local.accounts
  display_name               = format(local.account_name_template, title(replace(each.key, "_", " ")), "QA")
  description                = format(local.azure_group_description_template, title(replace(each.key, "_", " ")), "QA")
  mail_enabled               = true
  mail_nickname              = format(each.value, "qa")
  security_enabled           = false
  types                      = ["Unified"]
  assignable_to_role         = false
  visibility                 = "Private"
  auto_subscribe_new_members = true

  behaviors = ["WelcomeEmailDisabled", "SubscribeNewGroupMembers"]

  members = local.email_group_owners
  owners  = local.email_group_owners

  # This is currently broken via API per https://docs.microsoft.com/en-us/graph/known-issues#allowexternalsenders-property-can-only-be-accessed-on-unified-groups
  lifecycle {
    ignore_changes = [external_senders_allowed]
  }
}
```

### Debug Output

I can debug if necessary.

### Expected Behavior

It should ignore patching the `external_senders_allowed` if it isn't set.

### Actual Behavior

It is failing with error because I previous had `external_senders_allowed = true` after manually changing the group setting outside of Terraform. After removing the option it is now failing to unset.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

I had previously manually enabled external sending for these groups. Was `external_senders_allowed` a noop when I added it to those existing ones?

[...]

PR #731, opened 2022-02-07
Bugfix: avoid setting features for unified groups when unconfigured

manicminer:
Make toggleable features for unified groups Computed and use GetOkExists to avoid sending them when unconfigured

Fixes: #729


stephybun (approved):
Thanks @manicminer, LGTM 🌴

PR #734, opened 2022-02-10
Clarifying note in azuread_application docs regarding the Application.ReadWrite.OwnedBy role

manicminer:
The note should be self explanatory, open to wording suggestions if this doesn't read well.


stephybun (approved):
Reads good to me, LGTM 👍🏼

Issue #732, opened 2022-02-08
Add azuread_service_principal_token_signing_certificate to simplify certificate management


dhohengassner:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Azure API provides a method (addTokenSigningCertificate) to create a self signed certificate for a service principal: https://docs.microsoft.com/en-us/graph/application-saml-sso-configure-api#create-a-signing-certificate

This option avoids managing a certificate by yourself and is currently not supported via terraform.

### New or Affected Resource(s)

* azuread_service_principal_token_signing_certificate

### References

https://docs.microsoft.com/en-us/graph/application-saml-sso-configure-api#step-4-configure-a-signing-certificate

This is a different request than https://github.com/hashicorp/terraform-provider-azuread/issues/410 because I do not want to manage the certificate myself.

**Thanks for your feedback!**

manicminer:
Hi @dhohengassner, thanks for requesting this feature! Looking through the linked walkthrough, it looks like a few additional pieces are needed here?

- New resource for token signing certificates
- New resource for claims mapping policies (a community member is already working on this)
- Support for the preferredTokenSigningKeyThumbprint field for service principals

Does this look right to you?

dhohengassner:
@manicminer thanks that is exactly what I had in mind :+1: 

I will try to come up with PRs for that. Thanks for the feedback

dhohengassner:
created a PR to add support for the preferredTokenSigningKeyThumbprint field for service principals

https://github.com/manicminer/hamilton/pull/151

dhohengassner:
@manicminer Experimented today with certification creation on service principal.

It seems the Graph API supports adding a certificate but there is no option to remove them. I can see the generated keys and passwords but I cannot remove them via `removePassword` or `removeKey` calls. In case `addtokensigningcertificate` is called it will add a new cert to the principal.

Ref: https://docs.microsoft.com/en-us/graph/api/serviceprincipal-addtokensigningcertificate

I am able to remove the certs on Enterprise apps but this is using a different internal API.

[...]

PR #740, opened 2022-02-17
✨ add support for token signing certificate

dhohengassner:
We do like to add support to the provider to create a token signing certificate.

This commit adds a new field `token_signing_certificate_name` to the
resource `azuread_service_principal`.

If this is set a ceritificate will be generated and the thumbprint
will be stored in a computed field called `preferred_token_signing_key_thumbprint`.

Issue: https://github.com/hashicorp/terraform-provider-azuread/issues/732

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -77,3 +77,5 @@ require (
 )
 
 go 1.17
+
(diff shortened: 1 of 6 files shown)


dhohengassner:
opened by accident - not yet ready

manicminer:
@dhohengassner That's no problem, you're welcome to open a draft PR if you're looking for early feedback

Issue #739, opened 2022-02-17
AAD plugin crash on azuread_conditional_access_policy each.value approach


josefismael:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v1.1.5
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v3.74.2
+ provider registry.terraform.io/hashicorp/azuread v2.18.0

### Affected Resource(s)

"azuread_conditional_access_policy" 

* `azuread_XXXXX`

### Terraform Configuration Files

Not at liberty to share config file, however basic logic is as follows:

./policy_properties.tfvars.v2.json - contains basic info about the resource (display_name, state, excluded_applications, etc)
./included_appliations.tfvars.json - contains list of included_applications
./modules/base-policy/base-policy-main.tf - "azuread_conditional_access_policy" with variables for each property
./main.tf - loops through policy_properties.tfvars.v2.json and sends info to base-policy module

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```

### Debug Output

https://gist.github.com/josefismael/338fc5ac33a266d872e4e8d59bc41818

### Expected Behavior

terraform config works as expected when included_applications is supplied to an already-populated "azuread_conditional_access_policy"  tf file

### Actual Behavior

receive error referenced in the gist

### Steps to Reproduce

terraform init
terraform plan

### References

* #0000

manicminer:
Hi @josefismael, many thanks for reporting this crash. I have pushed a fix which will be merged for the next release of the provider.

PR #747, opened 2022-03-01
Bugfix: nil check in diffsuppress func for conditional_access_policy

manicminer:
Fixes: #739


jackofallops (approved):
Thanks @manicminer - LGTM 🚀

Issue #721, opened 2022-01-25
Add in second app_role to existing `azuread_application` fails


StefanSchoof:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v1.1.4
on windows_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.15.0
+ provider registry.terraform.io/hashicorp/azurerm v2.93.1
### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

old:
```hcl
resource "azuread_application" "backend_ad_application" {
  count = var.ad_application == true ? 1 : 0

  display_name    = local.ad_application_name
  identifier_uris = ["api://apimanagement"]

  group_membership_claims = ["None"]

  owners = data.azuread_group.aop_ad_owner_group.members
  web {
    implicit_grant {
      access_token_issuance_enabled = true
      id_token_issuance_enabled     = true
    }
  }

  optional_claims {
    access_token {
      name = "email"
    }

    id_token {
      name = "email"
    }
  }

  required_resource_access {
    resource_app_id = "00000003-0000-0000-c000-000000000000" // graph api
    resource_access {
      id   = "e1fe6dd8-ba31-4d61-89e7-88639da4683d" // User.Read
      type = "Scope"
    }
    resource_access {
      id   = "64a6cdd6-aab1-4aaf-94b8-3cc8405e90d0" // mail
      type = "Scope"
    }
    resource_access {
      id   = "14dad69e-099b-42c9-810b-d002981feec1" // profile
      type = "Scope"
    }
  }

  app_role {
    allowed_member_types = ["User"]
    description          = "Company Admin"
    display_name         = "Company Admin"
    enabled              = true
    value                = "Admin"
    id                   = "683ff300-db10-48af-0482-fda9e35c6688"
  }

  api {
    requested_access_token_version = 2
    oauth2_permission_scope {
      admin_consent_description  = "Administer ${local.ad_application_name} application"
      admin_consent_display_name = "Administer ${local.ad_application_name} application"
      enabled                    = true
[...]

PR #750, opened 2022-03-07
azuread_application: Scope/Role validation for magic string "User"

manicminer:
Fixes: #721


jackofallops (approved):
Thanks @manicminer - LGTM 🚀

PR #768, opened 2022-04-07
Dependencies: update deps and use golangci-lint 1.45.2

manicminer:
Deferring update to Go 1.18 for now, due to https://github.com/bflad/tfproviderlint/issues/255


katbyte (approved):
LGTM 🌻

PR #769, opened 2022-04-11
Update claims_mapping_policy.md

MarkDordoy:
There is no description option within the internal service code


manicminer (approved):
Thanks for the fix @MarkDordoy!

Issue #779, opened 2022-04-26
SubscribeMembersToCalendarEventsDisabled missing as valid behavior in azuread_group


jlundy2:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v1.1.7
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.20.0
+ provider registry.terraform.io/hashicorp/azurerm v2.99.0
+ provider registry.terraform.io/hashicorp/http v2.1.0

### Affected Resource(s)

* `azuread_group`

### Terraform Configuration Files

```hcl
resource "azuread_group" "test_behavior_group" {
  display_name               = "test-behavior"
  description                = "Show bad behavior"
  owners                     = ["REDACTED"]
  security_enabled           = true
  assignable_to_role         = true
  prevent_duplicate_names    = true
  mail_enabled               = true
  mail_nickname              = "test"
  types                      = ["Unified"]
  members                    = []
  provisioning_options       = []
  behaviors                  = [
    "SubscribeMembersToCalendarEventsDisabled",
  ]
}
```

### Expected Behavior

Group should be created with that behavior. Or, as is our current issue, an imported group should not be recreated if it has this behavior.

### Actual Behavior

Error

```
Error: expected behaviors.0 to be one of [AllowOnlyMembersToPost HideGroupInOutlook SubscribeNewGroupMembers WelcomeEmailDisabled], got SubscribeMembersToCalendarEventsDisabled
│ 
│   with azuread_group.test_behavior_group,
│   on groups.tf line 317, in resource "azuread_group" "test_behavior_group":
│  317:   behaviors                  = [
│  318:     "SubscribeMembersToCalendarEventsDisabled",
│  319:   ]
```

### Steps to Reproduce

1. `terraform apply`

### References

This is the related issue on the microsoft graph api documentation: https://github.com/microsoftgraph/microsoft-graph-docs/issues/11504

manicminer:
@jlundy2 Thanks for reporting, we'll add this in tomorrow's provider release.

PR #785, opened 2022-04-27
`azuread_group` - support the `SubscribeMembersToCalendarEventsDisabled` value in the `behaviors` property

manicminer:
This isn't documented, see https://github.com/microsoftgraph/microsoft-graph-docs/issues/11504

Depends on: #786
Closes: #779


katbyte (approved):
LGTMN 🦀

PR #795, opened 2022-05-11
Updated Conditional Access Policy documentation

bubbletroubles:
The previous example had syntax which was not valid. This PR updates the example to be a working example which can be used by copy-pasting into a .tf file.


bubbletroubles:
@katbyte @manicminer can you review this one?

PR #810, opened 2022-05-27
Add sentance documenting naming mismatch for apps and SPs

sharebear:
Every time I look into applications and service principals I have to double and tripple check the Azure documentation to see if I got them the right way round. Someone on the Azure discord just claimed it was the other way around so I guess I'm not the only one struggling with this.


sharebear:
No wait... I've got this the wrong way around... again.

There's another nuance here I've not understood as well, what's the difference between an `azuread_service_principal` resource and `azuread_application` with `feature_tags.enterprise = true`?

If someone can help me understand this I can try and update the PR to be correct.

Issue #742, opened 2022-02-17
Azure AD - Scoped Assignments for Custom Azure AD roles


dpowley:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

We would like to see the provider support the ability to assign custom directory roles to a specific scope (i.e. application) instead of assigning the role with the entire directory as a scope.

See https://docs.microsoft.com/en-us/azure/active-directory/roles/custom-create#assign-a-custom-role-scoped-to-a-resource as to what I'm talking about.

### New or Affected Resource(s)

* azuread_directory_role_assignment

### Potential Terraform Configuration

```hcl
resource "azuread_app_role_assignment" "example" {
  directory_role_id    = azuread_custom_directory_role.example.object_id
  principal_object_id  = azuread_user.example.object_id
  resource_object_id   = azuread_service_principal.example.object_id
}
```

### References

* #0000

timothyleung1:
Thinking about working on something like `azuread_custom_directory_role_member`. Is this request asking for the same thing, but for service principal?

manicminer:
Hi @timothyleung1, you're welcome to take a look at this. I would prefer to use the existing `azuread_directory_role_member` resource to avoid resource proliferation and confusion, though that needs some investigation to see if it's feasible. The SDK supports the unified role management endpoints (see https://github.com/manicminer/hamilton/blob/main/msgraph/role_definitions.go and https://github.com/manicminer/hamilton/blob/main/msgraph/role_assignments.go) and this should allow us to manage assignments/memberships at the tenant level (as we do currently for built-in roles) and at all levels for custom roles.

PR #826, opened 2022-06-21
New resource: `azuread_directory_role_assignment`

manicminer:
This uses the newer "unified role assignment" endpoints to manage
directory role assignments for built-in roles and custom roles. Also
supports scoping via directory object IDs or apps.

Closes: #742


manicminer:
Test results (AAD P2)

Screenshot 2022-06-21 at 02 22 46

jackofallops (approved):
LGTM 🚀

Issue #813, opened 2022-06-02
Unable to assign AAD Group to Enterprise App


ausfestivus:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v1.2.1
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.22.0
```

### Affected Resource(s)

* `data.azuread_application`
* `data.azuread_service_principal`
* `resource.azuread_app_role_assignment`
* `resource.azuread_group`

### Terraform Configuration Files

```hcl
data "azuread_application" "tfc-application" {
  display_name = local.aad_application_name[var.environment]
}

data "azuread_service_principal" "tfc-application" {
  display_name = local.aad_application_name[var.environment]
}

resource "azuread_app_role_assignment" "name" {
  app_role_id         = data.azuread_service_principal.tfc-application.app_role_ids
  principal_object_id = azuread_group.cpt-aad-admins.object_id
  resource_object_id  = data.azuread_service_principal.tfc-application.object_id
}
```

### Debug Output

Trimmed output of the `data.azuread_service_principal.tfc-application` data source.
You can see that the `app_role_ids` attribute is an empty object.
```
      + account_enabled               = true
      + alternative_names             = []
      + app_role_assignment_required  = true
      + app_role_ids                  = {}
      + app_roles                     = [
          + {
              + allowed_member_types = [
                  + "User",
                ]
              + description          = "msiam_access"
              + display_name         = "msiam_access"
              + enabled              = true
              + id                   = "GUID"
              + value                = ""
            },
        ]
      + application_id                = "GUID"
```

### Panic Output

`nil`

### Expected Behavior

Terraform should add the group to the Enterprise Application so the group members can access the app.

### Actual Behavior

```
 Error: Invalid index
│ 
│   on tenants_cpt.tf line 65, in resource "azuread_app_role_assignment" "name":
│   65:   app_role_id         = data.azuread_service_principal.tfc-application.app_role_ids[0]
│     ├────────────────
[...]

PR #828, opened 2022-06-24
Example documents based on #813

ausfestivus:
Per my promise in #813 here is an example of how to assign an AAD Group to an Application


examples/assign-group-to-application/main.tf
```diff
@@ -0,0 +1,56 @@
...
+# https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/application_password
+resource "azuread_application_password" "example" {
+  application_object_id = azuread_application.example.object_id
+  end_date_relative     = "4320h" # expire in 6 months
+}
+
+# Assign our AAD group access to the App
+# https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/app_role_assignment
+resource "azuread_app_role_assignment" "tenant-readers" {
+  app_role_id         = azuread_service_principal.example.app_roles[local.msiam_access_index].id
```
manicminer, line 53: I have a couple of questions about the app role - firstly that it doesn't seem to be defined in the example, and secondly it should be possible to look it up by its name using some Terraform builtin functions, rather than using a hardcoded index (which will fail if more roles are added).

manicminer (commented):
Hi @ausfestivus, many thanks for this contribution! This would be a great addition to the examples - I just have a comment regarding the actual app role. Thanks!

manicminer:
Closing this one out due to inactivity. If you'd like to continue work on this PR, please feel free to comment and we'll reopen it.

PR #830, opened 2022-06-26
Add support for automatic provisioning using synchronization jobs

### LINKED ISSUE #744: Support identity synchronization/provisioning
### Community Note

### Description

We would like to configure [AzureAD Synchronization](https://docs.microsoft.com/en-us/graph/api/resources/synchronization-overview?view=graph-rest-beta) via Terraform. Adding the necessary secret and creating a job would already be enough for our use case. 
I would be happy to work in this if the idea looks good :slightly_smiling_face: 

* azuread_synchronization_job
* azuread_synchronization_secret

This could potentially be extended with data sources for templates and schemas, but I think those two resources should already provide a good basis.

```hcl
resource "azuread_synchronization_secret" "this" {
  service_principal_id = azuread_service_principal.this.id

  credentials {
    key   = "BaseAddress"
    value = "user@domain.com"
  }
  credentials {
    key   = "SecretToken"
    value = "password-value"
  }
}

resource "azuread_synchronization_job" "this" {
  service_principal_id = azuread_service_principal.this.id
  template_id          = "BoxOutDelta"
  state                = "paused" 
}
```

#### Graph API Endpoints:
* [Synchronization](https://docs.microsoft.com/en-us/graph/api/resources/synchronization-overview?view=graph-rest-beta)
* [Add Secret](https://docs.microsoft.com/en-us/graph/api/synchronization-synchronization-secrets?view=graph-rest-beta)
* [Job](https://docs.microsoft.com/en-us/graph/api/resources/synchronization-synchronizationjob?view=graph-rest-beta)

JoostvDoorn:
Fixes #744. This adds support for the synchronization job endpoints documented [here](https://docs.microsoft.com/en-us/graph/api/resources/synchronization-overview?view=graph-rest-beta).

A partially working example for databricks SCIM provisioning (you'll need to add app roles and assign users/group to app roles to make it work 100%):

```terraform
terraform {
  required_providers {
    databricks = {
      source = "databricks/databricks"
    }
  }
}

provider "azuread" {

}
provider "azurerm" {
  features {
  }
}
provider "databricks" {
  host = azurerm_databricks_workspace.test.workspace_url
}

resource "azurerm_resource_group" "test" {
  name     = "databricks-test"
  location = "westeurope"
}

data "azuread_application_template" "test" {
  display_name = "Azure Databricks SCIM Provisioning Connector"
}

resource "azurerm_databricks_workspace" "test" {
  name                = "databricks-test"
  location            = azurerm_resource_group.test.location
  resource_group_name = azurerm_resource_group.test.name
  sku                 = "trial"

  managed_resource_group_name = "databricks-test-rg"
}

resource "databricks_token" "scim" {
  comment = "SCIM Integration"
}

  display_name = "test-databricks"
  template_id  = data.azuread_application_template.test.template_id
  feature_tags {
    enterprise = true
    gallery    = true
  }
}

resource "azuread_service_principal" "test" {
  application_id = azuread_application.test.application_id
  use_existing   = true
}

resource "azuread_synchronization_secret" "test" {
  service_principal_id = azuread_service_principal.test.id

  credentials {
    key   = "BaseAddress"
    value = "https://${azurerm_databricks_workspace.test.workspace_url}/api/2.0/preview/scim"
  }
  credentials {
    key   = "SecretToken"
    value = databricks_token.scim.token_value
  }
}

resource "azuread_synchronization_job" "test" {
  service_principal_id = azuread_service_principal.test.id
  template_id          = "dataBricks"
  state                = "Active"
}
```

### DIFF
--- a/docs/resources/synchronization_job.md
+++ b/docs/resources/synchronization_job.md
@@ -1,0 +1,90 @@
+---
+subcategory: "Synchronization"
+---
+
+# Resource: azuread_synchronization_job
+
+Manages a synchronization job associated with a service principal (enterprise application) within Azure Active Directory.
+
+## API Permissions
+
+The following API permissions are required in order to use this resource.
+
+When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
+
+## Example Usage
+
+*Basic example*
+
+```terraform
+data "azuread_application_template" "example" {
+  display_name = "Azure Databricks SCIM Provisioning Connector"
+}
+
+resource "azuread_application" "example" {
+  display_name = "example"
+  template_id  = data.azuread_application_template.example.template_id
+  feature_tags {
+    enterprise = true
+    gallery    = true
+  }
+}
+
+resource "azuread_service_principal" "example" {
+  application_id = azuread_application.example.application_id
+  use_existing   = true
+}
+
+resource "azuread_synchronization_secret" "example" {
+  service_principal_id = azuread_service_principal.example.id
+
+  credential {
+    key   = "BaseAddress"
+    value = "https://adb-example.azuredatabricks.net/api/2.0/preview/scim"
+  }
+  credential {
+    key   = "SecretToken"
+    value = "some-token"
+  }
+}
+
+resource "azuread_synchronization_job" "example" {
+  service_principal_id = azuread_service_principal.example.id
+  template_id          = "dataBricks"
+  enabled              = true
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `enabled` - (Optional) Whether or not the provisioning job is enabled. Default state is `true`.
+* `service_principal_id` - (Required) The object ID of the service principal for which this synchronization job should be created. Changing this field forces a new resource to be created.
+* `template_id` - (Required) Identifier of the synchronization template this job is based on.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - An ID used to uniquely identify this synchronization job.
+* `schedule` - A `schedule` list as documented below.
+
+---
+
+`schedule` block exports the following attributes:
+
+* `expiration` - Date and time when this job will expire, formatted as an RFC3339 date string (e.g. `2018-01-01T01:02:03Z`).
+* `interval` - The interval between synchronization iterations ISO8601. E.g. PT40M run every 40 minutes.
+* `state` - State of the job.
+
+## Import
+
+Synchronization jobs can be imported using the `id`, e.g.
+
+```shell
+terraform import azuread_synchronization_job.example 00000000-0000-0000-0000-000000000000/job/dataBricks.f5532fc709734b1a90e8a1fa9fd03a82.8442fd39-2183-419c-8732-74b6ce866bd5
+```
+
+-> This ID format is unique to Terraform and is composed of the Service Principal Object ID and the ID of the Synchronization Job Id in the format `{servicePrincipalId}/job/{jobId}`.
--- a/docs/resources/synchronization_secret.md
+++ b/docs/resources/synchronization_secret.md
@@ -1,0 +1,75 @@
+---
+subcategory: "Synchronization"
+---
+
+# Resource: azuread_synchronization_secret
+
+Manages synchronization secrets associated with a service principal (enterprise application) within Azure Active Directory.
+
+## API Permissions
+
+The following API permissions are required in order to use this resource.
+
+When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
+
+## Example Usage
+
+*Basic example*
+
+```terraform
+data "azuread_application_template" "example" {
+  display_name = "Azure Databricks SCIM Provisioning Connector"
+}
+
+resource "azuread_application" "example" {
+  display_name = "example"
+  template_id  = data.azuread_application_template.example.template_id
+  feature_tags {
+    enterprise = true
+    gallery    = true
+  }
+}
+
+resource "azuread_service_principal" "example" {
+  application_id = azuread_application.example.application_id
+  use_existing   = true
+}
+
+resource "azuread_synchronization_secret" "example" {
+  service_principal_id = azuread_service_principal.example.id
+
+  credential {
+    key   = "BaseAddress"
+    value = "abc"
+  }
+  credential {
+    key   = "SecretToken"
+    value = "some-token"
+  }
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `credential` - (Optional) One or more `credential` blocks as documented below.
+* `service_principal_id` - (Required) The object ID of the service principal for which this synchronization secrets should be stored. Changing this field forces a new resource to be created.
+
+---
+
+`credential` block supports the following:
+
+* `key` - (Required) The key of the secret.
+* `value` - (Required) The value of the secret.
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - An ID used to uniquely identify this synchronization sec.
+
+## Import
+
+This resource does not support importing.
--- a/internal/services/serviceprincipals/synchronization_job_resource.go
+++ b/internal/services/serviceprincipals/synchronization_job_resource.go
@@ -1,0 +1,243 @@
+package serviceprincipals
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/manicminer/hamilton/msgraph"
+	"github.com/manicminer/hamilton/odata"
+
+	"github.com/hashicorp/terraform-provider-azuread/internal/clients"
+	"github.com/hashicorp/terraform-provider-azuread/internal/helpers"
+	"github.com/hashicorp/terraform-provider-azuread/internal/services/serviceprincipals/parse"
+	"github.com/hashicorp/terraform-provider-azuread/internal/tf"
+	"github.com/hashicorp/terraform-provider-azuread/internal/utils"
+	"github.com/hashicorp/terraform-provider-azuread/internal/validate"
+)
+
+func synchronizationJobResource() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: synchronizationJobResourceCreate,
+		ReadContext:   synchronizationJobResourceRead,
+		UpdateContext: synchronizationJobResourceUpdate,
+		DeleteContext: synchronizationJobResourceDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(15 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Importer: tf.ValidateResourceIDPriorToImport(func(id string) error {
+			_, err := parse.SynchronizationJobID(id)
+			return err
+		}),
+
+		SchemaVersion: 0,
+
+		Schema: map[string]*schema.Schema{
+			"service_principal_id": {
+				Description:      "The object ID of the service principal for which this synchronization job should be created",
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validate.UUID,
+			},
+			"template_id": {
+				Description: "Identifier of the synchronization template this job is based on.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			"enabled": {
+				Description: "Whether or not the synchronization job is enabled",
+				Type:        schema.TypeBool,
+				Default:     true,
+				Optional:    true,
+			},
+			"schedule": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"expiration": {
+							Description: "Date and time when this job will expire, formatted as an RFC3339 date string (e.g. `2018-01-01T01:02:03Z`).",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"interval": {
+							Description: "The interval between synchronization iterations ISO8601. E.g. PT40M run every 40 minutes.",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"state": {
+							Description: "State.",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func synchronizationJobResourceCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client).ServicePrincipals.SynchronizationJobClient
+	spClient := meta.(*clients.Client).ServicePrincipals.ServicePrincipalsClient
+	objectId := d.Get("service_principal_id").(string)
+
+	tf.LockByName(servicePrincipalResourceName, objectId)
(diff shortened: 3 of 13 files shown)


JoostvDoorn:
The PR should be ready for review now. I ran into a typing issue after some further testing which I solved in the vendored library https://github.com/manicminer/hamilton/pull/172

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		SchemaVersion: 0,
+
+		Schema: map[string]*schema.Schema{
+			"service_principal_id": {
+				Description:      "The id of the application's service principal for which this synchronization job should be created",
```
manicminer, line 44:
```suggestion
				Description:      "The object ID of the service principal for which this synchronization job should be created",
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validate.UUID,
+			},
+			"id": {
+				Description: "Unique synchronization job identifier.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
```
manicminer, line 54: The `id` property should not be declared as it is implicit

```suggestion
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+			"state": {
+				Description: "Desired state.",
+				Type:        schema.TypeString,
+				Default:     "Active",
+				Optional:    true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"Active",
+					"Disabled",
+				}, false),
+			},
```
manicminer, line 70: Can we make this a boolean instead?

```suggestion
			"enabled": {
				Description: "Whether or not the synchronization job is enabled",
				Type:        schema.TypeBool,
				Default:     true,
				Optional:    true,
			},
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+		Timeout:                   time.Until(timeout),
+		MinTimeout:                1 * time.Second,
+		ContinuousTargetOccurence: 5,
+		Refresh: func() (interface{}, string, error) {
+			_, status, err := client.Get(ctx, id.JobId, id.ServicePrincipalId)
+			if err != nil {
+				if status == http.StatusNotFound {
+					return "stub", "Waiting", nil
+				}
+				return nil, "Error", fmt.Errorf("retrieving member")
```
manicminer, line 149:
```suggestion
				return nil, "Error", fmt.Errorf("retrieving synchronization job")
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+	d.SetId(id.String())
+
+	// Start job if desired
+	desiredState := d.Get("state").(string)
+	if desiredState == "Active" {
+		_, err := client.Start(ctx, id.JobId, id.ServicePrincipalId)
+		if err != nil {
+			return tf.ErrorDiagF(err, "Starting synchronization job %q", id.JobId)
+		}
+	}
```
manicminer, line 168: Maybe something like this is we are to change this to a boolean?
```suggestion
	if d.Get("enabled").(bool) {
		_, err := client.Start(ctx, id.JobId, id.ServicePrincipalId)
		if err != nil {
			return tf.ErrorDiagF(err, "Starting synchronization job %q", id.JobId)
		}
	}
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+	job, status, err := client.Get(ctx, id.JobId, id.ServicePrincipalId)
+	if err != nil {
+		if status == http.StatusNotFound {
+			log.Printf("[DEBUG] Synchronization job with ID %q for service principal %q was not found - removing from state!", id.JobId, id.ServicePrincipalId)
+			d.SetId("")
+			return nil
+		}
+		return tf.ErrorDiagF(err, "Retrieving synchronization job with object ID %q", id.JobId)
+	}
+	tf.Set(d, "id", job.ID)
```
manicminer, line 190: This should not be set, either manually or at all in the read func

```suggestion
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+	tf.Set(d, "schedule", flattenSynchronizationSchedule(job.Schedule))
+	tf.Set(d, "template_id", job.TemplateId)
+	tf.Set(d, "state", job.Schedule.State)
```
manicminer, line 193: Looks like this should be in the `schedule` block?
```suggestion
```

internal/services/serviceprincipals/synchronization_job_resource.go
```diff
@@ -0,0 +1,251 @@
...
+				return tf.ErrorDiagF(err, "Starting synchronization job %q", id.JobId)
+			}
+		}
+		if desiredState == "Disabled" {
+			_, err := client.Pause(ctx, id.JobId, id.ServicePrincipalId)
+			if err != nil {
+				return tf.ErrorDiagF(err, "Pausing synchronization job %q", id.JobId)
+			}
+		}
+	}
```
manicminer, line 217: Can this be reworked as per a boolean property `enabled`? Also, whilst it's legal to call `d.Get()` with nested properties, we typically prefer to roll container properties (blocks, lists, sets etc) into flatten/expand functions.

internal/services/serviceprincipals/synchronization_secret_resource.go
```diff
@@ -0,0 +1,217 @@
...
+			Read:   schema.DefaultTimeout(4 * time.Minute),
+			Update: schema.DefaultTimeout(3 * time.Minute),
+			Delete: schema.DefaultTimeout(3 * time.Minute),
+		},
+
+		SchemaVersion: 0,
+
+		Schema: map[string]*schema.Schema{
+			"service_principal_id": {
+				Description:      "The id of the application's service principal for which this synchronization secret should be created",
```
manicminer, line 43:
```suggestion
				Description:      "The object ID of the service principal for which this synchronization secret should be created",
```

internal/services/serviceprincipals/synchronization_secret_resource.go
```diff
@@ -0,0 +1,217 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validate.UUID,
+			},
+			"id": {
+				Description: "Unique synchronization secret identifier.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
```
manicminer, line 53: As per the other resource, the `id` property is implicit and shouldn't be defined in the schema.

```suggestion
```

internal/services/serviceprincipals/synchronization_secret_resource.go
```diff
@@ -0,0 +1,217 @@
...
+			"credentials": {
```
manicminer, line 54: As this is intended to be used as a block, we prefer to use singular rather than plural naming

```suggestion
			"credential": {
```

internal/services/serviceprincipals/synchronization_secret_resource.go
```diff
@@ -0,0 +1,217 @@
...
+	_, err = (&resource.StateChangeConf{
+		Pending:                   []string{"Waiting"},
+		Target:                    []string{"Done"},
+		Timeout:                   time.Until(timeout),
+		MinTimeout:                1 * time.Second,
+		ContinuousTargetOccurence: 5,
+		Refresh: func() (interface{}, string, error) {
+			newSynchronizationSecret, _, err := client.GetSecrets(ctx, id.ServicePrincipalId)
+			if err != nil {
+				return nil, "Error", fmt.Errorf("retrieving member")
```
manicminer, line 117:
```suggestion
				return nil, "Error", fmt.Errorf("retrieving synchronization secret")
```

internal/services/serviceprincipals/synchronization_secret_resource.go
```diff
@@ -0,0 +1,217 @@
...
+				return "stub", "Waiting", nil
+			}
+		},
+	}).WaitForStateContext(ctx)
+
+	if err != nil {
+		return tf.ErrorDiagF(err, "Waiting for synchronization secret %q", id.ServicePrincipalId)
+	}
+
+	d.SetId(id.String())
```
manicminer, line 134: Since this is used for updating too, this should only be set on creation
```suggestion
	if d.IsNewResource() {
		d.SetId(id.String())
	}
```

docs/resources/synchronization_job.md
```diff
@@ -0,0 +1,84 @@
+---
+subcategory: "Synchronization"
+---
+
+# Resource: azuread_synchronization_job
+
+Manages a synchronization job associated with an enterprise application service principal within Azure Active Directory.
```
manicminer, line 7:
```suggestion
Manages a synchronization job associated with a service principal (enterprise application) within Azure Active Directory. 
```

docs/resources/synchronization_job.md
```diff
@@ -0,0 +1,84 @@
...
+  credentials {
+    key   = "SecretToken"
+    value = "some-token"
+  }
+}
+
+resource "azuread_synchronization_job" "example" {
+  service_principal_id = azuread_service_principal.example.id
+  template_id          = "dataBricks"
+  state                = "Active"
```
manicminer, line 54: Note to update this with a boolean `enabled` property

docs/resources/synchronization_job.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `service_principal_id` - (Required) The object ID of the service principal for which this synchronization job should be created. Changing this field forces a new resource to be created.
+* `state` - (Optional) Desired state of the provisioning job, 'Active' or 'Disabled'. Default state is Active.
```
manicminer, line 64: Also here

docs/resources/synchronization_job.md
```diff
@@ -0,0 +1,84 @@
...
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - An ID used to uniquely identify this synchronization job.
+* `schedule` - A `schedule` list as documented below.
+
+---
+
+`schedule` is a list of objects with the following attributes:
```
manicminer, line 76:
```suggestion
`schedule` block exports the following attributes:
```

docs/resources/synchronization_secret.md
```diff
@@ -0,0 +1,75 @@
+---
+subcategory: "Synchronization"
+---
+
+# Resource: azuread_synchronization_secret
+
+Manages synchronization secrets associated with an enterprise application service principal within Azure Active Directory.
```
manicminer, line 7:
```suggestion
Manages synchronization secrets associated with a service principal (enterprise application) within Azure Active Directory. 
```

docs/resources/synchronization_secret.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Resource: azuread_synchronization_secret
+
+Manages synchronization secrets associated with an enterprise application service principal within Azure Active Directory. 
+
+## API Permissions
+
+The following API permissions are required in order to use this resource.
+
+When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.OwnedBy` or `Directory.ReadWrite.All`
```
manicminer, line 13:
```suggestion
When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
```

docs/resources/synchronization_job.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Resource: azuread_synchronization_job
+
+Manages a synchronization job associated with an enterprise application service principal within Azure Active Directory. 
+
+## API Permissions
+
+The following API permissions are required in order to use this resource.
+
+When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.OwnedBy` or `Directory.ReadWrite.All`
```
manicminer, line 13:
```suggestion
When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
```

docs/resources/synchronization_job.md
```diff
@@ -0,0 +1,84 @@
...
+
+`schedule` is a list of objects with the following attributes:
+
+* `expiration` - Date and time when this job will expire, formatted as an RFC3339 date string (e.g. `2018-01-01T01:02:03Z`).
+* `interval` - The interval between synchronization iterations ISO8601. E.g. PT40M run every 40 minutes.
+* `state` - State of the job.
+
+## Import
+
+This resource does not support importing.
```
manicminer, line 84: We should be able to import this resource?

docs/resources/synchronization_secret.md
```diff
@@ -0,0 +1,75 @@
...
+  service_principal_id = azuread_service_principal.example.id
+
+  credentials {
+    key   = "BaseAddress"
+    value = "abc"
+  }
+  credentials {
+    key   = "SecretToken"
+    value = "some-token"
+  }
```
manicminer, line 48: Per above note on pluralisation

```suggestion
  credential {
    key   = "BaseAddress"
    value = "abc"
  }
  credential {
    key   = "SecretToken"
    value = "some-token"
  }
```

docs/resources/synchronization_secret.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `credentials` - (Optional) A collection of `credentials` blocks as documented below.
```
manicminer, line 57: Also here

```suggestion
* `credential` - (Optional) One or more `credential` blocks as documented below.
```

manicminer (changes requested):
@JoostvDoorn Thanks for this contribution, and for your patience with the delay on reviewing this. This is looking really great, I have some comments below which are largely related to each other. Your [Hamilton fix](https://github.com/manicminer/hamilton/pull/172) was merged earlier today and this should be vendored into main fairly soon.

If you could take a look at the suggested changes, and also if we can add an acceptance test for a disabled synchronization job, then this should be good to merge. Thanks again!

internal/services/serviceprincipals/synchronization_job_resource.go
JoostvDoorn: I have changed it to use the boolean logic as you suggested.

internal/services/serviceprincipals/synchronization_job_resource.go
JoostvDoorn: I resolved this by using `d.HasChange("enabled")` since the state is already set to "enabled" in the read call.

docs/resources/synchronization_job.md
JoostvDoorn: Good point. I added the import functionality.

JoostvDoorn:
@manicminer I refactored this over the weekend (and squashed the history). This should be good to go, let me know if there is still some issues.

manicminer (approved):
Thanks for making those changes @JoostvDoorn.

I see that you have changed the `owners` property for both `azuread_application` and `azuread_service_principal` to be optional & computed. Whilst there are some bugs related to these properties, this is not our preferred way to dismiss an unwanted diff and I am hoping to be able to fix some of these bugs separately. Accordingly, I've reverted these additions but this PR otherwise LGTM and I'm happy to merge this. Thanks again!

akansha23191:
Hi, I have used the above example to implement SCIM configuration for Syncing AAD users with Databricks Workspace but am not able to pass the value of Databricks URL programmatically. The above syntax throws below error:

```
Error: Invalid template interpolation value
│
│   on ..\modules\databricks\main.tf line 84, in resource "azuread_synchronization_secret" "test":
│   84:     value = "https://${azurerm_databricks_workspace.dbricks.workspace_url}/api/2.0/preview/scim"
│     ├────────────────
│     │ azurerm_databricks_workspace.dbricks.workspace_url is null
│
│ The expression result is null. Cannot include a null value in a string template.
```


Could you please help me understand why it might be appearing?

Issue #787, opened 2022-04-28
Updating existing AD users with attributes that are null on the Graph API side results in a PATCH with a false value


ghost:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

* `Terraform v1.1.9 on darwin_amd64`
* `hashicorp/azuread => 2.21.0`

### Terraform Configuration Files

```hcl
resource "azuread_user" "user" {
  user_principal_name   = "${var.username}@${var.domain_name}"
  display_name          = var.display_name
  password              = var.password
  force_password_change = true
  company_name          = var.company_name
  account_enabled       = var.account_enabled
  mail                  = var.mail
  department            = var.department
  given_name            = var.given_name
  job_title             = var.job_title
  surname               = var.surname
  usage_location        = var.usage_location
  other_mails           = var.other_mails
  preferred_language    = "en-GB"

  lifecycle {
    ignore_changes = [
      password,
      force_password_change,
      preferred_language,
    ]
  }
}

```

### Debug Output

```
Error: Could not update user with ID: "XXXXXXX"

  with module.user["XXXXXXX"].azuread_user.user,
  on ../modules/user/main.tf line 1, in resource "azuread_user" "user":
   1: resource "azuread_user" "user" {

UsersClient.BaseClient.Patch(): unexpected status 400 with OData error:
Request_BadRequest: Unable to update the specified properties for objects
that have originated within an external service.
```

### Expected Behavior

Existing users in AD that have attributes with 'null' on the Graph API side should have those attributes ignored when we want to update other, non-null attributes.

### Actual Behavior

We observe that null valued attributes are unmarshalled to False values, which results in some API errors on resubmission, with the actual amended attributes.

### Steps to Reproduce

1. Find a user that originates from an Exchange Directory, not created within Azure AD directly.
2. Update an attribute such as email address or job title in the Terraform config
3. `terraform apply`
4. Observe an update error

### Important Factoids

[...]

PR #831, opened 2022-06-27
Bugfix: Only set `show_in_address_list` when changed in config as it is a potentially read-only attribute

manicminer:
Closes: #787


katbyte (approved):
LGTM ❄️

PR #837, opened 2022-07-08
Bugfix: fix a bug with scoped role assignments and fix incorrect schema validation

manicminer:
Also rename these properties, with deprecation path:
- `app_scope_object_id` to `app_scope_id`
- `directory_scope_object_id to `directory_scope_id`

### DIFF
--- a/internal/services/directoryroles/directory_role_assignment_resource.go
+++ b/internal/services/directoryroles/directory_role_assignment_resource.go
@@ -13,7 +13,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-azuread/internal/clients"
 	"github.com/hashicorp/terraform-provider-azuread/internal/tf"
-	"github.com/hashicorp/terraform-provider-azuread/internal/utils"
 	"github.com/hashicorp/terraform-provider-azuread/internal/validate"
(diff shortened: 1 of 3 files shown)


internal/services/directoryroles/directory_role_assignment_resource.go
```diff
@@ -43,35 +42,58 @@ func directoryRoleAssignmentResource() *schema.Resource {
 			"role_id": {
 				Description:      "The object ID of the directory role for this assignment",
 				Type:             schema.TypeString,
-				Optional:         true,
+				Required:         true,
```
manicminer, line 45: This is documented as required, it being optional is a bug.

internal/services/directoryroles/directory_role_assignment_resource.go
```diff
@@ -43,35 +42,58 @@ func directoryRoleAssignmentResource() *schema.Resource {
...
 				ForceNew:         true,
 				ValidateDiagFunc: validate.UUID,
 			},
 
 			"principal_object_id": {
 				Description:      "The object ID of the member principal",
 				Type:             schema.TypeString,
-				Optional:         true,
+				Required:         true,
```
manicminer, line 53: This is documented as required, it being optional is a bug.

katbyte (approved):
LGTM 🦀

PR #838, opened 2022-07-08
azuread_group: fix a bug where new group creation can error out before the timeout due to API inconsistency

manicminer:


jackofallops (approved):
LGTM 👍

Issue #832, opened 2022-06-28
data azuread_group -> More than one group found matching specified filter (displayName eq '<DisplayName>' and securityEnabled eq true)


BrianRichardson:
I am attempting to retrieve data for one of two Azure AD groups:

data "azuread_group" "group" {
  display_name = "\<DisplayName\>"
  security_enabled = true
  mail_enabled = false
}

Two AD groups exist with the same name where the difference being one is of type "Security" and the other of type "Mail enabled security".

![image](https://user-images.githubusercontent.com/12933801/176142998-498587b2-d1da-4d39-be2d-73426a4ce04e.png)

"terraform plan" command results:
More than one group found matching specified filter (displayName eq '<DisplayName>' and securityEnabled eq true)

Instead I expect the group without mail enabled to be retrieved. Is my expectation incorrect?
[Related functionality](https://github.com/hashicorp/terraform-provider-azuread/pull/393)

Terraform v1.2.3
on windows_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.25.0
+ provider registry.terraform.io/hashicorp/azurerm v3.11.0

PR #841, opened 2022-07-11
data.azuread_group: ensure security/mail enabled groups are excluded when explicitly `false` in config

manicminer:
Fixes: #832


jackofallops (approved):
LGTM 👍

Issue #848, opened 2022-07-18
Changing display_name parameter value during second terraform apply results in force replacement of the resource in azuread_application_password


ccsandhanshive:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
terraform 1.0.0
azuread 2.20.0

### Affected Resource(s)

* `azuread_application_password`

### Debug Output
```
$ terraform apply
*************.this["***********"]: Refreshing state... [id=*********************/password/a74dd7fc-4a83-4ef1-87a2-02a04dda37f7]

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # *************.this["*********"] must be replaced
-/+ resource "azuread_application_password" "this" {
      + display_name          = "trial" # forces replacement
      ~ end_date              = "2024-07-18T15:50:21Z" -> (known after apply)
      ~ id                    = "*******************/password/a74dd7fc-4a83-4ef1-87a2-02a04dda37f7" -> (known after apply)
      ~ key_id                = "a74dd7fc-4a83-4ef1-87a2-02a04dda37f7" -> (known after apply)
      ~ start_date            = "2022-07-18T15:50:21Z" -> (known after apply)
      ~ value                 = (sensitive value)
        # (1 unchanged attribute hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.

Do you want to perform these actions in workspace "**********"?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

********************.this["***************"]: Destroying... [id=****************/password/a74dd7fc-4a83-4ef1-87a2-02a04dda37f7]
********************.this["*************"]: Still destroying... [id=******************/pa...d/a74dd7fc-4a83-4ef1-87a2-02a04dda37f7, 10s elapsed]
********************.this["****************"]: Still destroying... [id=****************/pa...d/a74dd7fc-4a83-4ef1-87a2-02a04dda37f7, 20s elapsed]
[...]

PR #849, opened 2022-07-21
Update application_password.md to reflect 'recreate resource' for display_name field.

Threpio:
As per issue: https://github.com/hashicorp/terraform-provider-azuread/issues/848


manicminer (approved):
Thanks for this docs fix @Threpio, this LGTM! 🖨️

PR #858, opened 2022-08-05
Docs: azuread_group.dynamic_membership.rule is required, not optional

### LINKED ISSUE #857: In azuread_group rule parameter of 'dynamic membership' block is required but in terraform documentation it is marked as 'Optional'
**Terraform (and AzureAD Provider) Version**

- Terraform v1.0.0
- Azuread Plugin Version v2.20.0

**Affected Resource**

- azuread_group

**Expected Behavior**

- azuread_group resource should get successfully created even if rule parameter from 'Dynamic membership' block is not specified.

**Actual Behavior**

- terraform throws following error if 'rule' parameter is not specified -

           
image

        
- Seems like terraform documentation provides a wrong information 'rule' parameter is 'Required' and this should be updated in future releases.

**Steps to Reproduce**

- terraform plan
- terraform apply

manicminer:
Fixes #857

### DIFF
--- a/docs/resources/group.md
+++ b/docs/resources/group.md
@@ -154,7 +154,7 @@ The following arguments are supported:
 `dynamic_membership` block supports the following:
 
 * `enabled` - (Required) Whether rule processing is "On" (true) or "Paused" (false).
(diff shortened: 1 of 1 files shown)


katbyte (approved):
LGTM 🎩

PR #864, opened 2022-08-17
Add Support for optional additional property

MarkDordoy:
Adding `max_size_limit` as an optional property. When this is used, it allows the groups added to the token or SAML response to go from 150 to a max of 1000.


internal/services/applications/migrations/application_resource.go
```diff
@@ -328,6 +328,7 @@ func ResourceApplicationInstanceResourceV0() *schema.Resource {
 													"emit_as_roles",
 													"include_externally_authenticated_upn",
 													"include_externally_authenticated_upn_without_hash",
+												        "max_size_limit",
 													"netbios_domain_and_sam_account_name",
```
manicminer, line 332: tabs vs spaces :)

```suggestion
													"include_externally_authenticated_upn_without_hash",
													"max_size_limit",
													"netbios_domain_and_sam_account_name",
```

katbyte (approved):
LGTM 🌳

internal/services/applications/migrations/application_resource.go
```diff
@@ -374,6 +375,7 @@ func ResourceApplicationInstanceResourceV0() *schema.Resource {
 													"emit_as_roles",
 													"include_externally_authenticated_upn",
 													"include_externally_authenticated_upn_without_hash",
+													"max_size_limit",
```
manicminer, line 378: On a second pass, we don't need to add this here as this should be a representative snapshot of the schema available at that time.
```suggestion
```

internal/services/applications/migrations/application_resource.go
```diff
@@ -328,6 +328,7 @@ func ResourceApplicationInstanceResourceV0() *schema.Resource {
...
```
manicminer, line 331: Same here
```suggestion
```

Issue #869, opened 2022-08-23
Support generic OIDC implementation via environment variable


samgladstone:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As per the [issue](https://github.com/hashicorp/terraform-provider-azurerm/issues/16901) in [terraform-provider-azurerm](https://github.com/hashicorp/terraform-provider-azurerm), the OIDC implementation is only supported when run in GitHub Actions. 

It would be great to be able to supply a token and so be able to use OIDC in other places (Kubernetes, GitLab, etc)

samgladstone:
I've raised a PR #870 to add a warning that this is implementation is specific to GitHub Action. This could then be removed once the generic work is done

PR #870, opened 2022-08-23
Add warning that OIDC auth only works in GitHub Actions

samgladstone:
Add a warning that the OIDC implementation is specific to GitHub actions and cannot be used in a generic way

This warning could then be removed once the work in #869 is done


manicminer (approved):
Thanks for this docs note @samgladstone. Whilst I agree this should probably be added, I've adjusted the wording and removed the reference to the GitHub issue.

PR #877, opened 2022-08-31
Docs: explain default values for start_date for certificate resources

### LINKED ISSUE #866: azuread_application_certificate: For certificate, the default value of start_date is the “creation date” of the certificate, whereas for symmetric key, the default value is the current timestamp when TF apply is executed.
### Community Note

Terraform v1.0.0
provider registry.terraform.io/hashicorp/azuread v2.20.0

### Affected Resource(s)

* `azuread_application_certificate`

- For example, if the certificate is issued on **7th July 2022** & if it is valid for 1 year which means that the expiry date(end date) would be 8th July 2023.
- So the terraform configuration is as follows:
  - **start_date** is optional parameter & it is not specified.
  - specified **type** as **AsymmetricX509Cert** because it is a certificate.

```
  display_name = "example"
}

resource "azuread_application_certificate" "example" {
  application_object_id = azuread_application.example.id
  encoding              = "pem"
  end_date              = "2023-07-08T00:00:00Z"
  type                  = "AsymmetricX509Cert"
  value                 = file("cert.pem")
}
```

### Expected Behavior

- For Asymmetric certificate (when type=”AsymmetricX509Cert”):
  - As per above configuration, if start_date is not specified for a certificate then after terraform apply is executed, the start_date assigned must be the start date of the certificate when it was issued (it should not be the current timestamp when terraform apply is executed)
  - The same behaviour is observed when a certificate is uploaded on the Azure portal i.e. the start_date of the certificate is the creation date of the certificate & not the current timestamp when it was uploaded.

- For Symmetric key (when type=”Symmetric”):
  - If the start_date is not specified for a symmetric key then the start date for symmetric key should be the current timestamp when terraform apply is executed.

### Actual Behavior

- For Asymmetric certificate (when type=”AsymmetricX509Cert”):
  - The Actual behaviour matches with the Expected behaviour for the certificate .i.e. when start_date is not specified for a certificate then the start_date assigned must be the creation date of the certificate when it was issued. The similar behaviour is observed on Azure portal.
  - But this behaviour is not as per the terraform documentation, because the documentation states that “when start_date is not specified then the current timestamp is used”.

  Please find the below output of terraform apply executed on **17th Aug 2022**
  image

  **On Azure portal:**
   image

  On the azure portal, we can see that the start date mentioned for the certificate is the creation date of the certificate & not the current timestamp.

- For Symmetric key(when type=”Symmetric”):
  - The Actual behaviour matches with the Expected behaviour for the symmetric key & there is no issue. Terraform documentation is correct for Symmetric key type.

The terraform documentation for start_date parameter has a **valid** statement for **Symmetric key** but for **Asymmetric certificate** it is **invalid** statement. Terraform has not explicitly mentioned the behaviour for both Symmetric & Asymmetric type.

**Can you please confirm if this is a bug or it needs correction in the terraform documentation?**

### Steps to Reproduce

1. `terraform apply`

### References

- Terraform reference
    https://registry.terraform.io/providers/hashicorp/azuread/2.20.0/docs/resources/application_certificate

manicminer:
Closes #866

### DIFF
--- a/docs/resources/application_certificate.md
+++ b/docs/resources/application_certificate.md
@@ -131,7 +131,7 @@ The following arguments are supported:
 ~> One of `end_date` or `end_date_relative` must be set. The maximum allowed duration is determined by Azure AD.
 
 * `key_id` - (Optional) A UUID used to uniquely identify this certificate. If omitted, a random UUID will be automatically generated. Changing this field forces a new resource to be created.
-* `start_date` - (Optional) The start date from which the certificate is valid, formatted as an RFC3339 date string (e.g. `2018-01-01T01:02:03Z`). If this isn't specified, the current date and time are used.  Changing this field forces a new resource to be created.
+* `start_date` - (Optional) The start date from which the certificate is valid, formatted as an RFC3339 date string (e.g. `2018-01-01T01:02:03Z`). If this isn't specified, the value is determined by Azure Active Directory and is usually the start date of the certificate for asymmetric keys, or the current timestamp for symmetric keys. Changing this field forces a new resource to be created.
 * `type` - (Required) The type of key/certificate. Must be one of `AsymmetricX509Cert` or `Symmetric`. Changing this fields forces a new resource to be created.
 * `value` - (Required) The certificate data, which can be PEM encoded, base64 encoded DER or hexadecimal encoded DER. See also the `encoding` argument.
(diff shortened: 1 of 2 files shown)


katbyte (approved):
LGTM 🦠

PR #883, opened 2022-09-08
Docs: Office365 in included_applications and excluded_applications

robertbrandso:
Added Office365 as an option for included_applications.


manicminer (changes requested):
@robertbrandso Thanks for this docs update! Could you also add this note to the `excluded_applications` property (as it looks like you can exclude `Office365`, but not `All`)? Thanks!

robertbrandso:
@manicminer Updated as requested. Good catch. I also tested this for excluded_applications. Confirmed working.

manicminer (approved):

PR #884, opened 2022-09-09
Docs: Add 'AllTrusted' as an option for 'excluded_locations'

robertbrandso:
`AllTrusted` can also be set on `excluded_locations`.

### DIFF
--- a/docs/resources/conditional_access_policy.md
+++ b/docs/resources/conditional_access_policy.md
@@ -129,7 +129,7 @@ The following arguments are supported:
 
 `locations` block supports the following:
 
-* `excluded_locations` - (Optional) A list of location IDs excluded from scope of policy.
+* `excluded_locations` - (Optional) A list of location IDs excluded from scope of policy. Can also be set to `AllTrusted`.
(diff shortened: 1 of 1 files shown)


manicminer (changes requested):
Thanks for this @robertbrandso. As this isn't specifically [documented](https://docs.microsoft.com/en-us/graph/api/resources/conditionalaccesslocations?view=graph-rest-1.0), could you add an acceptance test for this, or ensure coverage for this value in an existing test? Thanks!

robertbrandso:
Hi, @manicminer.

I opened a [pull request](https://github.com/microsoftgraph/microsoft-graph-docs/pull/18441) for the MS Graph docs as well.

Regarding the acceptance test. It looks like this is already covered in one of the tests here: https://github.com/hashicorp/terraform-provider-azuread/blob/d8b9594ce8c53c1aa242035081969dce2043a9a3/internal/services/conditionalaccess/conditional_access_policy_resource_test.go#L336

Is that enough?

manicminer (approved):
Thanks @robertbrandso, sorry I missed that already being in the test config!

Issue #888, opened 2022-09-15
[azuread_application] The 'type' parameter in 'oauth2_permission_scope' block is marked as Required but behaves as Optional


ShaheenK27:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
```
Terraform v1.0.0
+ provider registry.terraform.io/hashicorp/azuread v2.20.0
```

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files
The **oauth2_permission_scope** block is specified without **type**
```hcl
resource "azuread_application" "app-test01" {
  display_name = "app-test01"
  api {
    oauth2_permission_scope {
      id                         = "c8c8c26a-d30e-4135-886f-490ea092e7b8"
      admin_consent_display_name = "Read-only access to Employee records"
      admin_consent_description  = "Allow the application to have read-only access to all Employee data."
      value                      = "Employees.Read.All"
    }
  }
}
```
### Expected Behavior
* The **type** parameter should be marked as **Optional** in the documentation.

### Actual Behavior
* The **type** parameter has been marked as **Required** in the documentation.

* The 'Required' specification **conflicts** with the displayed behavior and the parameter description that **type** defaults to **User**.

**Terraform apply (oauth2_permission_scope block specified without type)**
```hcl
Terraform will perform the following actions:

  # azuread_application.app-test01 will be created
  + resource "azuread_application" "app-test01" {
     ...
      + api {
          + requested_access_token_version = 1
          + oauth2_permission_scope {
              + admin_consent_description  = "Allow the application to have read-only access to all Employee data."
              + admin_consent_display_name = "Read-only access to Employee records"
              + enabled                    = true
              + id                         = "c8c8c26a-d30e-4135-886f-490ea092e7b8"
              + type                       = "User"  # Defaults to User. Parameter isn't specified in the terraform configuration
              + value                      = "Employees.Read.All"
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

### Steps to Reproduce

1. `terraform apply` (without type in oauth2-permission_scope block)

### References
[...]

PR #890, opened 2022-09-21
Fix: [azuread_application] The 'type' parameter in 'oauth2_permission_scope' block is marked as Required but behaves as Optional #888

shijiadong2022:
Fix issue #888


shijiadong2022:
@manicminer  when you are available can you take a look at the change? Thanks.

manicminer (approved):
@shijiadong2022 Many thanks for spotting and correcting this.

PR #908, opened 2022-10-18
Docs/app_role_assignment: add example for role assignment to default app role

audunsolemdal:
I didn't find it intuitive how to assign to the default app role, so I added an example.

### DIFF
--- a/docs/resources/app_role_assignment.md
+++ b/docs/resources/app_role_assignment.md
@@ -147,6 +147,29 @@ resource "azuread_app_role_assignment" "example" {
 }
 ```
 
+*Assign a group to the default app role for an internal application*
+
+```terraform
+resource "azuread_application" "internal" {
+  display_name = "internal"
+}
+
+resource "azuread_service_principal" "internal" {
+  application_id = azuread_application.internal.application_id
+}
+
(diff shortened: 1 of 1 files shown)


manicminer (changes requested):
Thanks for this contribution @audunsolemdal. Would you be able to add a note about the magic default value `00000000-0000-0000-0000-000000000000` to the `app_role_id` property? Thanks!

audunsolemdal:
I was thinking about that, but I am not sure whether this should be as simple comment behind the id
```
app_role_id = "00000000-0000-0000-0000-000000000000" # Default app role
```

Or a note with a reference to Microsoft documentation before the example code starts. What do you think is best?

manicminer:
The example you have put together is great - I don't think it needs a trailing comment given the brevity of the config - but I do think it needs mentioning alongside the property description. Perhaps something like this?

```markdown
`app_role_id` - (Required) The ID of the app role to be assigned, or the default role ID `00000000-0000-0000-0000-000000000000`. Changing this forces a new resource to be created.
```

manicminer (approved):
Thanks again @audunsolemdal, this LGTM! 👍

Issue #920, opened 2022-11-04
Clarify the least privilege permissions needed for azuread_application


mathias-nyman:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The permissions `Application.ReadWrite.All` and `Directory.ReadWrite.All` are extremely high privilege permissions on the Azure tenant level and should not be used unless absolutely unavoidable.

The current documentation states:
> When authenticated with a service principal, this resource requires one of the following application roles: Application.ReadWrite.All or Directory.ReadWrite.All
>
> It may be possible to create applications using this resource with the Application.ReadWrite.OwnedBy application role, provided the principal being used to run Terraform is included in the owners property. ...

The statement that the permission `Application.ReadWrite.OwnedBy` **may** be sufficient is vague and leads engineers to needlessly resort to tenant level write permissions at the first error they face.

There is one common pattern at the moment for which users get 403 with just the `Application.ReadWrite.OwnedBy` permission: if you specify additional owners of an azuread_application you, in addition to `Application.ReadWrite.OwnedBy` also need `User.Read.All`. This should be added to the documentation.

### New or Affected Resource(s)

* azuread_application

### References

Issues about receiving a 403 when using additional `owners` and resolved by adding the `User.Read.All` permission:
https://github.com/hashicorp/terraform-provider-azuread/issues/703
https://github.com/hashicorp/terraform-provider-azuread/issues/853

PR #921, opened 2022-11-04
Clarify the least privilege permissions needed for the azuread_application

mathias-nyman:
Solves https://github.com/hashicorp/terraform-provider-azuread/issues/920


manicminer (approved):
Thanks for this @mathias-nyman, this LGTM

PR #922, opened 2022-11-04
Update service principal permissions requirements

aangelisc:
It is possible now to create an application and service principal by making use of a service principal with the `Application.ReadWrite.OwnedBy` permissions. The documentation should be updated to reflect this.

### DIFF
--- a/docs/resources/service_principal.md
+++ b/docs/resources/service_principal.md
@@ -12,7 +12,7 @@ The following API permissions are required in order to use this resource.
 
 When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
 
-It is possible to manage service principals whilst having only the `Application.ReadWrite.OwnedBy` role granted, however only the service principal itself can be set as the owner.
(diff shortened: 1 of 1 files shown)


docs/resources/service_principal.md
```diff
@@ -12,7 +12,7 @@ The following API permissions are required in order to use this resource.
 
 When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
 
-It is not currently possible to manage service principals whilst having only the `Application.ReadWrite.OwnedBy` role granted.
+It is possible to manage service principals whilst having only the `Application.ReadWrite.OwnedBy` role granted, however only the service principal itself can be set as the owner.
```
manicminer, line 15:
```suggestion
It may be possible to manage service principals whilst having only the `Application.ReadWrite.OwnedBy` role granted, however you must ensure that both the underlying application and the service principal have the Terraform principal as an owner.
```

manicminer (approved):
Thanks for this @aangelisc. Since this has been known to work for awhile now I'm happy to add this to the docs, however I've edited slightly to clarify that both the application and the service principal must have appropriate ownership.

PR #925, opened 2022-11-10
Update conditional_access_policy.md - missing none value for included_applications

arkenfar:
Add missing value to "included_applications" attribute for setting no included applications.  If this value isn't set to "None" Terraform fails with the following error:

Error: Could not create conditional access policy
│ 
│   with module.conditional-access.azuread_conditional_access_policy.block_basic_auth_policy,
│   on ../conditional-access/main.tf line 30, in resource "azuread_conditional_access_policy" "block_basic_auth_policy":
│   30: resource "azuread_conditional_access_policy" "block_basic_auth_policy" {
│ 
│ ConditionalAccessPoliciesClient.BaseClient.Post(): unexpected status 500 with OData error: Internal Server Error: There was an internal server error while processing the request. Error ID: c9b4f7ac-cbe5-45fe-8d4c-16961e3b872f

### DIFF
--- a/docs/resources/conditional_access_policy.md
+++ b/docs/resources/conditional_access_policy.md
@@ -96,7 +96,7 @@ The following arguments are supported:
 `applications` block supports the following:
 
 * `excluded_applications` - (Optional) A list of application IDs explicitly excluded from the policy. Can also be set to `Office365`.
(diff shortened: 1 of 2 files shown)


manicminer (approved):
Thanks @arkenfar! Since this value is undocumented, but appears to work, I've updated one of the acceptance tests to use this value so that we can better catch potential future changes.

manicminer:
Test results

Screen Shot 2022-11-14 at 15 50 56

PR #928, opened 2022-11-16
Prevent trailing slash for identifierUris when no path

MarkDordoy:
Using the Graph API, if you try and patch or create an application that has an `IdentifierUris` without a path and a trailing slash, the api will respond with:
```
ApplicationsClient.BaseClient.Patch(): unexpected status 400 with OData error: IdentifierUrisEndsWithSlash: Application alias 'https://mydomain.com/' value is invalid.
```

This PR Extends the function `validate.IsUriFunc` to have a new argument `allowTrailingSlash` and its set to false for `IdentifierUris` validation.


manicminer (approved):
Thanks for this @MarkDordoy, LGTM! 🥧

PR #930, opened 2022-11-17
Only run acctest-oidc action when tokens are available (i.e. not from forks)

manicminer:


.github/workflows/acctest-oidc.yaml
```diff
@@ -13,28 +13,46 @@ permissions:
...
-      version: ${{ steps.go-version.outputs.version }}
+    permissions:
+      contents: read
+      id-token: write
     steps:
-      - uses: actions/checkout@v2
+      - id: oidc-check
+        run: |
+          set -v 
+          if [ "${ACTIONS_ID_TOKEN_REQUEST_URL}" = "" ]; then
```
tombuildsstuff, line 25: shouldn't this be a double-equals?

.github/workflows/release.yaml
```diff
@@ -37,8 +27,8 @@ jobs:
 
   terraform-provider-release:
     name: 'Terraform Provider Release'
-    needs: [go-version, release-notes]
-    uses: hashicorp/ghaction-terraform-provider-release/.github/workflows/hashicorp.yml@v2
+    needs: [release-notes]
+    uses: manicminer/ghaction-terraform-provider-release/.github/workflows/hashicorp.yml@1dba3fa320b576c92d2206c9514e5f9622730010
```
tombuildsstuff, line 31: is there a PR tracking moving this back?

tombuildsstuff (commented):
Left a couple of comments inline but otherwise 👍

.github/workflows/acctest-oidc.yaml
manicminer: Single bracket and single equals keeps this posix-compliant fwiw, but happy to change if it reads better

.github/workflows/release.yaml
manicminer: Not yet, I'm trying to test this end-to-end first as I'm not sure that blindly passing in blank values will work with the eventual upstream action (it should in theory, but GHA keeps confounding me)

PR #936, opened 2022-11-29
Bug Fix for US Gov cloud current issue

MarkDordoy:
Microsoft US Gov cloud has an issue with the [application Template: instantiate](https://learn.microsoft.com/en-us/graph/api/applicationtemplate-instantiate?view=graph-rest-1.0&tabs=http) API call. They are actively working on fixing the bug, but the fix is not expected until February 2023. 

The bug is intermittent, however when it occurs you can end up with multiple app registrations and Service Principals, all with the same name. The bug under the hood shows itself in the form of a 504 status code (timeout) and the Azure Library (Hamilton) performs a retry. Eventually that 504 call will complete and you end up with Mutiple applications created. This bug does not exist on the beta API. 

This PR will place anyone using the environment `MsGraphUSGovL4Endpoint` onto the beta endpoint just for the application templates client. The response model changes slightly going to beta so this PR also accommodates for this. 

This PR required https://github.com/manicminer/hamilton/pull/198 to be merged and released which has a breaking change regarding the DirectoryObject ID struct field. This has been replaced with a function of `.ID()` which returns the Id regardless of objectID or Id return response from the Graph API


manicminer (approved):
Thanks for working on this @MarkDordoy, this LGTM! 🚀

PR #949, opened 2022-12-13
Add debug support

marcgs:
Add debug mode to support provider debugging as described in the [plugin development documentation](https://developer.hashicorp.com/terraform/plugin/debugging#debugger-based-debugging).

The azurerm provider supports debug mode (see [main.go](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/main.go#L21)) as well. I am not sure if it was intended not to support it in azuread, but to me it would seem reasonable as it might help when developing new features or fixing bugs. The name of the flag to activate/deactivate debug mode is "debug" in contrast to "debuggable", which seemed more intuitive to me.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,20 +6,14 @@ import (
 )
 
 func main() {
-	var debugMode bool
+	var debug bool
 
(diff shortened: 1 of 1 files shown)


manicminer:
Superseded by #1329

PR #952, opened 2022-12-16
Declare additional API IDs to support modern Teams apps

pablodiazgutierrez:
Microsoft requires declaring [access to several APIs](https://learn.microsoft.com/en-us/microsoftteams/platform/m365-apps/extend-m365-teams-personal-tab?tabs=manifest-teams-toolkit#update-azure-ad-app-registration-for-sso) for modern Teams apps to pass validation. In this PR, we add those API Ids to the list of well-known ones, to facilitate deployment of such apps.

These API IDs are taken from Microsoft's own documentation, but this is only a partial import based on those definitions.

### DIFF
--- a/vendor/github.com/manicminer/hamilton/environments/published.go
+++ b/vendor/github.com/manicminer/hamilton/environments/published.go
@@ -5,6 +5,8 @@ type ApiAppId = string
 // PublishedApis is a map containing Application IDs for well known APIs published by Microsoft.
 // They can be used to acquire access tokens, but are primarily described here for easy inclusion in
 // application manifests and service principal assignments.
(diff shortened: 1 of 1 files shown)


manicminer (commented):
@pablodiazgutierrez, thanks for this contribution. Whilst we're happy to add these app IDs, because this file is part of an upstream dependency we cannot merge this and it will need to be changed in the corresponding repository: https://github.com/manicminer/hamilton/blob/main/environments/published.go

Thanks!

manicminer:
Note: these were added upstream in https://github.com/manicminer/hamilton/pull/216

PR #953, opened 2022-12-17
Check for github action updates monthly

jlosito:
This will use dependabot to check for updates against any github action which this project uses and submit pull requests with version bumps in order to keep them up-to-date.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


manicminer:
Thanks for suggesting this @jlosito. We have since adopted some internal tooling ([example](https://github.com/hashicorp/terraform-provider-azuread/pull/1140)) which offers increased functionality over dependabot and so we're not looking to adopt dependabot at this time. Appreciate the suggestion!

Issue #956, opened 2023-01-02
Failed Acceptance Tests on Data Sources Imports


tagur87:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
terraform = 1.3.4
terraform-provider-azuread = v2.31.0

### Affected Resource(s)

Acceptance tests on any resource that has a data source dependency.

### Terraform Configuration Files

N/A

### Debug Output
https://gist.github.com/tagur87/ebea9d00dd22ad4b10204e1f49c0e8fa

### Panic Output
https://gist.github.com/tagur87/ebea9d00dd22ad4b10204e1f49c0e8fa

### Expected Behavior
Acceptance tests should pass.

### Actual Behavior
Error below:
`testcase.go:60: Failed state verification, resource with ID domains#9b4df314-e3af-46e6-b26d-342a0ddb28b0#NDtgcbj_4NQSnOubhRncUVD02M0= not found`

### Steps to Reproduce

1. Run acceptance tests against service principal `make testacc TESTARGS='-run=TestAccServicePrincipal'`

### Important Factoids

The terraform provider SDK is out of date and needs updated to fix the issue.

### References

Error reported here: https://github.com/hashicorp/terraform-provider-aws/issues/27049

Error fixed upstream here: https://github.com/hashicorp/terraform-plugin-sdk/pull/1074

PR #957, opened 2023-01-02
Update terraform-plugin-sdk/v2 -> 2.24.1

tagur87:
This fixes an error with acceptance tests fixed upstream here:

https://github.com/hashicorp/terraform-plugin-sdk/pull/1074

Closes https://github.com/hashicorp/terraform-provider-azuread/issues/956


tagur87:
Main change is bumping sdk version to 2.24.0. After that `go mod tidy` and `go mod vendor` were run. 

@bflad @manicminer - thoughts?

bflad:
Hi @tagur87 👋 While the maintainers of terraform-plugin-sdk generally consider minor version upgrades to be safe for consumers, it is up to consumers to choose if/when it is reasonable to test and merge those upgrades using whatever criteria and schedule is appropriate for the project. I'm not a maintainer on this project but it does appear that the appropriate maintainers are already watchers of the repository so I'll let them handle the PR notification and triage as they normally would.

Issue #823, opened 2022-06-13
Create a SSO SAML Signing Certificate


sseekamp0:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I would like to open a new feature request to enable creation of a SSO SAML Signing Certificate. This functionality currently exists via the API, but seems to be missing in the terraform provider.

[API Reference](https://docs.microsoft.com/en-us/graph/application-saml-sso-configure-api#step-4-configure-a-signing-certificate)

[Azure AD service principle certificate](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/service_principal_certificate) only provides importing a certificate. This request would be for creating a SSO certificate.

### New or Affected Resource(s)

* azuread_service_principal_certificate

### References

* https://github.com/hashicorp/terraform-provider-azuread/issues/173 - there is a reference here to creating a certificate.

rajish:
It is possible to create the certificate, but apparently, [the activation](https://docs.microsoft.com/en-us/graph/application-saml-sso-configure-api#activate-the-custom-signing-key) has to be done manually or using Azure CLI. My code so far:

```
data "azuread_client_config" "current" {}

resource "random_uuid" "oauth2_permission_scope" {
  keepers = {
    domain_name = var.saml_identifier_uri
  }
}

resource "random_uuid" "app_role_user" {
  keepers = {
    domain_name = var.saml_identifier_uri
  }
}

resource "random_uuid" "app_role_msiam_access" {
  keepers = {
    domain_name = var.saml_identifier_uri
  }
}

resource "azuread_application" "this" {
  display_name = var.display_name
  owners       = [data.azuread_client_config.current.object_id]
  identifier_uris = [var.saml_identifier_uri]
  api {
    oauth2_permission_scope {
      id = random_uuid.oauth2_permission_scope.id
      admin_consent_description = "Allow the application to access ${var.display_name} on behalf of the signed-in user."
      admin_consent_display_name = "Access ${var.display_name}"
      user_consent_description = "Allow the application to access ${var.display_name} on behalf of the signed-in user."
[...]

PR #968, opened 2023-01-10
Add token signing certificate resource

### LINKED ISSUE #732: Add azuread_service_principal_token_signing_certificate to simplify certificate management

tagur87:
This adds a new resource called
`service_principal_token_signing_certificate` that is used to manage the whole lifecycle of token signing certificates used for SAML authentication.

This resource makes use of the `AddTokenSigningCertificate` function that was added to hamilton previously here:
https://github.com/manicminer/hamilton/pull/158

MS Graphs Docs: https://learn.microsoft.com/en-us/graph/api/serviceprincipal-addtokensigningcertificate?view=graph-rest-1.0&tabs=http

As documented in the docs above, when the  `AddTokenSigningCertificate` function is invoked, 3 individual objects are created...
- Verify `keyCredential` (Public Cert)
- Sign `keyCredential` (Private Key)
- `passwordCredential` (Private Key Password)

When the object is returned, it includes the thumbprint, the public key pem value, and a `keyId`. However, we found an odd behavior that the `keyId` that is returned is actually for the Sign `keyCredential`.

Since the Verify certificate is the one that we acutally care about, we used the `customKeyIdentifier`, which is the same for all 3 values, to get the Verify `keyId`, which we then use in building the resource ID.

We additionally had to "calculate" the thumbprint value from the actual value of the Verify cert, as this value is not returned from the API, except after initial creation in the Create step. We did this by getting pem value of the Verify cert by adding the `$select=keyCredential` odata query to the GET of the service principal. By combining this value with the PEM header/footer, we can calculate the SHA-1 fingerprint, which matches up to the appropriate thumbprint.

Finally, to delete the certificate, we have to PATCH the service principal with all 3 objects mentioned previously removed. To gather this, we used the `customKeyIdentifier` value in a loop.

Closes #732
And part of https://github.com/hashicorp/terraform-provider-azuread/issues/823

### DIFF
--- a/docs/resources/service_principal_token_signing_certificate.md
+++ b/docs/resources/service_principal_token_signing_certificate.md
@@ -1,0 +1,87 @@
+---
+subcategory: "Service Principals"
+---
+
+# Resource: azuread_service_principal_token_signing_certificate
+
+Manages a token signing certificate associated with a service principal within Azure Active Directory.
+
+## API Permissions
+
+The following API permissions are required in order to use this resource.
+
+When authenticated with a service principal, this resource requires one of the following application roles: `Application.ReadWrite.All` or `Directory.ReadWrite.All`
+
+When authenticated with a user principal, this resource requires one of the following directory roles: `Application Administrator` or `Global Administrator`
+
+## Example Usage
+
+*Using default settings*
+
+```terraform
+resource "azuread_application" "example" {
+  display_name = "example"
+}
+
+resource "azuread_service_principal" "example" {
+  application_id = azuread_application.example.application_id
+}
+
+resource "azuread_service_principal_token_signing_certificate" "example" {
+  service_principal_id = azuread_service_principal.example.id
+}
+```
+
+*Using custom settings*
+
+```terraform
+resource "azuread_application" "example" {
+  display_name = "example"
+}
+
+resource "azuread_service_principal" "example" {
+  application_id = azuread_application.example.application_id
+}
+
+resource "azuread_service_principal_token_signing_certificate" "example" {
+  service_principal_id = azuread_service_principal.example.id
+  display_name         = "CN=example.com"
+  end_date             = "2023-05-01T01:02:03Z"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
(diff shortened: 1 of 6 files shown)


tagur87:
@manicminer @katbyte - Welcoming feedback on this. I am sure there are some things that need addressed, but it works really well right now for adding the cert. 

We just need to figure out how to add the preferred thumbprint to the service principal. 

Since it would cause a circular dependency to reference this resource in the service_principal resource, I thought about adding a field to this resource such as `preferred_thumbprint = true` to set it as preferred...but not sure if there is a way to prevent that from being set in multiple instances of the resource.

tagur87:
Fixed the lint issues.

manicminer (approved):
Hi @tagur87, thanks for this contribution! My changes below are just some minor tweaks, this otherwise looks great and is good to merge 🎸 

In terms of setting the preferred certificate, I think perhaps a third resource is the most future-proof approach right now - particularly as the next major version of the provider will see more of these sorts of virtual resources. Something like:

```hcl
resource "azuread_service_principal_preferred_token_signing_certificate" "example" {
  service_principal_id = azuread_service_principal.example.object_id
  certificate_id       = azuread_service_principal_token_signing_certificate.example.id
}
```
Whilst that approach doesn't mitigate being able to create overlapping resources, it does reduce the config to a single place and will help with graphing dependencies in a future with more of these split-up virtual resources.

Since that can be added separately, we can still merge this PR.

**Test results**

Screen Shot 2023-01-18 at 23 04 17

Issue #68, opened 2019-03-25
Privileged Identity Management (PIM)


Lachlan-White:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Ability to Create PIM Policies and Configure PIM access via Terraform

### New or Affected Resource(s)

New data source would be required 

* azuread_privileged_identity_management

### Potential Terraform Configuration

```hcl
resource "azurerm_priviliged_identity_management" "PIM-Group-1" {
  scope = "Subscription_PIM_1"
  role_definition_name = "Contributor"
  aad_group_id = "${var.aad_group_id}"
}
```

### References
https://docs.microsoft.com/en-us/azure/active-directory/privileged-identity-management/pim-configure

mharrison365:
What would also be super helpful here would be support for "Azure Resources" PIM.

https://docs.microsoft.com/en-us/azure/active-directory/privileged-identity-management/pim-resource-roles-assign-roles

manicminer:
Thanks for requesting this feature! It looks like PIM management is only available via the [MS Graph API](https://docs.microsoft.com/en-us/graph/api/resources/privilegedidentitymanagement-root?view=graph-rest-beta) which we are currently not in a position to use as we're [waiting on SDK support](https://github.com/Azure/azure-sdk-for-go/issues/7352). I've marked this request as upstream-dependent and we'll revisit as soon as we are able.

khole:
maybe add the assignment type (eligible/active)
and assignment duration (start/end) 
as parameters?

LudoNieuwenhuizen:
Would be super handy for our production setup. Currently, this missing functionality means quite some manual work

manicminer:
This feature is currently waiting on implementation of MS Graph. We are actively working on it and this issue will receive priority as soon as we are able. See #323 for more details. Thanks!

frazhamid:
A few use cases I would like to see met:
- Adding Subscriptions
- Configuring Role Settings for both Azure AD and Azure Resources
- Configuring Alerts
- Configuring Access Reviews
- Managing Role Assignments (Add/Update/Remove)

[...]

PR #974, opened 2023-01-12
Add azuread_directory_role_eligibility_schedule_request resource

JonasBak:
This PR adds a `azuread_directory_role_eligibility_schedule_request` resource. We would like to be able to manage eligible roles in PIM with terraform. This PR lets us do it like this:

```
resource "azuread_directory_role_eligibility_schedule_request" "example" {
  role_definition_id = "00000000-0000-0000-0000-000000000000"
  principal_id       = "00000000-0000-0000-0000-000000000000"
  directory_scope_id = "/"
  justification      = "abc"
}
```

This is a pretty bare-bones version of the resource, but it gives us what we need for now. I have also opened a PR in https://github.com/manicminer/hamilton here: https://github.com/manicminer/hamilton/pull/204.

Related to https://github.com/hashicorp/terraform-provider-azuread/issues/68.

As I wrote in the other PR, I havn't written any tests, as I wanted to get some input first. The user/service principal requires [a role and a permission](https://learn.microsoft.com/en-us/graph/api/resources/privilegedidentitymanagementv3-overview?view=graph-rest-1.0#permissions-and-privileges) to be able to use the API, should I set these roles and permissions up as part of the tests?

### DIFF
--- a/internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
+++ b/internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
@@ -32,8 +32,8 @@ func directoryRoleEligibilityScheduleRequestResource() *schema.Resource {
 		},
 
 		Importer: tf.ValidateResourceIDPriorToImport(func(id string) error {
-			if id == "" {
-				return errors.New("id was empty")
+			if _, err := uuid.ParseUUID(id); err != nil {
+				return fmt.Errorf("specified ID (%q) is not valid: %s", id, err)
 			}
 			return nil
 		}),
--- a/internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
+++ b/internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
@@ -3,11 +3,13 @@ package directoryroles
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log"
 	"net/http"
 	"time"
 
 	"github.com/hashicorp/go-azure-sdk/sdk/odata"
+	"github.com/hashicorp/go-uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-azuread/internal/clients"
--- a/internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
+++ b/internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
@@ -1,0 +1,177 @@
+package directoryroles
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/hashicorp/go-azure-sdk/sdk/odata"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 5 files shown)


joraff:
Anything I can do to help keep this moving? Will the work in hamilton need to be completed first?

JonasBak:
I rebased and added a basic test, I did the same for the PR in hamilton.

internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
```diff
@@ -0,0 +1,177 @@
...
+					return "stub", "Waiting", nil
+				}
+				return nil, "Error", fmt.Errorf("retrieving role eligibility schedule request")
+			}
+			return "stub", "Done", nil
+		},
+	}).WaitForStateContext(ctx)
+	if err != nil {
+		return tf.ErrorDiagF(err, "Waiting for role eligibility schedule request for %q to reflect in directory role %q", principalId, roleDefinitionId)
+	}
```
manicminer, line 135: We have a helper func for this `helpers.WaitForUpdate()`

manicminer (changes requested):
Hi @JonasBak, thanks for working on this!

This is off to a great start. I've made a suggestion inline below. It also looks like we're missing documentation for the new resource, if you can look at these then I'll be happy to take another look and we can see about merging this. Thanks!

StickNitro:
Any update on this PR, I see the upstream changes are available with v0.62.0, this would be extremely beneficial to a piece of work I need to do

JonasBak:
I've rebased with main, added some docs, and added `helpers.WaitForUpdate()` as suggested

JonasBak:
@manicminer Could you take a look at this again?

internal/services/directoryroles/directory_role_eligibility_schedule_request_resource.go
```diff
@@ -0,0 +1,166 @@
...
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Read:   schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Importer: tf.ValidateResourceIDPriorToImport(func(id string) error {
+			if id == "" {
+				return errors.New("id was empty")
+			}
```
manicminer, line 37: It looks like the ID is a UUID so we should validate this for imports

```suggestion
			if _, err := uuid.ParseUUID(id); err != nil {
				return fmt.Errorf("specified ID (%q) is not valid: %s", id, err)
			}
```

manicminer (approved):
@JonasBak Thanks for the reminder and apologies for the delay in re-reviewing. I have only one minor suggestion regarding the import validation (see below) which I'll update. Other than that, this looks great and the test passes!

I had a go at adding support for `appScopeId` too, but I struggled to work out what to specify for this field as the API seems to silently ignore it instead of erroring (I tried an app ID, object ID of an application, object ID for a service principal). We can look to add this later.

manicminer:
Test results

Screenshot 2023-09-22 at 01 49 27

joraff:
@manicminer Thank you for approving this! We've been wanting to use it for a while.

FWIW - I don't think `appScopeId` is fully implemented; possibly reserved for future use? The default value of `/` is all that works. `directoryScopeId` is what can scope to an admin unit.

manicminer:
No problem, sorry again for the delay. Thanks for the feedback; I was able to scope a request to an application using the Portal, but alas the portal uses the PIM data plane API - so perhaps the MS Graph interface is incomplete as you say.

joraff:
Oh! Perhaps it is used now. It's been a while since I've done an assignment via the portal. Thanks to this, it might be a bit longer ;)

PR #980, opened 2023-01-18
azuread_group: fix for issue that prevents replacing a group when `prevent_duplicate_names = true`

manicminer:
Rather than coming up with a heuristic for determining whether a create is likely to be a replacement, we'll omit the duplicate name check at plan time when creating (but keep the check when updating) and defer the check to the create func (i.e. apply time). This is a workaround to be revisited when we move to terraform-plugin-framework at a future point.

Closes: #697
Replaces: #702


katbyte (approved):
LGTM 🏷️

Issue #942, opened 2022-12-02
email invitation triggered when creating Microsoft 365 group contains incorrect group name


clementa-lexum:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
Terraform v1.3.6
provider registry.terraform.io/hashicorp/azuread v2.31.0

### Affected Resource(s)

* resource azuread_group

### Terraform Configuration Files

```hcl
resource "azuread_group" "test_sharepoint" {
  display_name     = "Test SharePoint site"
  description = "Created with Terraform."

  auto_subscribe_new_members = false
  external_senders_allowed = false
  hide_from_address_lists = true
  hide_from_outlook_clients = false
  mail_enabled     = true
  mail_nickname    = "test_sharepoint"
  prevent_duplicate_names = true
  # provisioning_options = "Team"
  types            = ["Unified"]
  visibility = "Private" # Private ou Public

  owners = [ data.azuread_user.foo.object_id ]
  members = [
    data.azuread_user.bar.object_id
  ]
}

```
### Expected Behavior

Creation of SharePoint group sends an email invitation stating "Welcome to the Test SharePoint site Group.

### Actual Behavior

Creation of SharePoint group sends an email invitation stating "Welcome to the 
TERRAFORM_UPDATE_523a376b-0f95-6b2f-09d2-b5b5a9435f5a Group". 

![2022-12-02_14-05-53](https://user-images.githubusercontent.com/118307861/205368561-2e629dd5-92da-4de1-93b2-b156fed552fc.png)

The name of the group is correct though; only the invitation email contains the wrong name.

![2022-12-02_14-17-22](https://user-images.githubusercontent.com/118307861/205368924-5948128c-fecf-4096-9bde-cabc3a5911ce.png)

### Steps to Reproduce

1. `terraform apply`

manicminer:
Hi @clementa-lexum, thanks for reporting this. This is less than ideal, I will look at ways we can fix this.

PR #982, opened 2023-01-18
azuread_group: set the display name correctly on creation, and patch afterwards for consistency checks

manicminer:
Closes: #942


katbyte (approved):
LGTM 🍄

PR #984, opened 2023-01-23
azuread_group - Allow creation of group in administrative unit

### LINKED ISSUE #906: Create an Azure AD Group in an Administrative Unit
### Community Note

### Description

The azuread provider can add an existing azure ad group created at the root of the directory. This issue is to raise a feature request to align with the portal that has now the option to create an Azure AD Group directly into the an Administrative Unit (AU). 

Data source should also include the administrative unit when it is created only in an AU

image

* azuread_group (resource and data source)

```hcl
resource "azuread_administrative_unit" "example" {
  display_name = "Example-AU"
  description  = "Just an example"
  visibility   = "Public"
}

resource "azuread_group" "example" {
  display_name     = "example"
  security_enabled = true
  administrative_unit_object_id = azuread_administrative_unit.example.id
}

```

### References

https://learn.microsoft.com/en-us/graph/api/administrativeunit-post-members?view=graph-rest-1.0&tabs=http#example-2-create-a-new-group

image

ccadruvi:
Allows the creation of a group in an administrative unit directly. This allows to have less permissions, e.g. Groups Administrator on an administrative unit scope instead of tenant scope.

Fixes #906 

Depends on this PR in hamilton: https://github.com/manicminer/hamilton/pull/206
Edit: now also depends on https://github.com/manicminer/hamilton/pull/220


internal/services/groups/group_resource.go
```diff
@@ -795,6 +822,14 @@ func groupResourceCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 	}
 
+	// We have observed that when creating a group with an administrative_unit_id and querying the group with the /groups endpoint and specifying $select=allowExternalSenders,autoSubscribeNewMembers,hideFromAddressLists,hideFromOutlookClients, it returns a 404 for ~11 minutes.
```
manicminer, line 825: Thanks for the note with your observations, this is very helpful ⭐

docs/data-sources/group.md
```diff
@@ -38,6 +38,7 @@ The following arguments are supported:
 
 The following attributes are exported:
 
+* `administrative_unit_id` - The optional administrative unit ID that the group is member of.
```
manicminer, line 41: I don't think we should include this in the data source, since a group could be a member of any number of AUs including none.
```suggestion
```

docs/resources/group.md
```diff
@@ -106,6 +106,7 @@ resource "azuread_group" "example" {
 
 The following arguments are supported:
 
+* `administrative_unit_id` - (Optional) The administrative unit ID that the group is member of.
```
manicminer, line 109: We should be clearer about why this property is here and nudge the practitioner towards the more flexible `azuread_administrative_unit_member` resource or the `members` property of `azuread_administrative_unit`, and use this property only if they need to create a single AU scoped group.
```suggestion
* `administrative_unit_id` - (Optional) The object ID of a single administrative unit in which the group is a member. If specified, new groups will be created in the scope of the administrative unit. If empty, new groups will be created at the tenant level.

!> **Warning** Do not use the members property at the same time as the [azuread_administrative_unit_member](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/administrative_unit_member) resource, or the `members` property of the [azuread_administrative_unit](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/administrative_unit#members) resource, _for the same group_. Doing so will cause a conflict and administrative unit members will be removed.
```

go.mod
```diff
@@ -57,3 +57,6 @@ require (
 )
 
 go 1.19
+
+// can be removed once https://github.com/manicminer/hamilton/pull/206 is merged
+replace github.com/manicminer/hamilton => github.com/SwissGipfel/hamilton v0.54.1
```
manicminer, line 62: Can rebase and drop these, then `go mod tidy`

internal/services/groups/group_resource.go
```diff
@@ -566,9 +577,25 @@ func groupResourceCreate(ctx context.Context, d *schema.ResourceData, meta inter
...
+		auClient := meta.(*clients.Client).AdministrativeUnits.AdministrativeUnitsClient
+
+		var status int
+		group, status, err = auClient.CreateGroup(ctx, v.(string), &properties)
+		if err != nil {
+			return tf.ErrorDiagF(err, "Creating group in administrative unit with ID %q, %q", v.(string), displayName)
+		}
+		if status != http.StatusCreated {
+			return tf.ErrorDiagF(err, "Invalid status code after creating group %q", displayName)
+		}
```
manicminer, line 593: This isn't needed since the SDK checks for a valid status code
```suggestion
```

internal/services/groups/group_resource.go
```diff
@@ -795,6 +822,14 @@ func groupResourceCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 	}
 
+	// We have observed that when creating a group with an administrative_unit_id and querying the group with the /groups endpoint and specifying $select=allowExternalSenders,autoSubscribeNewMembers,hideFromAddressLists,hideFromOutlookClients, it returns a 404 for ~11 minutes.
+	if _, ok := d.GetOk("administrative_unit_id"); ok {
+		meta.(*clients.Client).Groups.GroupsClient.BaseClient.DisableRetries = false
+		meta.(*clients.Client).Groups.GroupsClient.BaseClient.RetryableClient.RetryWaitMax = 1 * time.Minute
+		meta.(*clients.Client).Groups.GroupsClient.BaseClient.RetryableClient.RetryWaitMin = 10 * time.Second
+		meta.(*clients.Client).Groups.GroupsClient.BaseClient.RetryableClient.RetryMax = 15
+	}
```
manicminer, line 831: This workaround is probably OK for now, we'll go with this and revisit when we have more information about consistency in this case.

internal/services/groups/group_resource.go
```diff
@@ -1065,6 +1100,29 @@ func groupResourceUpdate(ctx context.Context, d *schema.ResourceData, meta inter
...
+			if _, err := administrativeUnitClient.AddMembers(ctx, newAdministrativeUnitId, &members); err != nil {
+				return tf.ErrorDiagF(err, "Could not add group to administrative unit with object ID: %q", newAdministrativeUnitId)
+			}
+		}
+		if oldAdministrativeUnitId != "" {
+			memberIds := []string{d.Id()}
+			if _, err := administrativeUnitClient.RemoveMembers(ctx, oldAdministrativeUnitId, &memberIds); err != nil {
+				return tf.ErrorDiagF(err, "Could not remove group from administrative unit with object ID: %q", oldAdministrativeUnitId)
+			}
+		}
```
manicminer, line 1123: Whilst leaning on the diff is a useful shortcut, in this case we should look up the AU memberships for the group and remove all but the specified one (e.g. https://learn.microsoft.com/en-us/graph/api/group-list-memberof?view=graph-rest-1.0). This ensures that Terraform is declarative here in all cases, e.g. when importing groups, changing from multiple AUs to a single AU etc.

internal/services/groups/group_resource.go
```diff
@@ -1153,6 +1211,21 @@ func groupResourceRead(ctx context.Context, d *schema.ResourceData, meta interfa
...
+			return tf.ErrorDiagPathF(err, "administrativeUnit", "Could not retrieve administrative Unit where this group belongs to %q", d.Id())
+		}
+		if isMemberOfAu {
+			tf.Set(d, "administrative_unit_id", auId)
+		} else {
+			tf.Set(d, "administrative_unit_id", nil)
+		}
+	} else {
+		tf.Set(d, "administrative_unit_id", nil)
+	}
```
manicminer, line 1227: As above, with this I think the [`/groups/{id}/memberOf`](https://learn.microsoft.com/en-us/graph/api/group-list-memberof?view=graph-rest-1.0) endpoint is a better fit here. In the event that multiple AUs are found, we should sort by object ID and populate with the first one; this should avoid a persistent diff.

internal/services/groups/groups.go
```diff
@@ -42,6 +45,45 @@ func groupFindByName(ctx context.Context, client *msgraph.GroupsClient, displayN
...
+
+	for _, v := range data.Members {
+		if v.Id == auId {
+			return true, status, nil
+		}
+	}
+
+	return false, status, nil
+}
+
```
manicminer, line 86: As per above, I don't _think_ this function will be needed if we look up by group membership rather than per AU?

manicminer (changes requested):
Thanks for this contribution @ccadruvi! This is looking promising, however we will need to be very clear about how this will conflict with the `azuread_administrative_unit_member` resource and the `members` property of the `administrative_unit` resource. I've made some suggestions below, if you can look at these and rebase to include the SDK changes from https://github.com/manicminer/hamilton/pull/206 (released in v0.55.0), then I'll take another look. Thanks!

internal/services/groups/group_resource.go
ccadruvi: I'm not sure what would be a good way to proceed here.
With the current state of the PR, it is possible to create a group within an AU1 and then use the `administrative_unit_member` resource to add the same group to AU2.

If we follow your suggestion and make Terraform declarative in this case, reapplying the group would remove the administrative unit membership of AU2. That would mean that it's not possible to both create a group within an AU and also add it to a second AU, which seems limiting to me. When creating this PR I was wondering on whether `administrative_unit_id` should instead be a slice of IDs, as that is closer to how it works in Azure AD. We could follow a similar approach to how `owners` work: create the group within one AU and add other AU memberships afterwards.

internal/services/groups/group_resource.go
ccadruvi: In the following scenario:
```
resource "azuread_group" "group" {
  ...
  administrative_unit_id = "bcd"
}
```
then adding the group to a second AU with object ID `abc` (e.g. via Portal), this would result in the provider reading in the wrong object_id.

This also relates to the other comment on whether `administrative_unit_id` should be multiple IDs instead.

internal/services/groups/groups.go
ccadruvi: I think I don't understand what you mean here.
I can see two ways to get the information, whether a group is member of an AU:
1. List the "member of" of a group: https://learn.microsoft.com/en-us/graph/api/group-list-memberof?view=graph-rest-1.0&tabs=http
2. List the mebers of an AU: https://learn.microsoft.com/en-us/graph/api/administrativeunit-list-members?view=graph-rest-1.0&tabs=http

At the moment we're using option 1, are you suggesting to use option 2 instead?
I think from a permissions point of view, option 2 requires less privileges than option 1 (GroupMember.Read.All for option 1,  AdministrativeUnit.Read.All for option 2).

internal/services/groups/group_resource.go
manicminer: Ah I see what you mean. In this case I think your suggestion would work best - if we change `adminsitrative_unit_id` to `administrative_unit_ids` this would alleviate the challenge for users of introducing conflicts with `azuread_administrative_unit_member`, and should still offer decent functionality. We'll note in the docs that permission will be needed to add/create in each of the specified AUs else group creation will fail.

internal/services/groups/group_resource.go
manicminer: As above, let's make this property a list of AUs `administrative_unit_ids` and absolve ourselves of this problem :)

internal/services/groups/groups.go
manicminer: Sorry I seem to have misread this. Looking up by group membership feels like the right way to accomplish this, however I would amend this function to return a []string containing AU IDs (then iterating over that slice with a local function). And by doing that, there's no reason why this shouldn't make it upstream into the SDK as part of the GroupsClient.

e.g.

```go

// PROVIDER

isGroupInAdministrativeUnit := func(aus *[]msgraph.AdministrativeUnit, id string) bool {
	for _, au := range *aus {
		if au.ID == id {
			return true
		}
	}
	return false
}


// SDK

func (*GroupsClient) ListAdministrativeUnitMemberships(ctx context.Context, id string) (*[]AdministrativeUnit, int, error) {
	// ...
	return &[]AdministrativeUnit{}
}
```

Hope that makes sense?

internal/services/groups/group_resource.go
ccadruvi: I have changed it to take multiple `administrative_unit_ids` now. Could you look it over again please?
Also I added a note in the permissions section.

internal/services/groups/groups.go
ccadruvi: I have created this PR in hamilton:
https://github.com/manicminer/hamilton/pull/220
We don't need to check if the group is part of the au IDs that are part of the terraform configuration anymore, as now the `administrative_unit_ids` is declarative.

manicminer (approved):
Thanks for updating @ccadruvi. I made a small change so that the current AU memberships are retrieved when updating the group. Just waiting for test results but this LGTM 👍

manicminer:
Test results (failures are unrelated)

Screenshot 2023-02-21 at 17 27 53

PR #995, opened 2023-02-03
Bug in error message

rayterrill:
Error message is incorrect for mailNickname logic - trying to use upn vs mailNickname, results in an invalid error message/makes troubleshooting more difficult.

```
│ Error: User not found with email alias: ""
│
│   with module.groupmember["test"].data.azuread_user.user,
│   on ../../../modules/aadgroupassignment/main.tf line 3, in data "azuread_user" "user":
│    3:   mail_nickname = "test@test.com"
│
```


manicminer (approved):
Thanks for this fix @rayterrill, this LGTM 👍

Issue #997, opened 2023-02-06
Better support for multiple applications with same name from azuread_application datasource


gtmtech:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Trying to use a datasource to look up an azuread_application by name and receiving this error:

```
 Error: Multiple applications found
│ 
│   with module.foo.data.azuread_application.this,
│   on .terraform/modules/.../datasources.tf line 1, in data "azuread_application" "this":
│    1: data "azuread_application" "this" {
│ 
│ Found multiple applications matching filter: "displayName eq
│ 'myapp'"
```

Problem is that we have more than one application with the same name. I think some of our apps have been deleted, only one appears in searching in portal.azure, but they remain in the catalog, and the deletedDateTime is null for all of them. The Graph API returns all 2 or 3 apps, so I'm not sure how to distinguish the "real" one, except for the creation date being the latest. 

Therefore I would like to be able to return a single app with the most recent created date, in the event that there are more than one app found by name. 

### New or Affected Resource(s)

* data.azuread_application

### Potential Terraform Configuration

TBD?

Threpio:
This is an interesting issue and I am unaware that Azure would return multiple applications in the case that they are not visible within the portal and had been deleted/inactivated.

I would propose adding an additional argument along the lines of: 
```
data "azuread_application" "example" {
   display_name = "This Duplicate Application"
   return_latest_only = true
}
`
``

Would we want this to throw an error or exception if it only returns a single application and we expected multiple? Is there a situation where two different active applications would be required and working?

manicminer:
@gtmtech Thanks for reporting this. As far as I am aware, deleted applications should not be returned by the API, and in fact when testing it just now I do not observe this behavior. I suspect you that replication delay (aka eventual consistency) may have been at play when you observed this? If indeed the API is returning deleted applications, this is most definitely an API bug and should be reported as such to Microsoft.

[...]

PR #999, opened 2023-02-07
Updates to `application_data_source` - Issue #997

Threpio:
Changes whether an error will be thrown when multiple applications are returned from the API.


manicminer:
Hi @Threpio, thanks for proposing this. Unfortunately, I don't believe we'll merge this as I don't think this behavior is a good fit for Terraform and is likely to introduce more problems than it tries to solve. A data source should be a reliable, definitive source and this is the main reason why we currently return an error if multiple matching applications are found. For searching multiple apps, we should use a plural data source and the selection of a particular application should be the responsibility of the practitioner, so that they can be conscious of the associated risks and pitfalls.

Accordingly, I'm going to close this one out, but you are welcome to submit a PR for a (plural) `azuread_applications` data source, along the same lines as the `azuread_groups` and `azuread_users` plural data sources. Thanks!

Issue #914, opened 2022-10-27
Enabling mapped_claims_enabled returns 403 if set to true during application creation


PabloPie:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

- Terraform v1.2.7

- provider registry.terraform.io/hashicorp/azuread v2.29.0

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
resource "azuread_application" "app" {
  display_name = "myapp"

  api {
    mapped_claims_enabled          = true
    requested_access_token_version = 2
  }
}
```

### Debug Output

│ Error: Could not create application
│ 
│  OMITTED
│ 
│ ApplicationsClient.BaseClient.Post(): unexpected status 403 with OData
│ error: Authorization_RequestDenied: Insufficient privileges to complete the
│ operation.

### Expected Behavior

The application should have been created.

### Actual Behavior

The application fails to be created.

If the application is created with the mapped_claims set to false, and it is only enabled after creation, it works.

### Steps to Reproduce

1. Create an azuread_application with mapped_claims_enabled set to true.

This is done through a Service Principal that only has Application.ReadWrite.OwnedBy and not Application.ReadWrite.All permissions. I can't test it with the other permission.

waynejnicklin:
I'm experiencing this issue also. Is there a resolution?

waynejnicklin:
I had a notification that this "This functionality has been released in [v2.34.0 of the Terraform Provider]"... but that seems not to be released yet. Do you have a target date?

Thanks,
Wayne

manicminer:
Sorry about that, sometimes a false notification is posted if we neglect to change the milestone prior to release. I am hoping to get this merged soon.

waynejnicklin:
No worries thanks I hope so too! Microsoft still have no ETA on fixing the underlying API issue though!

manicminer:
I have been trying to code a workaround for this but unfortunately I suspect another breaking change in the Applications API which is preventing testing this.

waynejnicklin:
OK great thank you for the update :-)

[...]

PR #1008, opened 2023-02-13
Bugfix: work around an API bug where `acceptMappedClaims` cannot be set on create when holding `Application.ReadWrite.OwnedBy` role

manicminer:
Fixes: #914


katbyte (approved):
LGTM 🚜

PR #1011, opened 2023-02-15
Add `template_ids` to `azuread_directory_roles`

alexwilcox9:
Just a minor one, it turns out when using this with conditional access policies you need to provide the template IDs not the object IDs.. woops!

Adding them in so they can be easily used


manicminer (approved):
Thanks @alexwilcox9, this LGTM 🦘

manicminer:
Test result

Screenshot 2023-02-21 at 17 34 50

PR #1014, opened 2023-02-16
Bugfix: retrieve authenticated principal from MS Graph when `oid` claim is missing from access token

manicminer:
Related: https://github.com/hashicorp/terraform-provider-azurerm/pull/20465


katbyte (approved):
LGTM 🌩️

PR #1025, opened 2023-02-21
Workaround: auto-retry when looking for existing service principals, assuming that one is likely to exist when `use_existing = true`

manicminer:
When trying to adopt an existing service principal, through use of `use_existing = true`, that was very recently created - for example by instantiating an application template - perform retries when looking up the existing service principal to try and defeat eventual consistency. This is necessary because the 'existing' check requires a List operation with a filter, rather than a typical Get operation.


stephybun (approved):
Thanks @manicminer LGTM 🦁

manicminer:
Test results

Screenshot 2023-02-21 at 17 37 50

PR #1032, opened 2023-02-22
fixed #1031 adds provision on demand

### LINKED ISSUE #1031: Support synchronization provision on demand
### Community Note

### Description

Add support for provision on demand to allow terraform to ensure a group/user exists on a target system if that system is also terraform managed (e.g. aws sso ensure the group is provisioned so the permission set can be attached in a single apply)

* azuread_synchronization_job_provision_on_demand

```hcl
resource "azuread_synchronization_job_provision_on_demand" "example" {
  service_principal_id = "a service principle id"
  job_id = "a job id"
  parameters {
    rule_id = "an rule object id"
    subjects {
      object_id        = "an group/user object id"
      object_type_name = "Group"
    }
  }
}
```

### References

* #0000

iwarapter:
Hey @manicminer,

I have tested this with an AWS SSO configured app and it worked well, however I don't have access to an Azure tenant I can test the providers functional tests on. Not sure how best to proceed atm.

```

  # azuread_synchronization_job_provision_on_demand.example will be created
  + resource "azuread_synchronization_job_provision_on_demand" "example" {
      + id                   = (known after apply)
      + job_id               = "aWSSingleSignon.hidden"
      + service_principal_id = "hidden"

      + parameters {
          + rule_id = "hidden"

          + subjects {
              + object_id        = "hidden"
              + object_type_name = "Group"
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

azuread_synchronization_job_provision_on_demand.example: Creating...
azuread_synchronization_job_provision_on_demand.example: Creation complete after 4s [id=d66c068e-2711-409e-068d-8b55b8bb852e]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

Contains existing PR changes for reference only, will rebase once they are merged.

### DIFF
--- a/docs/resources/synchronization_job_provision_on_demand.md
+++ b/docs/resources/synchronization_job_provision_on_demand.md
@@ -1,0 +1,109 @@
+---
+subcategory: "Synchronization"
+---
+
+# Resource: azuread_synchronization_job_provision_on_demand
+
+Manages synchronization job on demand provisioning associated with a service principal (enterprise application) within Azure Active Directory.
+
+## API Permissions
+
+The following API permissions are required in order to use this resource.
+
+When authenticated with a service principal, this resource requires one of the following application roles: `Synchronization.ReadWrite.All`
+
+## Example Usage
+
+*Basic example*
+
+```terraform
+data "azuread_client_config" "current" {}
+
+resource "azuread_group" "example" {
+  display_name     = "example"
+  owners           = [data.azuread_client_config.current.object_id]
+  security_enabled = true
+}
+
+data "azuread_application_template" "example" {
+  display_name = "Azure Databricks SCIM Provisioning Connector"
+}
+
+resource "azuread_application" "example" {
+  display_name = "example"
+  template_id  = data.azuread_application_template.example.template_id
+  feature_tags {
+    enterprise = true
+    gallery    = true
+  }
+}
+
+resource "azuread_service_principal" "example" {
+  application_id = azuread_application.example.application_id
+  use_existing   = true
+}
+
+resource "azuread_synchronization_secret" "example" {
+  service_principal_id = azuread_service_principal.example.id
+
+  credential {
+    key   = "BaseAddress"
+    value = "https://adb-example.azuredatabricks.net/api/2.0/preview/scim"
+  }
+  credential {
+    key   = "SecretToken"
+    value = "some-token"
+  }
+}
+
+resource "azuread_synchronization_job" "example" {
+  service_principal_id = azuread_service_principal.example.id
+  template_id          = "dataBricks"
+  enabled              = true
+}
+
+resource "azuread_synchronization_job_provision_on_demand" "example" {
+  service_principal_id   = azuread_service_principal.example.id
+  synchronization_job_id = azuread_synchronization_job.example.id
+  parameter {
+    # see specific synchronization schema for rule id https://learn.microsoft.com/en-us/graph/api/synchronization-synchronizationschema-get?view=graph-rest-beta
+    rule_id = ""
+    subject {
+      object_id        = azuread_group.example.object_id
+      object_type_name = "Group"
+    }
+  }
+}
+
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+
+- `synchronization_job_id` (Required) Identifier of the synchronization template this job is based on.
+- `parameter` (Required) One or more `parameter` blocks as documented below.
+- `service_principal_id` (Required) The object ID of the service principal for the synchronization job.
+
+---
+
+`parameter` block supports the following:
+
+* `rule_id` (Required) The identifier of the synchronizationRule to be applied. This rule ID is defined in the schema for a given synchronization job or template.
+* `subject` (Required) One or more `subject` blocks as documented below.
+
+---
+
+`subject` block supports the following:
+
+* `object_id` (String) The identifier of an object to which a synchronizationJob is to be applied.
+* `object_type_name` (String) The type of the object to which a synchronizationJob is to be applied.
+
+## Attributes Reference
+
+No additional attributes are exported.
+
+## Import
+
+This resource does not support importing.
--- a/docs/resources/synchronization_job_provision_on_demand.md
+++ b/docs/resources/synchronization_job_provision_on_demand.md
@@ -39,8 +39,8 @@ resource "azuread_application" "example" {
 }
 
 resource "azuread_service_principal" "example" {
-  application_id = azuread_application.example.application_id
-  use_existing   = true
+  client_id    = azuread_application.example.client_id
+  use_existing = true
 }
 
 resource "azuread_synchronization_secret" "example" {
@@ -91,15 +91,15 @@ The following arguments are supported:
 
 `parameter` block supports the following:
 
-* `rule_id` (Required) The identifier of the synchronizationRule to be applied. This rule ID is defined in the schema for a given synchronization job or template.
+* `rule_id` (Required) The identifier of the synchronization rule to be applied. This rule ID is defined in the schema for a given synchronization job or template.
 * `subject` (Required) One or more `subject` blocks as documented below.
 
 ---
 
 `subject` block supports the following:
 
-* `object_id` (String) The identifier of an object to which a synchronizationJob is to be applied.
-* `object_type_name` (String) The type of the object to which a synchronizationJob is to be applied.
+* `object_id` (String) The identifier of an object to which a synchronization job is to be applied. Can be one of the following: (1) An onPremisesDistinguishedName for synchronization from Active Directory to Azure AD. (2) The user ID for synchronization from Azure AD to a third-party. (3) The Worker ID of the Workday worker for synchronization from Workday to either Active Directory or Azure AD.
+* `object_type_name` (String) The type of the object to which a synchronization job is to be applied. Can be one of the following: `user` for synchronizing between Active Directory and Azure AD, `User` for synchronizing a user between Azure AD and a third-party application, `Worker` for synchronization a user between Workday and either Active Directory or Azure AD, `Group` for synchronizing a group between Azure AD and a third-party application.
 
 ## Attributes Reference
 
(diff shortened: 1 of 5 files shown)


internal/services/serviceprincipals/registration.go
```diff
@@ -38,5 +38,6 @@ func (r Registration) SupportedResources() map[string]*schema.Resource {
 		"azuread_service_principal_token_signing_certificate":        servicePrincipalTokenSigningCertificateResource(),
 		"azuread_synchronization_job":                                synchronizationJobResource(),
 		"azuread_synchronization_secret":                             synchronizationSecretResource(),
+		"azuread_synchronization_job_provision_on_demand":            synchronizationProvisionOnDemandResource(),
```
manicminer, line 41: Can we order these alphabetically?

internal/services/serviceprincipals/synchronization_provision_on_demand_resource.go
```diff
@@ -0,0 +1,140 @@
...
+				ForceNew:         true,
+				ValidateDiagFunc: validate.UUID,
+			},
+			"job_id": {
+				Description: "The identifier for the synchronization jop.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			"parameters": {
```
manicminer, line 48: Since this is most likely to be used in a block, we can use singular naming since that translates well into config and is inline with other resources.

```suggestion
			"parameter": {
```

internal/services/serviceprincipals/synchronization_provision_on_demand_resource.go
```diff
@@ -0,0 +1,140 @@
...
+				ForceNew:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"rule_id": {
+							Description: "The identifier of the synchronizationRule to be applied. This rule ID is defined in the schema for a given synchronization job or template.",
+							Type:        schema.TypeString,
+							Required:    true,
+							ForceNew:    true,
+						},
+						"subjects": {
```
manicminer, line 61: As above, if this is likely to be used as a block then singular naming works best

```suggestion
						"subject": {
```

internal/services/serviceprincipals/synchronization_provision_on_demand_resource.go
```diff
@@ -0,0 +1,140 @@
...
+							Description: "The identifiers of one or more objects to which a synchronizationJob is to be applied.",
+							Type:        schema.TypeList,
+							Required:    true,
+							ForceNew:    true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"object_id": {
+										Description: "The identifier of an object to which a synchronizationJob is to be applied. Can be one of the following:\nAn onPremisesDistinguishedName for synchronization from Active Directory to Azure AD.\nThe user ID for synchronization from Azure AD to a third-party.\nThe Worker ID of the Workday worker for synchronization from Workday to either Active Directory or Azure AD.",
```
manicminer, line 69: We probably shouldn't use newlines in the description value

```suggestion
										Description: "The identifier of an object to which a synchronizationJob is to be applied. Can be one of the following: (1) An onPremisesDistinguishedName for synchronization from Active Directory to Azure AD. (2) The user ID for synchronization from Azure AD to a third-party. (3) The Worker ID of the Workday worker for synchronization from Workday to either Active Directory or Azure AD.",
```

internal/services/serviceprincipals/synchronization_provision_on_demand_resource.go
```diff
@@ -0,0 +1,140 @@
...
+										Type:        schema.TypeString,
+										Required:    true,
+									},
+									"object_type_name": {
+										Description: "The type of the object to which a synchronizationJob is to be applied. Can be one of the following:\nuser for synchronizing between Active Directory and Azure AD.\nUser for synchronizing a user between Azure AD and a third-party application.\nWorker for synchronization a user between Workday and either Active Directory or Azure AD.\nGroup for synchronizing a group between Azure AD and a third-party application.",
```
manicminer, line 74: As above
```suggestion
										Description: "The type of the object to which a synchronization job is to be applied. Can be one of the following: `user` for synchronizing between Active Directory and Azure AD, `User` for synchronizing a user between Azure AD and a third-party application, `Worker` for synchronization a user between Workday and either Active Directory or Azure AD, or `Group` for synchronizing a group between Azure AD and a third-party application.",
```

internal/services/serviceprincipals/synchronization_provision_on_demand_resource.go
```diff
@@ -0,0 +1,140 @@
...
+										Type:        schema.TypeString,
+										Required:    true,
```
manicminer, line 76: If these values are known, we should validate them:

```suggestion
										Required:    true,
										ValidateFunc: validation.StringInSlice([]string{
											"Group",
											"user",
											"User",
											"Worker".
										},
```

internal/services/serviceprincipals/synchronization_provision_on_demand_resource.go
```diff
@@ -0,0 +1,140 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"service_principal_id": {
+				Description:      "The object ID of the service principal for which this synchronization job should be provisioned",
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validate.UUID,
+			},
+			"job_id": {
```
manicminer, line 42: Let's make this more explicit

```suggestion
			"synchronization_job_id": {
```

manicminer (changes requested):
@iwarapter Thanks for working on this. If you aren't able to run acceptance tests that's no problem, we can run them in our testing tenant. I'm happy to make small fixes to the tests as needed and I can report test results back to you for potentially larger fixes.

On the whole this is looking good. We will need to rewrite the doc as the format/style produced by tfplugindocs unfortunately is different to the provider docs - I recommend copying an existing doc and changing it as needed. Additionally if you can look at the comments below, and finish up the Read, Update and Delete funcs, then I'll be happy to take another look. Thanks!

iwarapter:
hey @manicminer, thanks for the feedback. I have addressed all the parts you raised. For the acceptance test i'd need to know what the `rule_id` are available in the databricks schema in order for it to run. We could also do with a data source for this, but didnt see it in the graph api lib - https://learn.microsoft.com/en-us/graph/api/synchronization-synchronizationschema-get?view=graph-rest-beta&tabs=http

iwarapter:
@manicminer is there anything further I can do to help with this?

manicminer:
@iwarapter Thanks, at this time I don't have the bandwidth to create tests for this, so if you could have a go at writing some acceptance tests that will help enormously. I don't mind kicking the tires and tweaking existing tests to get them working reliably, but this will be waiting a little while if I have to write the tests. Thanks!

iwarapter:
hey @manicminer i had a quick stab - https://github.com/iwarapter/terraform-provider-azuread/commit/9da78c0ed05382143b0e90ee0cce37b1b143c576 there are a few //TODO which i need extra info for

iwarapter:
@manicminer have pushed the tests - from what i can see the databricks integration used for acctests isnt fully integrated so this can't be tested e2e with it (im just testing config and then the expected failure). is there another app the acctests can use to test full e2e provisioning?

internal/services/serviceprincipals/synchronization_job_provision_on_demand_resource_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (r SynchronizationJobProvisionOnDemandResource) basic(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+%[1]s
+
+resource "azuread_synchronization_job_provision_on_demand" "test" {
+  service_principal_id = azuread_service_principal.test.id
+  synchronization_job_id = trimprefix(azuread_synchronization_job.test.id, "${azuread_service_principal.test.id}/job/")
+  parameter {
+	rule_id = "03f7d90d-bf71-41b1-bda6-aaf0ddbee5d8" //no api to check this so assuming the rule id is the same globally :finger_crossed:
```
Dave-Snigier, line 77: FWIW, this ID is the same in my tenant

manicminer (commented):
@iwarapter Apologies for the delay in re-reviewing this. I was originally anticipating the resource would have Read and Delete functions, but on closer inspection of the API and provisioning process, I see that there is no way to track a single "provision".

This makes it impossible to track state in Terraform for this operation, as this is more of a fire and forget task than a resource having some sort of lifecycle. With that in mind, I'm unsure whether this would be a good fit for Terraform unless we can find some other way to track state here? Alternatively, if the primary use case is to provision users, this could possibly make sense as an optional block in the `azuread_user` resource instead? That way, we'd have some tangible state to track after the sync job has been triggered. WDYT?

iwarapter:
Hi @manicminer, thanks for getting back to this. Thats correct, but there is prior art for this style of resource - mainly the [aws lambda invocation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_invocation) resource.

The only way you could check if the resource has been provisioned currently is by calling the provision api and checking the response code - but this is definitely not what we want in the READ part 😄. I certainly wouldn't bundle this into any other resources as the closest one is the `azuread_synchronization_job`.

I disagree that it's not a good fit for terraform as it enables a key part of fully managing resources across azuread and a given service provider. Here's a quick example use case:

Add a group to a azure application (all covered with this provider)
GAP
Assign a SCIM managed group to a given permission set within AWS (all covered with the AWS provider)

This resource solves the gap between the two where the group won't exist in AWS as you at the mercy of the SCIM scheduler. I deploy many hundreds of groups/role bindings and this is quite painful.

manicminer:
@iwarapter Thanks for the context, I see now that the sync can happen in either direction and the resulting object(s) are not likely to be directly managed anyway. In which case this seems like a reasonable way to implement this. Do you think it's worth adding a `triggers` faux-meta argument (which is conventionally a map) to enable programmatic reprovisioning?

iwarapter:
its not a bad shout, i cant think of any (personal) use cases for it. But simple enough to just add a map trigger

iwarapter:
I'll try add this tomorrow morning

iwarapter:
@manicminer lifted straight from the aws provider 😂

manicminer (approved):
@iwarapter Thanks, I pushed some minor style fixes and also a crash fix in `expandSynchronizationJobApplicationParameters()`. LGTM! 👍

iwarapter:
cool glad to see this getting merged :)

manicminer:
Test result

Screenshot 2024-05-09 at 12 49 18

PR #1038, opened 2023-02-27
Acceptance testing: build separate client for checks

manicminer:
Needed since terraform-plugin-sdk v2.24.1 cancels the context for the test provider once the config has been applied, causing checks to fail

Based on the [AzureRM implementation](https://github.com/hashicorp/terraform-provider-azurerm/blob/42367ac6c4799252c9c3ac54da51d2e76314266c/internal/acceptance/check/that.go#L42-L45)


katbyte (approved):
LGTM 🚂

PR #1039, opened 2023-03-01
Update Hamilton to v0.59.0

manicminer:
Changelog:
* **Provider:** requests to Microsoft Graph no longer include the tenant ID as part of the URI path

Notably related: https://github.com/manicminer/hamilton/pull/230


katbyte (approved):
LGTM 📦

PR #1045, opened 2023-03-14
Fix accuracy in documentation of Application permissions for azuread_group and azuread_group_member

audunsolemdal:
Tested the changes described today and it works. This is a more least-privilege way to grant access to create AAD groups and memberships.

### DIFF
--- a/docs/resources/group.md
+++ b/docs/resources/group.md
@@ -10,7 +10,9 @@ Manages a group within Azure Active Directory.
 
 The following API permissions are required in order to use this resource.
 
-When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`. Alternatively you can grant the principal the application role `Group.Create`, and make the principal as part of the owners of the group.
+When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`.
+
+Alternatively, if the authenticated service principal is also an owner of the group being managed, this resource can use the application role: `Group.Create`.
 
 If using the `assignable_to_role` property, this resource additionally requires one of the following application roles: `RoleManagement.ReadWrite.Directory` or `Directory.ReadWrite.All`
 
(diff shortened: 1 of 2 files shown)


docs/resources/group.md
```diff
@@ -10,7 +10,7 @@ Manages a group within Azure Active Directory.
 
 The following API permissions are required in order to use this resource.
 
-When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`
+When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`. Alternatively you can grant the principal the application role `Group.Create`, and make the principal as part of the owners of the group.
```
manicminer, line 13:
```suggestion
When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`.

Alternatively, if the authenticated service principal is also an owner of the group being managed, this resource can use the application role: `Group.Create`.
```

docs/resources/group_member.md
```diff
@@ -12,7 +12,7 @@ Manages a single group membership within Azure Active Directory.
 
 The following API permissions are required in order to use this resource.
 
-When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`
+When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`. If the service principal is an owner of the group, these permissions are not required.
```
manicminer, line 15:
```suggestion
When authenticated with a service principal, this resource requires one of the following application roles: `Group.ReadWrite.All` or `Directory.ReadWrite.All`.

However, if the authenticated service principal is an owner of the group being managed, an application role is not required.
```

manicminer (approved):
@audunsolemdal Thanks for clarifying and suggesting these changes. I have tweaked the wording slightly to avoid conflating permissions/roles, but this otherwise LGTM 👍

Issue #1049, opened 2023-03-14
azuread_service_principal cannot clear all owners


Noel-Jones:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform version: 1.4.0

provider registry.terraform.io/hashicorp/azuread v2.36.0

### Affected Resource(s)

* `azuread_service_principal`

### Terraform Configuration Files

```hcl
data "azuread_client_config" "current" {}

resource "azuread_application" "example" {
  display_name = "noeltest"
  owners       = [data.azuread_client_config.current.object_id]
}

resource "azuread_service_principal" "example" {
  application_id               = azuread_application.example.application_id
  app_role_assignment_required = false
  owners                       = [data.azuread_client_config.current.object_id]
}
```

Terraform Apply, then set the servce principal ownerd to an empy set (or comment it out)

```hcl
data "azuread_client_config" "current" {}

resource "azuread_application" "example" {
  display_name = "noeltest"
  owners       = [data.azuread_client_config.current.object_id]
}

resource "azuread_service_principal" "example" {
  application_id               = azuread_application.example.application_id
  app_role_assignment_required = false
  owners                       = []
}
```

Apply again, and again.

### Expected Behavior

Owner should be removed on second apply and no change should show after this.

### Actual Behavior

Plan shows owner will be removed on second apply but is not actually removed. Same with third apply etc

### Steps to Reproduce

As detailed above

1. `terraform apply`

### Important Factoids

None. I have however done some investigation and would point towards  line 542 in servicePrincipalResourceUpdate

`	if v, ok := d.GetOk("owners"); ok && d.HasChange("owners") {

If I undesrstand it correctly this will not be "ok" with an empty or null set so the code to remove the owner(s) is not executed.

### References

* #901 appears to resolve this same issue at the equivalent line

Issue #901, opened 2022-10-06
azuread_application "owners" field keeps appearing in plans even after the change has been applied


leewilson86:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v1.0.11
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.28.1
+ provider registry.terraform.io/hashicorp/azurerm v2.71.0
+ provider registry.terraform.io/hashicorp/consul v2.14.0
+ provider registry.terraform.io/hashicorp/external v2.1.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/hashicorp/random v3.4.3
```

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

```hcl
resource "azuread_application" "azureAdApplication" {
  count                   = length(local.applications)
  display_name            = local.applications[count.index].display_name
  sign_in_audience        = local.signInAudience
  group_membership_claims = var.groupMembershipClaims

  api {
    known_client_applications      = []
    mapped_claims_enabled          = false
    requested_access_token_version = 1

    oauth2_permission_scope {
      admin_consent_description  = format("Allow the application to access %s on behalf of the signed-in user.", local.applications[count.index].display_name)
      admin_consent_display_name = format("Access %s", local.applications[count.index].display_name)
      enabled                    = true
      id                         = local.applications[count.index].oauth2_permission_scope_id
      type                       = "User"
      user_consent_description   = format("Allow the application to access %s on your behalf.", local.applications[count.index].display_name)
      user_consent_display_name  = format("Access %s", local.applications[count.index].display_name)
      value                      = "user_impersonation"
    }
  }

  web {
    implicit_grant {
      id_token_issuance_enabled = var.idTokenIssuanceEnabled
    }
    redirect_uris = var.redirectUris
  }

  dynamic "required_resource_access" {
    iterator = resource
    for_each = var.requiredResourceAccess

    content {
      resource_app_id = resource.value.id

      dynamic "resource_access" {
        iterator = access
[...]

PR #1050, opened 2023-03-15
Bugfix: serviceprincipal owner removal #1049

Noel-Jones:
GetOK does not (I believe) return ok when passed an empty set or null value. Therefore the code section is not executed when trying to clear all owners. Resolution matches that for #901 as applied to applications. Reviewed tests and these are the same as those made against applications - though there is a gap in the tests given this issue was not caught but I see no way to address it.


manicminer (approved):
Hi @Noel-Jones, thanks for picking this up and opening this PR! This LGTM, I have just updated the tests so that we can check for the owner removal.

PR #1052, opened 2023-03-20
Fix api permission docs on azuread_application resource

PSanetra:
According to the provider docs one of `Directory.ReadWrite.All` or `Application.ReadWrite.All` are sufficient to use the `azuread_application` resource. This is not true. According to my experience and to the official REST api docs just `Application.ReadWrite.All` is required: https://learn.microsoft.com/en-us/graph/api/application-post-applications?view=graph-rest-beta&tabs=http#permissions

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -10,7 +10,7 @@ Manages an application registration within Azure Active Directory.
 
 The following API permissions are required in order to use this resource.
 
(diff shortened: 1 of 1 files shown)


manicminer (approved):
Thanks for this @PSanetra. I've tested this and `Directory.ReadWrite.All` indeed appears to no longer work.

Issue #1067, opened 2023-04-14
azuread_group onpremises_group_type change appears during terraform plan even if writeback is disabled


dennis1f:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v1.3.6
on windows_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.37.0
``` 

### Affected Resource(s)

* `azuread_group`

### Terraform Configuration Files

```hcl
resource "azuread_group" "aad_groups" {
  for_each = {
    for group in var.Groups :
      "${var.EnvironmentId}_${group.Role}_${group.Scope}" => group
  }

  display_name = "${var.EnvironmentId}_${each.value.Role}_${each.value.Scope}"
  security_enabled = true

  # Trying to get rid of onpremises_group_type change
  writeback_enabled = null
  onpremises_group_type = null

  # This one is an ugly one:
  # https://github.com/hashicorp/terraform-provider-azuread/issues/624
  # https://github.com/microsoftgraph/msgraph-metadata/issues/92
  # So during creation ms automatically adds the executing user as owner
  # However then complains about duplicate ids, because the executing user
  # is one of the owners.
  # Set the flag SkipOwner to true when creating resource and remove the flag again
  # once resource is created.
  owners = setunion(
    var.OwnerIds,
    toset([
      for owner in data.azuread_users.owners.users :
        owner.object_id if each.value.SkipOwner == null || each.value.SkipOwner == false || owner.object_id != data.azuread_client_config.current.object_id
    ])
  )

  members = each.value.Members
}
```

### Debug Output

https://gist.github.com/dennis1f/1eacc03071c3f3886b9d5fd104a2cb3b

### Expected Behavior
No change should show up during terraform plan

### Actual Behavior
I believe #964 introduced a regression when writeback is disabled for a group.
In our environment writeback is disabled: `"writebackConfiguration":{"isEnabled":null,"onPremisesGroupType":null}` response from MSGraph API
However since version 2.37.0 the following change appears during terraform plan.
```
[...]

PR #1070, opened 2023-04-17
Fix regression: writeback configuration should be omitted on creation

manicminer:
Writeback configuration should be omitted on creation instead of using the default value, and unsetting `onpremises_group_type` should cause a ForceNew

Fixes: #1067

### DIFF
--- a/docs/resources/group.md
+++ b/docs/resources/group.md
@@ -141,7 +141,7 @@ The following arguments are supported:
 
 !> **Warning** Do not use the `members` property at the same time as the [azuread_group_member](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/group_member) resource for the same group. Doing so will cause a conflict and group members will be removed.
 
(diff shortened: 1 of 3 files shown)


catriona-m (approved):
LGTM!

katbyte (approved):
LGTM 🌩️

manicminer:
**Test results**

Screenshot 2023-04-17 at 22 33 09

audunsolemdal:
I had already applied this change to quite a few groups in 2.37.0, and I would like the groups not to be recreated as it would cause some extra work in fixing broken role assignments

Is it possible to reset the onpremsies_group_type value to null without recreating the group, via Azure Powershell for instance?

adding ignore_changes is an option, but I consider it suboptimal

manicminer:
@audunsolemdal Unfortunately it doesn't seem to be possible to unset this field via Azure Portal or PowerShell. This is an API bug that will hopefully be fixed at some point, which will enable us to remove the conditional ForceNew enforcement for this property. In the meantime, the best workaround is to allow Terraform to recreate the group, or to use the `ignore_changes` meta property to suppress this diff,

Issue #1072, opened 2023-04-18
azuread_group: 2.37.0 > 2.37.1 causes group replacement


Official-James:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

AzureAD: 2.37.0 > 2.37.1

### Affected Resource(s)

Seems to be all azuread resources

### Terraform Configuration Files

```hcl
terraform {
  backend "azurerm" {}
  required_providers {
    azuread = {
      version = "2.37.1" #Originally 2.27.0
    }
  }
}
```

### Expected Behavior

No destroy/create of resources

### Actual Behavior

All resources will destroy and create (force replaced) due to moving from version 2.37.0 to 2.37.1. It seems to be the following causing this:

```cli
- onpremises_group_type          = "UniversalSecurityGroup" -> null # forces replacement
```

### Steps to Reproduce

1. Already run `terraform apply` using provider version `2.37.0`
2. Amend provider version to `2.37.1`
3. Run `terraform apply`

### Important Factoids

Yesterday's release of 2.37.0 has a bug that was introduced. During this time, resources were deployed, which amended all azuread resources. Now moving to anything other than 2.37.0 results in all resources being destroyed and recreated.

### References

* [1067](https://github.com/hashicorp/terraform-provider-azuread/issues/1067) - was found to identify the problem and fix was implemented.

manicminer:
@Official-James Thanks for reporting this issue. As you have accurately summarised, this breaking change was reluctantly introduced in order to fix a bug introduced with v2.37.0. Ordinarily this would not have as much of an issue, but in this case, the API has a bug which makes it impossible to unset this field.

Unfortunately, working around this any other way would have created other side effects and simply postponed the problem to a future release where it's likely that more practitioners would be affected. However, there are a couple of options available to you if you are now presented with this unexpected diff.

The first and easiest option would be to allow Terraform to delete and recreate the groups. Assuming that you are using Terraform to manage manual group memberships and other dependencies such as role assignments, this will momentarily break your setup but should restore everything in a desired state after the apply is completed.

[...]

PR #1076, opened 2023-04-19
azuread_group: regression fixes

manicminer:
- `onpremises_group_type` should be optional+computed to silently ignore an unsettable value, see https://github.com/hashicorp/terraform-provider-azuread/issues/1072 and https://github.com/microsoftgraph/msgraph-metadata/issues/332
- workaround for reading M365-only fields in a non-M365 tenant, see https://github.com/microsoftgraph/msgraph-metadata/issues/333

Resolves: #1072


manicminer:
**Test results**

Screenshot 2023-04-19 at 23 43 41

mbfrahry (approved):
LGTM!

Issue #1083, opened 2023-04-25
Inconsistent provider retry behavior due to DisableRetries manipulation


benoit74:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v1.3.7
on linux_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.37.2
```

### Affected Resource(s)

Behavior observed on following resource but probably most of them are concerned

* `azuread_application`

### Terraform Configuration Files

```hcl
data "azuread_application_template" "aws_single_account_access" {
  display_name = "AWS Single-Account Access"
}

resource "azuread_application" "one_app" {
  display_name = "One AWS Access App"
  template_id  = data.azuread_application_template.aws_single_account_access.template_id
}
```

### Debug Output

First execution (init/plan/apply) is OK : https://gist.github.com/benoit74/23617a452a131367f76aad4e308f95a1#file-first_apply-log

The issue arises if for some reason the resource has to be recreated (due to force-recreate or taint typically) : https://gist.github.com/benoit74/23617a452a131367f76aad4e308f95a1#file-taint_and_reapply-log

### Expected Behavior

Taint should has simply allowed to recreate the application.

### Actual Behavior

Application recreation failed ; destroy was ok but recreation failed very quickly.

### Steps to Reproduce

1. `terraform plan -out plan.out`
1. `terraform apply plan.out`
1. `terraform taint azuread_application.one_app`
1. `terraform plan -out plan.out`
1. `terraform apply plan.out`

manicminer:
@benoit74 Thank you for raising this issue and for digging into the behavior. I agree this is most likely a case of clients being shared between operations, particularly during replace operations where a resource is deleted and recreated whilst sharing the same client in memory.

I've pushed some mitigations which will be included in today's provider release. This may not solve the issue entirely but should hopefully prevent the particular case of resource replacement that you are encountering. Unfortunately this is due to design flaws in the SDK client, and whilst it's unlikely we'll fix these in the current SDK, we'll take this learning into our new SDK to prevent this happening in the future.

PR #1090, opened 2023-04-27
Bugfix: mitigate inheritence or long race condition between resource functions when clients have retries disabled

manicminer:
Because clients are shared between different operations on resources, sometimes a client can have Retries disabled and this is inherited by another resource function. This aims to mitigate this by always trying to re-enable retries wherever they are temporarily disabled for polling etc.

Closes: #1083 

(Linking the above issue since whilst this doesn't entirely mitigate the shared memory aspect, it should mitigate the particular case being reported in that issue)


mbfrahry (approved):
LGTM

PR #1098, opened 2023-05-11
Fix typo on debugging docs

iwarapter:
Fixing typo as per https://developer.hashicorp.com/terraform/plugin/log/managing


manicminer (approved):
Good spot, thanks for this fix @iwarapter 👍

PR #1122, opened 2023-06-15
allow zero length string for user password

BeLeap:
According to [document](https://github.com/hashicorp/terraform-provider-azuread/blob/debc25099fb72e88b0cb2e2d796c7825188102ef/docs/resources/user.md?plain=1#L60), setting password to empty string is allowed.
Also, [code](https://github.com/hashicorp/terraform-provider-azuread/blob/debc25099fb72e88b0cb2e2d796c7825188102ef/internal/services/users/user_resource.go#L550-L555) checks if password is empty and does not update password if it is.
Therefore, I think zero length string(empty string) should be allowed.

<details>
<summary>Screenshot</summary>

1. Document
image

2. Code
image
</details>

I created a test, but I'm not sure if I created it correctly.

### DIFF
--- a/internal/services/users/user_resource.go
+++ b/internal/services/users/user_resource.go
@@ -230,7 +230,7 @@ func userResource() *schema.Resource {
 				Optional:     true,
 				Computed:     true,
 				Sensitive:    true,
-				ValidateFunc: validation.StringLenBetween(1, 256), // Currently the max length for AAD passwords is 256
+				ValidateFunc: validation.StringLenBetween(0, 256), // Currently the max length for AAD passwords is 256
 			},
 
 			"disable_strong_password": {
--- a/internal/services/users/user_resource_test.go
+++ b/internal/services/users/user_resource_test.go
@@ -125,6 +125,21 @@ func TestAccUser_passwordOmitted(t *testing.T) {
 	})
 }
 
+func TestAccUser_passwordEmpty(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azuread_user", "test")
+	r := UserResource{}
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config: r.passwordEmpty(data),
+			Check: resource.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+			),
+		},
+		data.ImportStep("force_password_change", "password"),
+	})
+}
+
 func (r UserResource) Exists(ctx context.Context, clients *clients.Client, state *terraform.InstanceState) (*bool, error) {
 	client := clients.Users.UsersClient
 	client.BaseClient.DisableRetries = true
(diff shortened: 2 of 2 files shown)


manicminer:
Hi @BeLeap, thanks for suggesting this change. Could you explain a little further what you'd like to achieve by allowing an empty password? The API does not permit this, yielding a 400 error for example:

```json
{
  "error": {
    "code": "Request_BadRequest",
    "message": "A password must be specified to create a new user."
  }
}
```

If you are struggling to make your configuration work for existing users, have you tried setting the property to `null` instead of an empty string? This indicates to Terraform that the property is unset.

BeLeap:
@manicminer

Thank you for your response.
I thought if `password` is a blank string then it does not include the password in the request.

Following is my use case.
Because I have plenty of users to manage, I used `for_each` for `azurerm_user` resources.
```terraform
locals {
  members = {
    "foo@example.com" = {
      display_name = "foo"
    }
    "bar@example.com" = {
      display_name = "bar"
    }
  }
}

resource "azuread_user" "users" {
  for_each            = local.members
  user_principal_name = each.key
  display_name        = each.value.display_name
  usage_location      = "KR"
}
```

However, adding a user requires set password explicitly and I don't want to update the password of users afterward.
It can be accomplished by adding `ignore_changes` but I found that there is a document that says "setting password to a blank string" will not remove the password.
Thus I tried the following code to create a new user and change password to a blank string afterward.
```terraform
locals {
  members = {
    "foo@example.com" = {
      display_name = "foo"
      password     = ""
    }
    "bar@example.com" = {
      display_name = "bar"
      password     = ""
    }
	"quuz@exmaple.com" = {
      display_name = "quuz"
      password     = "some-random-password"
    }
  }
}

resource "azuread_user" "users" {
  for_each            = local.members
  user_principal_name = each.key
  display_name        = each.value.display_name
  usage_location      = "KR"
}
```

The above code was not working because azurerm terraform provider doesn't allows it.

Therefore I read the code of terraform provider and I found some code that looks like not update password if the given password is a blank string but the validator doesn't allow it.
That's why I proposed these changes.

If this changes is wrong or not intended, I think the document that says "setting password to a blank string, will not remove the password" should be updated because "setting password to a blank string" is invalid.

manicminer:
Thanks for the feedback @BeLeap. I agree the doc needs a correction here since blank string values are not allowed.

To clarify the provider behavior around passwords, a password is only set when creating a user, or when Terraform detects that the password value in the configuration has changed. The API does not return the password so it's not possible for Terraform to enforce a particular password over time. This means that the password is not updated by Terraform after the user sets their own password, so you can create users with a temporary password (set in your configuration), and provided you do not change this to a different value in your configuration over time, the password will not be modified by Terraform again. You can also remove the `password` property from your configuration and similarly, Terraform will not attempt to update or clear the password, so you can also import users in this way without accidentally manipulating their password.

Hope this clarifies things a little! Since I don't believe there's anything for us to do here to achieve your desired outcome, I'm going to close this for now, but please feel free to comment further if I have misunderstood and I'll be happy to take another look. Thanks!

PR #1152, opened 2023-07-24
directory role templates

MattGarnerAWR:


docs/data-sources/directory_role_templates.md
```diff
@@ -0,0 +1,44 @@
+---
+subcategory: "Directory Role Templates"
+---
+
+# Resource: azuread_directory_role_templates
+
+Use this data source to access information about activated directory role templates within Azure Active Directory.
```
manicminer, line 7: Does this return only activated templates, or does it return all templates whether activated or not?

internal/services/directoryroles/directory_role_templates_data_source.go
```diff
@@ -0,0 +1,104 @@
...
+		return tf.ErrorDiagF(err, "Could not retrieve role templates")
+	}
+	if directoryRoleTemplates == nil {
+		return tf.ErrorDiagF(errors.New("API error: nil directoryRoleTemplates were returned"), "Retrieving all directory role templates")
+	}
+
+	objectIds := make([]string, 0)
+	templateList := make([]map[string]interface{}, 0)
+
+	for _, r := range *directoryRoleTemplates {
```
manicminer, line 82: Should we be filtering out the `Users` role template? Per the docs:

> Note: A directory role template is exposed for the Users directory role. The Users directory role is implicit and is not visible to directory clients. Every User in the tenant is assigned to this role by the infrastructure. The role is already activated. Do not use this template.

manicminer (commented):
Hi @MattGarnerAWR, thank you for this PR! This is looking great, and I do not have any changes to suggest - just a couple questions to clarify before we look to merge this.

docs/data-sources/directory_role_templates.md
MattGarnerAWR: Removed word, not needed as all templates are returned

internal/services/directoryroles/directory_role_templates_data_source.go
MattGarnerAWR: Excluded "Users"

MattGarnerAWR:
> Hi @MattGarnerAWR, thank you for this PR! This is looking great, and I do not have any changes to suggest beyond what the tests have flagged - just a couple questions to clarify before we look to merge this.

Thanks, I have resolved all of them and hopefully the right package will mean the tests run this time :)

internal/services/directoryroles/directory_role_templates_data_source.go
```diff
@@ -0,0 +1,106 @@
...
+		if r.DisplayName != "User" {
+			objectIds = append(objectIds, *r.ID())
+
+			template := make(map[string]interface{})
+			template["description"] = r.Description
+			template["display_name"] = r.DisplayName
+			template["object_id"] = r.ID()
+			templateList = append(templateList, template)
+		}
```
manicminer, line 91: Just a minor tweak, but it's marginally nicer to skip the loop early rather than nest an `if` block :)

```suggestion
		// Skip the implicit "Users" role as it's non-assignable
		if r.DisplayName != "User" {
			continue
		}

		objectIds = append(objectIds, *r.ID())

		template := make(map[string]interface{})
		template["description"] = r.Description
		template["display_name"] = r.DisplayName
		template["object_id"] = r.ID()
		templateList = append(templateList, template)
```

manicminer (approved):
Thanks @MattGarnerAWR, just a couple minor tweakd which I've made below but otherwise LGTM 👍

internal/services/directoryroles/directory_role_templates_data_source.go
```diff
@@ -0,0 +1,109 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func directoryRoleTemplatesDataSourceRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client).DirectoryRoleTemplates.DirectoryRoleTemplatesClient
```
manicminer, line 69:
```suggestion
	client := meta.(*clients.Client).DirectoryRoles.DirectoryRoleTemplatesClient
```

internal/services/directoryroles/directory_role_templates_data_source.go
```diff
@@ -0,0 +1,109 @@
...
+	if directoryRoleTemplates == nil {
+		return tf.ErrorDiagF(errors.New("API error: nil directoryRoleTemplates were returned"), "Retrieving all directory role templates")
+	}
+
+	objectIds := make([]string, 0)
+	templateList := make([]map[string]interface{}, 0)
+
+	for _, r := range *directoryRoleTemplates {
+		// Skip the implicit "Users" role as it's non-assignable
+		if r.DisplayName == nil || *r.DisplayName == "User" {
```
manicminer, line 84:
```suggestion
		if r.ID == nil || r.DisplayName == nil || *r.DisplayName == "User" {
```

internal/services/directoryroles/directory_role_templates_data_source.go
```diff
@@ -0,0 +1,109 @@
...
+			continue
+		}
+
+		objectIds = append(objectIds, *r.ID())
+
+		template := make(map[string]interface{})
+		template["description"] = r.Description
+		template["display_name"] = r.DisplayName
+		template["object_id"] = r.ID()
```
manicminer, line 93:
```suggestion
		template["object_id"] = r.ID
```

manicminer:
@MattGarnerAWR Just a couple things to note:

- New resources / data sources need to be registered, typically there's a `registration.go` file in each package for this
- Whenever we dereference a pointer, we need to first check that it's non-nil or it may cause a crash at runtime

I fixed these up in the PR as I didn't spot them earlier on, hope you don't mind. Thanks again for the contribution!

MattGarnerAWR:
> @MattGarnerAWR Just a couple things to note:
> 
> * New resources / data sources need to be registered, typically there's a `registration.go` file in each package for this
> * Whenever we dereference a pointer, we need to first check that it's non-nil or it may cause a crash at runtime
> 
> I fixed these up in the PR as I didn't spot them earlier on, hope you don't mind. Thanks again for the contribution!

Thanks for your help!

PR #1168, opened 2023-08-07
`azuread_conditional_access_policy` set `built_in_controls` to optional

alexwilcox9:
During some recent work I noticed you could not create a CA policy in terraform that only has a terms of use set and no other controls.

This is possible through the portal, the below policy is valid and with this modification applies successfully.

```hcl
resource "azuread_conditional_access_policy" "tou" {
  display_name = "Require All Users agree to terms"
  state        = "disabled"

  conditions {
    client_app_types = ["all"]

    applications {
      included_applications = ["All"]
    }

    users {
      included_users = ["All"]
    }

    locations {
      included_locations = ["All"]
    }
  }

  grant_controls {
    operator                      = "OR"
    terms_of_use                  = ["<<TOU-GUID>>"]
  }
}
```

Creating a policy with neither field is not possible (at least not until authentication strengths are supported) so I have set it to at least one of `built_in_controls` or `terms_of_use`

### DIFF
--- a/docs/resources/conditional_access_policy.md
+++ b/docs/resources/conditional_access_policy.md
@@ -227,10 +227,12 @@ The following arguments are supported:
 
 `grant_controls` block supports the following:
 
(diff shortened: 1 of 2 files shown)


nauxliu:
Can we make the grant_controls completely optional? There are cases in which we only want to apply session_controls, and this is also possible through the portal.

manicminer (approved):
Thanks for this @alexwilcox9, LGTM 👍

PR #1173, opened 2023-08-21
data.azuread_group: allow lookup by mail_nickname

nobbs:
Adding the option to look up `data.azuread_group` items by `mail_nickname` field.

The need for this arose as we ran into the issue with display name not necessarily being unique in a larger AAD instance, thus making it impossible to pick the data for such a group from AAD without resorting to the object id. This isn't really possible in our case, as obtaining the object UUID from the group name is the reason we're using this data source in the first place 🙈 .

I've also added tests based on the ones already present for the `displayName` option but was not able to run the acceptance tests due to missing a proper Azure subscription satisfying the requirements to run those.

### DIFF
--- a/docs/data-sources/group.md
+++ b/docs/data-sources/group.md
@@ -28,11 +28,12 @@ data "azuread_group" "example" {
 The following arguments are supported:
 
 * `display_name` - (Optional) The display name for the group.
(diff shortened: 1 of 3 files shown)


manicminer (approved):
Thank you @nobbs for submitting this addition to the provider. This looks great and the test is passing.

Screenshot 2023-09-15 at 00 17 07

PR #1189, opened 2023-09-13
Allow for setting client ID / secret through files

mpminardi:
Adds the ability to specify a path to the client ID and client secret in place of supplying their values directly in the provider.

This adds matching capabilities to what was added to the AzureRM provider in https://github.com/hashicorp/terraform-provider-azurerm/pull/21764

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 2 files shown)


manicminer (approved):
Thanks for adding this @mpminardi, this LGTM 👍 

Test results

Screenshot 2023-09-14 at 23 58 41

Issue #1172, opened 2023-08-18
`azuread_group_member` resource fails to refresh after the ID of the group has changed


LoicBer:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

* terraform 1.5.5, terraform 1.2.9
* azuread 2.39, azuread 2.41

### Affected Resource(s)

* `azuread_group_member`

### Terraform Configuration Files

Two separate modules, with **distinct tfstates**:

*  module1: AAD group using `azuread_group` resource

*  module2: Member of the new AAD group using `azuread_group_member` resource and the `azuread_group` data source 

#### module1: group
```hcl
data "azuread_client_config" "current" {}

resource "azuread_group" "group" {
  display_name     = "sample-group"
  owners           = [data.azuread_client_config.current.object_id]
  security_enabled = true
}
```

#### module2: group member

```hcl
data "azuread_client_config" "current" {}

data "azuread_group" "external_group" {
  display_name = "sample-group"
  security_enabled = true
}

resource "azuread_group_member" "group_member" {
  group_object_id  = data.azuread_group.external_group.id
  member_object_id = data.azuread_client_config.current.object_id
}
```

### Debug Output

After the following sequence:

1. `terraform apply` on module1 (create new AAD group)
2. `terraform apply` on module2 (create new member of the new AAD group)
3. `terraform destroy` and `terraform apply` on module1 (create a new ID for the AAD group)
4. `terraform apply` on module2 (reapply to refresh the new group member)

Terraform apply on module2 gives:

```
$ terraform apply
data.azuread_client_config.current: Reading...
data.azuread_group.external_group: Reading...
data.azuread_client_config.current: Read complete after 0s [id=*********]
data.azuread_group.external_group: Read complete after 1s [id=3d**********]
azuread_group_member.group_member: Refreshing state... [id=4c***********]
╷
│ Error: Retrieving members for group with object ID: "4c**************"
│
│   with azuread_group_member.group_member,
│   on main.tf line 9, in resource "azuread_group_member" "group_member":
│    9: resource "azuread_group_member" "group_member" {
[...]

PR #1198, opened 2023-09-22
Fix group member resource to survive refresh when group is missing

LoicBer:
This PR proposes a fix for issue #1172 regarding `azuread_group_member` resource failing to refresh  when the referenced group is missing.

It handles the case where `client.ListMembers()` returns a 404 error (missing group) by removing the group member from the state.

I could validate with [this simple bug reproducer](https://github.com/LoicBer/azuread-groupmember-bug) that this change solves the issue.

Here the output I get when running `terraform plan` when the group ID referenced by the group member has changed:
https://gist.github.com/LoicBer/53a8b974067d0ca0cfacb50f9a9ee911

Terraform wants to remove the member from the state and recreate the member with the new group ID, which is the expected behavior.

### DIFF
--- a/internal/services/groups/group_member_resource.go
+++ b/internal/services/groups/group_member_resource.go
@@ -126,8 +126,13 @@ func groupMemberResourceRead(ctx context.Context, d *schema.ResourceData, meta i
 		return tf.ErrorDiagPathF(err, "id", "Parsing Group Member ID %q", d.Id())
 	}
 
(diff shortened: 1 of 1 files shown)


manicminer (approved):
Hi @LoicBer, thanks for submitting this fix and taking the time to demo the use case that triggers this bug. We should definitely be gracefully removing the member resource from state in this case.

LGTM 👍

Issue #1218, opened 2023-10-23
/applications/ is appended to application_object_id, forcing replacement


gpatounas:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Terraform v1.6.2
hashicorp/azuread v2.44.0

### Affected Resource(s)

observed:
azuread_application_password
azuread_application_certificate

potentially any resource with the `application_object_id` argument

### Terraform Configuration Files

```hcl
resource "azuread_application_password" "sample" {
  application_object_id = azuread_application.sample.id
  ...
}
```

### Expected Behavior

No changes were expected in the plan which was triggered as part of automatically upgrading the `azuread` provider from `v2.43.0` to `v2.44.0`

### Actual Behavior

The given plan would replace all `azuread_application_password` and `azuread_application_certificate` resources due to the changes observed in the `application_object_id` argument e.g.:

```hcl
application_object_id = "00000000-0000-0000-0000-000000000000" -> "/applications/00000000-0000-0000-0000-000000000000"
```

### Steps to Reproduce

Upgrade existing configuration which includes resources with the `application_object_id` argument to `azuread` `v2.44.0`

beyondbill:
Try 

```
resource "azuread_application_password" "sample" {
  application_object_id = azuread_application.sample.object_id
  ...
}
```

manicminer:
A fix for this has been released in v2.44.1, please upgrade!

PR #1221, opened 2023-10-23
Applications: DiffSuppressFunc for `application_object_id`

manicminer:
This works around some existing configurations where a diff may arise as a result of the recent state migration to update the resource ID for `azuread_application` to use a typed ID.

Where an existing configuration mistakenly references the `id` attribute for an `azuread_application` resource, expecting it to contain a bare object ID, this avoids a subsequent diff where the existing bare UUID and the prospective ID point to the same application.

Fixes: #1218


manicminer:
Test results

Screenshot 2023-10-23 at 21 05 45

katbyte (approved):
LGTM ⚕️

Issue #1230, opened 2023-10-27
Fix import example in documentation for azuread_application


joonvena:
Currently documentation states that the resource should be imported using pattern:
```
terraform import azuread_application.example 00000000-0000-0000-0000-000000000000
```

It should be instead:
```
terraform import azuread_application.example /applications/00000000-0000-0000-0000-000000000000
```

dnisbet:
I get
` Error: parsing Resource ID "/applications/00000000-0000-0000-0000-000000000000": specified ID ("/applications/00000000-0000-0000-0000-000000000000") is not valid: uuid string is wrong length`

But works without the /applications/ path.

manicminer:
@dnisbet Please ensure you are using the latest version of the provider

PR #1231, opened 2023-10-27
fix: azuread_application resource import example

joonvena:
Fixes: #1230 

Currently documentation states that the resource should be imported using pattern:
```
terraform import azuread_application.example 00000000-0000-0000-0000-000000000000
```

It should be instead:
```
terraform import azuread_application.example /applications/00000000-0000-0000-0000-000000000000
```

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -343,5 +343,5 @@ In addition to all arguments above, the following attributes are exported:
 Applications can be imported using their object ID, e.g.
 
 ```shell
-terraform import azuread_application.example 00000000-0000-0000-0000-000000000000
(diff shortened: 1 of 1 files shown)


docs/resources/application.md
```diff
@@ -343,5 +343,5 @@ In addition to all arguments above, the following attributes are exported:
 Applications can be imported using their object ID, e.g.
```
manicminer, line 343:
```suggestion
Applications can be imported using their object ID, in the following format.
```

manicminer (approved):
Thanks for this @joonvena, this LGTM!

Issue #1216, opened 2023-10-20
Update Authentication Strength Documentation to Include allowed combinations


zach-griffin:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently, the documentation for the new Authentication Strengths resource has only fido2 and password as options in the example. We should list all available options in the documentation since the names do not 1:1 match with the options in Azure. I had to dig in the go test case to find the possible options.

We should describe the available options in the documentation, like its done on https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/conditional_access_policy#excluded_platforms

Thank you!

### References

https://github.com/hashicorp/terraform-provider-azuread/blob/d25b6df7b9e6052b6edc1475b902a66724a0be20/internal/services/conditionalaccess/authentication_strength_policy_resource_test.go#L111C3-L133C4

* #0000

manicminer:
Fixed by #1242

PR #1242, opened 2023-11-03
Issue: 1216 - Updated documentation to include second example

Threpio:
Incldued a second example which includes all of the possible allowed_combinations.

Covers Issue #1216


manicminer (approved):
Thanks @Threpio, this LGTM 👍 

I've taken the opportunity to move this resource to the correct service package - this is an internal refactor which doesn't affect practitioners.

Test results for completeness:

Screenshot 2023-11-15 at 13 03 26

Screenshot 2023-11-15 at 13 04 43

PR #1244, opened 2023-11-05
Fix example for azuread_application_api_access

StephenKing:
[Example](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/application_api_access#example-usage) refers to created resource, but should use the declared data resource instead.


manicminer (approved):
Thanks for this docs fix @StephenKing, this LGTM 👍

Issue #1259, opened 2023-11-24
Error: `mail_nickname` is required for mail-enabled groups


DeviaVir:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

* Terraform v1.6.4
* AzureAD v2.46.0

### Affected Resource(s)

* `azuread_group`

### Terraform Configuration Files

```hcl
resource "azuread_group" "admin_group" {
  display_name     = "test"
  mail_enabled     = true
  mail_nickname    = "testGroup"
  security_enabled = true
  types            = ["Unified"]
}
```

### Debug Output

Plan failed:

```
Error: `mail_nickname` is required for mail-enabled groups
```

### Panic Output

N/A

### Expected Behavior

Expected plan to succeed and show it was going to create the AzureAD group.

### Steps to Reproduce

1. `terraform plan`

### Important Factoids

N/A

### References

* [group_resource.go relevant logic](https://github.com/hashicorp/terraform-provider-azuread/blob/main/internal/services/groups/group_resource.go#L390) - could it be that the diff produces no result because there's no value yet if the resource hasn't been created yet?

manicminer:
@DeviaVir Thanks for reporting this, and for opening [the related PR](https://github.com/hashicorp/terraform-provider-azuread/pull/1260). Unfortunately as this is a required field for mail-enabled groups, we cannot make this optional. I've explained in more detail in my comment on your PR, but you might want to look at the [`ignore_changes` lifecycle meta-argument](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes).

Since there's nothing for us to do here, as we match the API behavior in our validation, I'm going to close out this issue for now. Thanks!

DeviaVir:
@manicminer up to you, but I am not making the field optional. Rather, there seems to be an issue with detecting a diff, while the only reason we are looking to compare is to figure out if it's empty and then return that error. In other places of the logic, we are just setting a default value. My proposal here is to if the mail_nickname is empty, is to just generate it for the user.

PR #1260, opened 2023-11-24
Bugfix[#1259]: remove mail nickname check

DeviaVir:
We are already always setting a default mail nickname, the check seems to fall apart for not-yet-existing groups, but the check itself doesn't make sense since we are just setting a default if none exists.

### DIFF
--- a/internal/services/groups/group_resource.go
+++ b/internal/services/groups/group_resource.go
@@ -386,10 +386,6 @@ func groupResourceCustomizeDiff(ctx context.Context, diff *pluginsdk.ResourceDif
 		return fmt.Errorf("`mail_enabled` must be true for unified groups")
 	}
 
-	if mailNickname := diff.Get("mail_nickname").(string); mailEnabled && mailNickname == "" {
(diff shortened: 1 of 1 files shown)


manicminer:
@DeviaVir Thanks for suggesting this change. Unfortunately, `mailNickname` is a required property for mail-enabled groups as per validation in the API. When creating a security-only group, we mimic the portal behavior by generating a pseudorandom value for `mailNickname`, but since this property is immutable once set, we don't do the same for mail-enabled groups as this would likely defeat the intent of the group.

Screenshot 2023-11-27 at 16 35 49


If you are managing existing groups with Terraform, you will need to import the group into your statefile, and then ensure the correct properties are set in your configuration. This will include the `mail_nickname` property.

I appreciate the time taken to test and submit this PR, however due to the above we will unfortunately not be able to relax the validation here as it does mirror the API. If you would like to manage groups without managing the `mail_nickname` property, I recommend looking at the [`ignore_changes` lifecycle meta-argument](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes).

Accordingly, I'm going to close this PR for now. Thanks again!

PR #1266, opened 2023-12-01
docs: update docs to show service principal login is now supported

jaredfholgate:
This is a documentation only update to provide details on how to use az cli --service-principal and az cli --identity login since support was added in 2.35.0.

### DIFF
--- a/docs/guides/azure_cli.md
+++ b/docs/guides/azure_cli.md
@@ -20,7 +20,7 @@ We recommend using either a Service Principal or Managed Identity when running T
 ## Important Notes about Authenticating using the Azure CLI
 
 * Terraform only supports authenticating using the `az` CLI (and this must be available on your PATH) - authenticating using the older `azure` CLI or PowerShell Az / AzureRM Cmdlets is not supported.
-* Prior to version 2.35, authenticating via the Azure CLI was only supported when using a User Account. For example `az login --service-principal` was not supported and you had to use either a [Client Secret](service_principal_client_secret.html) or a [Client Certificate](service_principal_client_certificate.html). From 2.35 upwards, authenticating via the Azure CLI is supported when using a Service Principal or Managed Identity.
+* Prior to version 2.35, authenticating via the Azure CLI was only supported when using a User Account. For example `az login --service-principal` was not supported and it was necessary to use either a [Client Secret](service_principal_client_secret.html) or a [Client Certificate](service_principal_client_certificate.html). From 2.35 upwards, authenticating via the Azure CLI is supported when using a Service Principal or Managed Identity. However, we still recommend using native provider support for Service Principal or Managed Identity authentication wherever possible.
 
 ---
 
@@ -41,19 +41,30 @@ az login --allow-no-subscriptions
 Service Principal with a Secret:
 
 ```shell
-az login --service-principal -u "CLIENT_ID" -p "CLIENT_SECRET" --tenant "TENANT_ID" --allow-no-subscriptions
+az login --service-principal \
+         --username 00000000-0000-0000-0000-000000000000 \
+         --password "MyCl1eNtSeCr3t" \
+         --tenant 10000000-2000-3000-4000-500000000000 \
+         --allow-no-subscriptions
 ```
 
 Service Principal with a Certificate:
 
 ```shell
-az login --service-principal -u "CLIENT_ID" -p "CERTIFICATE_PEM" --tenant "TENANT_ID" --allow-no-subscriptions
+az login --service-principal \
+         --username 00000000-0000-0000-0000-000000000000 \
+         --password /path/to/certificate \
+         --tenant 10000000-2000-3000-4000-500000000000 \
+         --allow-no-subscriptions
 ```
 
 Service Principal with Open ID Connect (for use in CI / CD):
 
 ```shell
-az login --service-principal -u "CLIENT_ID" --tenant "TENANT_ID" --allow-no-subscriptions
+az login --service-principal \
+         --username 00000000-0000-0000-0000-000000000000 \
+         --tenant 10000000-2000-3000-4000-500000000000 \
+         --allow-no-subscriptions
 ```
 
 Managed Identity:
@@ -63,7 +74,9 @@ az login --identity --allow-no-subscriptions
 
 or
 
-az login --identity --username "CLIENT_ID" --allow-no-subscriptions
+az login --identity \
+         --username 00000000-0000-0000-0000-000000000000 \
+         --allow-no-subscriptions
 ```
 
 The `--allow-no-subscriptions` argument enables access to tenants that have no linked subscriptions, in addition to tenants that do.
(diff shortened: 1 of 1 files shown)


manicminer (changes requested):
Hi @jaredfholgate, thanks for suggesting this change. Our language choice here was intentional; whilst authenticating az-cli with a service principal can and does work, we don't officially support this. Historically, az-cli authentication with a service principal did not work due to limitations of the tool, but with SDK advancements in the provider we have been able to permit this more recently. However, we do still recommend that practitioners use the native support for service principals, as it is more robust, more performant, and doesn't rely on a third party tool. So whilst we don't explicitly prevent authenticating az-cli in this way, we also can not strictly guarantee that this will always work as capabilities in the provider continue to progress.

I think we would be open to amending the docs to better represent this position, if you'd like to have a go at that? It would not need to include the technical context above, but could indicate the essential message that it might work but it's not supported in production and we recommend using the native authentication support.

jaredfholgate:
> Hi @jaredfholgate Jared Holgate FTE, thanks for suggesting this change. Our language choice here was intentional; whilst authenticating az-cli with a service principal can and does work, we don't officially support this. Historically, az-cli authentication with a service principal did not work due to limitations of the tool, but with SDK advancements in the provider we have been able to permit this more recently. However, we do still recommend that practitioners use the native support for service principals, as it is more robust, more performant, and doesn't rely on a third party tool. So whilst we don't explicitly prevent authenticating az-cli in this way, we also can not strictly guarantee that this will always work as capabilities in the provider continue to progress.
> 
> I think we would be open to amending the docs to better represent this position, if you'd like to have a go at that? It would not need to include the technical context above, but could indicate the essential message that it might work but it's not supported in production and we recommend using the native authentication support.

Hi @manicminer. Thanks for the feedback. Just to provide some insight into the reason for raising this change. When using Workload identity federation (OIDC) auth with Azure DevOps (and possibly GitHub and other tools), the id token is limited to 10 minutes. In some scenarios this can cause issues as even though Terraform exchanges the id token for an access token, it can try to do this more than once. The specific example being `terraform test`, where tests are executed sequentially and each test initialises the provider. If the tests last longer than 10 minutes, the run fails. However, if we use az-cli auth, this limitation is removed since the access token is cached by az-cli instead and the `azurerm`, `azuread` or `azapi` provider does not attempt to get a new access token.

I will look to update the wording as suggested, but could be worth a separate discussion. The Azure DevOps product group may be able to support the relevant testing required to make this a supported option on all three providers. It would also be great if the backend in terraform core supported this auth too (see here: https://github.com/hashicorp/terraform/issues/34322), although less important for the `terraform test` use case.

manicminer:
@jaredfholgate Thanks for the additional info. If we're failing due to short ID token validity/expiry, then I'd suggest that's something we should definitely look to improve.

In terms of updating our docs though, I would not want to specifically call out Azure DevOps as a scenario where users might want to use az-cli instead, without first trying to improve the native support.

docs/guides/azure_cli.md
```diff
@@ -20,22 +20,56 @@ We recommend using either a Service Principal or Managed Identity when running T
 ## Important Notes about Authenticating using the Azure CLI
 
 * Terraform only supports authenticating using the `az` CLI (and this must be available on your PATH) - authenticating using the older `azure` CLI or PowerShell Az / AzureRM Cmdlets is not supported.
-* Authenticating via the Azure CLI is only supported when using a User Account. If you're using a Service Principal (for example via `az login --service-principal`) you should instead authenticate via the Service Principal directly, either using a [Client Certificate](service_principal_client_certificate.html) or a [Client Secret](service_principal_client_secret.html).
+* Prior to version 2.35, authenticating via the Azure CLI was only supported when using a User Account. For example `az login --service-principal` was not supported and you had to use either a [Client Secret](service_principal_client_secret.html) or a [Client Certificate](service_principal_client_certificate.html). From 2.35 upwards, authenticating via the Azure CLI is supported when using a Service Principal or Managed Identity.
```
manicminer, line 23:
```suggestion
* Prior to version 2.35, authenticating via the Azure CLI was only supported when using a User Account. For example `az login --service-principal` was not supported and it was necessary to use either a [Client Secret](service_principal_client_secret.html) or a [Client Certificate](service_principal_client_certificate.html). From 2.35 upwards, authenticating via the Azure CLI is supported when using a Service Principal or Managed Identity. However, we still recommend using native provider support for Service Principal or Managed Identity authentication wherever possible.
```

docs/guides/azure_cli.md
```diff
@@ -20,22 +20,56 @@ We recommend using either a Service Principal or Managed Identity when running T
...
+User Account:
+
+```shell
+az login --allow-no-subscriptions
+```
+
+Service Principal with a Secret:
+
+```shell
+az login --service-principal -u "CLIENT_ID" -p "CLIENT_SECRET" --tenant "TENANT_ID" --allow-no-subscriptions
```
manicminer, line 44:
```suggestion
az login --service-principal \
         --username 00000000-0000-0000-0000-000000000000 \
         --password "MyCl1eNtSeCr3t" \
         --tenant 10000000-2000-3000-4000-500000000000 \
         --allow-no-subscriptions
```

docs/guides/azure_cli.md
```diff
@@ -20,22 +20,56 @@ We recommend using either a Service Principal or Managed Identity when running T
...
+```
+
+Service Principal with a Certificate:
+
+```shell
+az login --service-principal -u "CLIENT_ID" -p "CERTIFICATE_PEM" --tenant "TENANT_ID" --allow-no-subscriptions
```
manicminer, line 50:
```suggestion
az login --service-principal \
         --username 00000000-0000-0000-0000-000000000000 \
         --password /path/to/certificate \
         --tenant 10000000-2000-3000-4000-500000000000 \
         --allow-no-subscriptions
```

docs/guides/azure_cli.md
```diff
@@ -20,22 +20,56 @@ We recommend using either a Service Principal or Managed Identity when running T
...
+```
+
+Service Principal with Open ID Connect (for use in CI / CD):
+
+```shell
+az login --service-principal -u "CLIENT_ID" --tenant "TENANT_ID" --allow-no-subscriptions
```
manicminer, line 56:
```suggestion
az login --service-principal \
         --username 00000000-0000-0000-0000-000000000000 \
         --tenant 10000000-2000-3000-4000-500000000000 \
         --allow-no-subscriptions
```

docs/guides/azure_cli.md
```diff
@@ -20,22 +20,56 @@ We recommend using either a Service Principal or Managed Identity when running T
...
+```
+
+Managed Identity:
+
+```shell
+az login --identity --allow-no-subscriptions
+
+or
+
+az login --identity --username "CLIENT_ID" --allow-no-subscriptions
```
manicminer, line 66:
```suggestion
az login --identity \
         --username 00000000-0000-0000-0000-000000000000 \
         --allow-no-subscriptions
```

manicminer (approved):
@jaredfholgate Thanks again for suggesting this documentation change. Due to the necessary use of az-cli in some environments, we now fully support authenticating as a service principal via az-cli. Due to this, I'd be happy to make these changes to the authentication guide. I've also expanded the examples a little using the same placeholder values that we use in other auth guides.

PR #1293, opened 2024-01-22
Include Example Of ARM_USE_OIDC

tgoodsell-tempus:
- In the `service_principal_oidc.md` document, it explains using ENV VARs for configuring the client, sub, and tenant ID env vars; as well as the use of the `OIDC` env var flavours available.
- Since the `USE_OIDC` option is also exposed as a env var, include this in the example. It makes sense to also use this when performing environment based configuration for OIDC access.


manicminer (approved):
Thanks @tgoodsell-tempus, this LGTM 👍

Issue #1296, opened 2024-01-25
azuread_application_pre_authorized does not properly destroy


KenSpur:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```
Terraform v1.7.1
on linux_amd64
+ provider registry.terraform.io/hashicorp/azuread v2.47.0
```

### Affected Resource(s)

* `azuread_application_pre_authorized`

### Expected Behavior

All specified applications should be removed from the pre-authorized list.

### Actual Behavior

The concurrent destruction operations interfere with each other, leading to a scenario where one application remains in the pre-authorized list. It appears that each operation updates the list without being aware of the other's modifications, resulting in one application erroneously remaining.

### Steps to Reproduce

clone [KenSpur/terraform-azuread-application-pre-authorized-destroy-issue](https://github.com/KenSpur/terraform-azuread-application-pre-authorized-destroy-issue)

```bash
# Login to Azure

az login

# Setup Applications

cd setup

terraform init
terraform apply

# Recreate Issue

cd ..

terraform init
terraform apply
terraform destroy
terraform apply

# Error: A resource with the ID "*/preAuthorizedApplication/*" already exists < will be thrown
```

KenSpur:
The issue seems to stem from concurrent operations both reading and writing to the pre-authorized applications array simultaneously. Each operation does not account for changes made by the other, leading to a race condition.

manicminer:
Fixed with #1299

PR #1299, opened 2024-01-26
Fix to issue: 1296 azuread_application_pre_authorized does not properly destroy

KenSpur:
#1296

### DIFF
--- a/internal/services/applications/application_pre_authorized_resource.go
+++ b/internal/services/applications/application_pre_authorized_resource.go
@@ -304,6 +304,9 @@ func applicationPreAuthorizedResourceDelete(ctx context.Context, d *pluginsdk.Re
 		return tf.ErrorDiagPathF(err, "id", "Parsing pre-authorized application ID %q", d.Id())
 	}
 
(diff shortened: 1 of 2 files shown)


KenSpur:
Ready for review!

Is `.github/labeler-pull-request-triage.yaml` something i should add?

manicminer (approved):
@KenSpur Thank you for submitting this PR, and apologies for the delay in reviewing.

We generally try to avoid using artificial delays as a way to combat concurrency or consistency issues. We have a locking mechanism in the provider which is intended to cover this kind of problem with parallel updates to the same underlying resource or object (i.e. an application in this case). This is present in the Create and Update functions for this resource, but it looks like I missed adding it to the Delete function.

```golang
tf.LockByName(applicationResourceName, id.ObjectId)
defer tf.UnlockByName(applicationResourceName, id.ObjectId)
```

Due to the tardiness of this review, I hope you don't mind but I've reworked this change using this approach, and added some checks to the acceptance test.

Thanks again for spotting this bug and working on the fix!

manicminer:
Test results

Screenshot 2024-05-08 at 23 17 12

Issue #881, opened 2022-09-07
Enable azuread_group to return indirect members


jakebracey:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently the azuread_group data source only returns direct members of the group. It would be helpful to be able to also get all transitive members of that group. I see that Microsoft Graph has the API https://graph.microsoft.com/v1.0/groups/GROUP_ID/transitiveMembers which could be leveraged for this. 

The existing data source could be updated as shown below or a new data source could be created. 

### New or Affected Resource(s)

* azuread_group (update exisiting)
* azuread_group_members (create separate)

### Potential Terraform Configuration

data "azuread_group" "example" {
  display_name     = "Group Name Here"
include_members = "transitive"
}
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

Threpio:
The current "groups List Members" method and client from Hamilton does not currently allow for this query to be passed through but would be relatively easy to implement:
[func (c *GroupsClient) ListMembers(ctx context.Context, id string) (*[]string, int, error) {](https://github.com/hashicorp/terraform-provider-azuread/blob/d8b9594ce8c53c1aa242035081969dce2043a9a3/vendor/github.com/manicminer/hamilton/msgraph/groups.go#L331)
(note the only params are the group id)

panic-kbutton:
Noting here that https://github.com/manicminer/hamilton/pull/191 has been merged, and transitive members can now be included in terraform's azuread_group data source. @Threpio please advise if a new issue should be opened for this functionality.

manicminer:
Woops, looks like this was closed prematurely :)

cran1um:
I'm just discovering this open issue, while searching for a solution to this problem.  At the moment, is there any way that this can be utilized in its current state?

TomasKunka:
I ended up publishing my own version...

```terraform
terraform {
  required_providers {
    azuread = {
      source  = "TomasKunka/azuread"
[...]

PR #1300, opened 2024-01-26
Feature: #881 Enable azuread_group to return indirect members

TomasKunka:
#881 
Please have mercy! My first Go code and my first open source contribution.


RichieBzzzt:
Hello @manicminer I'd also love to see this merged, is there a chance that it will be merged? Thanks in advance!

manicminer (approved):
Thanks for this contribution @TomasKunka! Apologies for the delay in reviewing the PR - this is an excellent first contribution. I made a couple of minor changes, which ordinarily I'd suggest in my review but given the tardiness of this review, and the diminutive changes needed, I thought it would be helpful to update your branch and note them here.

- We generally prefer to order the attributes in a schema starting with Required properties, followed by Optional ones, and lastly any Computed attributes - with loose secondary ordering by name. I moved the `include_transitive_members` property further up in the schema map, alongside the other Optional properties.
- Where you added an `if` statement to check the value of `include_transitive_members`, I moved the `var err error` statement outside of this `if` block so that it only needs to be declared once. Very much a minor optimization.
- I added the new property to the data source documentation page.

Thank you again for this addition, this LGTM 👍

manicminer:
Test results (failure is unrelated)

Screenshot 2024-05-08 at 20 15 55

TomasKunka:
Thank you for the kind words and thorough explanation @manicminer !

PR #1303, opened 2024-02-02
feat: allow data.azuread_application lookup using identifier_uri

JonasBak:
This PR introduces a small feature to the `azuread_application` data source that allows us to get the application based on an identifying URI. This is a feature we would really like to use. It looks like this:

```terraform
data "azuread_application" "example" {
  identifier_uris  = [
    "api://example-app",
  ]
}
```

I've implemented it in a way where you can provide a list of one element to the `identifier_uris` field, instead of introducing another field called `identifier_uri` that is just a string, as I thought it might be easy to mix up the two fields, but would love some input on this. I added an error if more than one element is provided in `identifier_uris`.

### DIFF
--- a/docs/data-sources/application.md
+++ b/docs/data-sources/application.md
@@ -33,7 +33,7 @@ The following arguments are supported:
 * `client_id` - (Optional) Specifies the Client ID of the application.
 * `display_name` - (Optional) Specifies the display name of the application.
 * `object_id` - (Optional) Specifies the Object ID of the application.
-* `identifier_uri` - (Optional) Specifies a identifying URI of the application. This should only be used for lookups, `identifier_uris` should be used to read identifier URIs.
+* `identifier_uri` - (Optional) Specifies any identifier URI of the application. See also the `identifier_uris` attribute which contains a list of all identifier URIs for the application.
 
 ~> One of `client_id`, `display_name`, `object_id`, or `identifier_uri` must be specified.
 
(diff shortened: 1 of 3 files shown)


manicminer (changes requested):
Hi @JonasBak, thanks for this contribution, this would be a useful addition to the data source! Apologies for the delayed review.

If we are going to match any one identifier URI, then I think it would be better to add a new property `identifier_uri` and leave the existing `identifier_uris` as a computed attribute. My thinking here is that the returned application could have >1 identifier URI, which we would be saving to state, and this would create a discrepency between configuration and state. There is possibly no consequence to this at the moment, but this may throw an error in future versions of the provider RPC protocol (which is how the provider communicates with Terraform).

If you could rework this to add this additional property whilst keeping the existing attribute as-is, then I think this should be good to merge. The remainder of the implementation looks good to me.

Apologies again for the tardiness of this review. If you aren't in a position to make this change, just let me know and I'll happily pick this up whilst preserving your original commits.

Thanks!

docs/data-sources/application.md
```diff
@@ -33,8 +33,9 @@ The following arguments are supported:
 * `client_id` - (Optional) Specifies the Client ID of the application.
 * `display_name` - (Optional) Specifies the display name of the application.
 * `object_id` - (Optional) Specifies the Object ID of the application.
+* `identifier_uri` - (Optional) Specifies a identifying URI of the application. This should only be used for lookups, `identifier_uris` should be used to read identifier URIs.
```
manicminer, line 36:
```suggestion
* `identifier_uri` - (Optional) Specifies any identifier URI of the application. See also the `identifier_uris` attribute which contains a list of all identifier URIs for the application.
```

manicminer (approved):
Thanks for updating @JonasBak, this LGTM! 👍

manicminer:
Test results

Screenshot 2024-05-15 at 23 57 37

PR #1328, opened 2024-03-08
Change validation of employeeType

khikita:
Change validation of employeeType
Checked for input of 64 characters or less


manicminer (approved):
THanks @khikita, this LGTM 👍

PR #1340, opened 2024-03-20
Update azuread_service_principal_password documentation to correct azuread_service_principal example

andrewluotechnologies:
The example is incorrect - azuread_service_principal only allows you to specify client_id not application_id


manicminer (approved):
Thanks @andrewluotechnologies, this LGTM 👍

PR #1357, opened 2024-04-18
Feature: add application filter to `azuread_conditional_access_policy`

### LINKED ISSUE #1318: application_filter param missing from azuread_conditional_access_policy applications condition
### Community Note

### Description

Currently the Entra ID **Conditional Access policy** can be created via terraform using this resource **azuread_conditional_access_policy**  (provider **azuread**) . Policies can be scoped to certain applications or user actions using these params

conditions->applications->
   included_applications, excluded_applications, included_user_actions

However for organisations having hundreds or thousands of apps, this is not a viable way of grouping applications for a certain policy.

Instead, the Azure Portal and also MS Graph REST API (beta and v1.0) provides an option of **application_filter** to apply the policy to certain applications based on the filter condition.

This is documented here https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessapplications?view=graph-rest-beta#properties. (Note that this is beta documentation, but even v1.0 supports this) . See the **applicationFilter** param option.

**It would be great to have terraform **azuread** provider also support this.**

This allows organisations to apply specific policies to a group of Applications based on certain security attributes or any other attributes. (One example being -  using the **Custom Security Attributes** within **Entra ID**)

* **azuread_conditional_access_policy**

As a result of this feature the conditions->applications will have a new "**application_filter**" option
```
resource "azuread_conditional_access_policy" "example" {
  display_name = "example-policy"
  state        = "enabled"

  conditions {
    client_app_types    = ["All"]

    applications {
      application_filter {
        rule = "CustomSecurityAttribute.AppSecAttributeSet_AppSecurityStatus -eq \"CriticalApp\""
      }
    }

    users {
      included_groups = ["All"]
    }
  }

  grant_controls {
    operator          = "OR"
    built_in_controls = ["mfa"]
  }

}
```

Thus in this case the policy will apply to all Entra ID Apps which have a Custom Security Attribute (AppSecurityStatus = CriticalApp ), within the AppSecAttributeSet (attribute set).

### References

https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessapplications?view=graph-rest-beta#properties

https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessfilter?view=graph-rest-beta

https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/conditional_access_policy#applications

* #0000

BrendanThompson:
## Description

This PR adds the ability to have an application filter on `azuread_conditional_access_policy`.

```
resource "azuread_conditional_access_policy" "this" {
  display_name = "access-policy"
  state        = "disabled"

  conditions {
    client_app_types    = ["other"]

    applications {
      filter {
        mode = "include"
        rule = "CustomSecurityAttribute.AttributeName -eq \"AttributeValue\""
      }
    }

    users {
      included_roles = ["Contributor"]
    }
  }
}
```

## Change Log

- `azuread_conditional_access_policy` – add the `filter` attribute to the `applications` block allowing for applications to be filtered [GH-1357]

## Related Issue(s)

Fixes #1318


BrendanThompson:
@tombuildsstuff — don't suppose you've time to give some feedback on this PR as well?

manicminer (changes requested):
Thanks @TomasKunka for this addition.

This is looking great - if you could add an acceptance test case that uses a filter for an application condition then this should be good to merge. Conditional Access Policies have shown to yield odd payload parsing bugs further down the line so having coverage is necessary to give us a good shot at spotting regressions.

Thanks!

BrendanThompson:
@manicminer thanks for reviewing this, I will try and get an acceptance test in today 😃

manicminer:
Great thanks, and sorry for the mis-tag! 😊

BrendanThompson:
@manicminer — sorry it's taken me a few days to get back to you. I started writing the tests then I realised that we will actually require some attributes to be set on wherever the Acceptance Tests are run, otherwise the tests will fail. There is currently also no mechanism to deploy said attributes with Terraform ☹️.

https://learn.microsoft.com/en-us/entra/fundamentals/custom-security-attributes-overview

The above link are the attributes we need in-place. How would you like to proceed on this, if possible were you able to add an attribute and I can test with that. Then perhaps the next thing is to work on integrating the attribute creation into the provider? (Happy to work in this)

BrendanThompson:
I cannot think of any way to test it properly without the custom attributes being there to be honest. And testing it without an attribute would yield odd results due to the API. We could add a notice on the resource itself, like a warning around it. _Buyer beware_ until the security attributes are available. What are your thoughts?

manicminer:
I would also love to ship this but unless we are able to demonstrate some level of testing, we unfortunately won't be able to add a property or block to a resource without it, which I guess means this will have to wait until we have a resource to support custom security attributes.

BrendanThompson:
Hey @manicminer,

I have opened a PR in `Hamilton` (https://github.com/manicminer/hamilton/pull/281) to add support for both `attribute sets` and `custom security attribute definitions` once that has gone through I will add support for them in the provider and update the tests here 😃

manicminer:
@BrendanThompson The hamilton version has been updated in the provider, hopefully this unblocks your work on this 🙂

jamie-oconnell:
Are there any current blockers to merge this? Happy holidays!

bitcarousel:
It seems to be almost a year ago since Feature created. Will this make it into a release soon would be great?

jackofallops (changes requested):
Hi @BrendanThompson - Not sure exactly what's happened with this PR, but can you rebase on latest main to see if we can get rid of all the unrelated "touched" files - we can't review/accept this PR as it stands with ~90 unrelated files modified in it. Sorry.

Thanks

bitcarousel:
Will this ever be merged into main?
I have honestly been waiting almost two years to finally be able to use application filters on Conditional Access policies with the azuread provider. I don’t think this is an edge case anymore.

bubbletroubles:
I think #1780 might be more appropriate to merge, its cleaner from a git history perspective and does the same thing. @jackofallops  can someone take a look?

sreallymatt:
superseded by #1780

PR #1362, opened 2024-04-29
docs: add to resource_app_id description and fix typos in service_principals.md

nayrsirhc:
# docs: add to resource_app_id description and fix typos in service_principals.md

I was recently using this provider and had some confusion from my side as I was unaware client id and application id were synonomous for one another, I use the terraform provider docs religiously and I believe that because application id is a property that is now deperecated this description should be changed to match what the provider expects, which from what I can tell is client id

### DIFF
--- a/docs/data-sources/service_principals.md
+++ b/docs/data-sources/service_principals.md
@@ -27,7 +27,7 @@ data "azuread_service_principals" "example" {
 }
 ```
 
-*Look up by application IDs (client IDs*
+*Look up by application IDs (client IDs)
 
 ```terraform
 data "azuread_service_principals" "example" {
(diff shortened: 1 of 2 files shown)


docs/data-sources/service_principals.md
```diff
@@ -27,7 +27,7 @@ data "azuread_service_principals" "example" {
 }
 ```
 
-*Look up by application IDs (client IDs*
+*Look up by application IDs (client IDs)
```
manicminer, line 30:
```suggestion
*Look up by application IDs (client IDs)*
```

manicminer (approved):
Thanks @nayrsirhc, aside from a slight formatting issue, this LGTM!

Please feel free to continue suggesting improvements to the provider docs, particularly around Applications if we have missed something or we can be more explicit.

nayrsirhc:
Its a pleasure and thank you @manicminer!

PR #1373, opened 2024-05-10
Fix example code for azuread_group_role_management_policy

audunsolemdal:


manicminer (approved):
Thanks for these fixes @audunsolemdal, this LGTM 👍

PR #1389, opened 2024-05-23
azuread_application: add password property

HappyTobi:
Add new inline property "app_password"

The app_password generate a application credential with the first request where the application is created.
The improvement here is that ist a bit faster than the external "application_password_resource" creation because there is no wait time until the credential can be used.

The feature set of the "app_password" is equal to "application_password_resource"

Regards


internal/helpers/credentials.go
```diff
@@ -185,6 +185,77 @@ func KeyCredentialForResource(d *pluginsdk.ResourceData) (*msgraph.KeyCredential
...
+	return []map[string]interface{}{{
+		"display_name":      exist["display_name"],
+		"start_date":        startDate,
+		"end_date":          endDate,
+		"value":             value,
+		"key_id":            keyId,
+		"end_date_relative": exist["end_date_relative"],
+	}}
+}
+
```
manicminer, line 222: This function will not be needed if we move this logic into the resource

```suggestion
```

internal/services/applications/application_resource.go
```diff
@@ -266,6 +266,58 @@ func applicationResource() *pluginsdk.Resource {
...
+							ValidateFunc: validation.IsRFC3339Time,
+						},
+
+						"end_date_relative": {
+							Description:  "A relative duration for which the password is valid until, for example `240h` (10 days) or `2400h30m`. Changing this field forces a new resource to be created",
+							Type:         pluginsdk.TypeString,
+							Optional:     true,
+							ValidateFunc: validation.StringIsNotEmpty,
+						},
+
```
manicminer, line 304: I think we should remove `end_date_relative`, since users can use the `timeadd()` Terraform function to compute this, and it is problematic when persisting state in cases where the `end_date` value is changed and a password has to be replaced.

```suggestion
```

internal/services/applications/application_resource.go
```diff
@@ -266,6 +266,58 @@ func applicationResource() *pluginsdk.Resource {
 				},
 			},
 
+			"password": {
```
manicminer, line 269: Can we move this block further down (after `owners`) to preserve ordering?

internal/services/applications/application_resource_test.go
```diff
@@ -33,6 +33,7 @@ func TestAccApplication_basic(t *testing.T) {
 				check.That(data.ResourceName).Key("application_id").Exists(),
 				check.That(data.ResourceName).Key("client_id").Exists(),
 				check.That(data.ResourceName).Key("object_id").Exists(),
+				check.That(data.ResourceName).Key("app_password").DoesNotExist(),
```
manicminer, line 36: Should be `password`

```suggestion
				check.That(data.ResourceName).Key("password").DoesNotExist(),
```

internal/services/applications/application_resource_test.go
```diff
@@ -594,6 +595,27 @@ func TestAccApplication_logo(t *testing.T) {
...
+			Config: r.password(data),
+			Check: acceptance.ComposeTestCheckFunc(
+				check.That(data.ResourceName).ExistsInAzure(r),
+				check.That(data.ResourceName).Key("app_password.#").HasValue("1"),
+				check.That(data.ResourceName).Key("app_password.0.key_id").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.value").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.start_date").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.end_date").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.end_data_relative").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.display_name").HasValue(data.ResourceName),
```
manicminer, line 613: These should be `password`, and we can remove the check for `end_date_relative`. We can also set a unique display name for the password for good measure.

```suggestion
				check.That(data.ResourceName).ExistsInAzure(r),
				check.That(data.ResourceName).Key("password.#").HasValue("1"),
				check.That(data.ResourceName).Key("password.0.key_id").Exists(),
				check.That(data.ResourceName).Key("password.0.value").Exists(),
				check.That(data.ResourceName).Key("password.0.start_date").Exists(),
				check.That(data.ResourceName).Key("password.0.end_date").Exists(),
				check.That(data.ResourceName).Key("password.0.display_name").HasValue(fmt.Sprintf("acctest-appPassword-%d", data.RandomInteger)),
```

internal/services/applications/application_resource_test.go
```diff
@@ -1572,3 +1594,20 @@ resource "azuread_application" "test" {
...
+provider "azuread" {}
+
+data "azuread_client_config" "current" {}
+
+resource "azuread_application" "test" {
+  display_name = "acctest-APP-%[1]d"
+  owners = [ data.azuread_client_config.current.object_id ]
+
+  password {
+    display_name = "acctest-APP-%[1]d"
```
manicminer, line 1609: (display name change as suggested above)

```suggestion
    display_name = "acctest-appPassword-%[1]d"
```

internal/services/applications/application_resource_test.go
```diff
@@ -594,6 +595,27 @@ func TestAccApplication_logo(t *testing.T) {
...
+				check.That(data.ResourceName).Key("app_password.0.value").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.start_date").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.end_date").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.end_data_relative").Exists(),
+				check.That(data.ResourceName).Key("app_password.0.display_name").HasValue(data.ResourceName),
+			),
+		},
+	})
+}
+
```
manicminer, line 618: We should test Update functionality with an additional test

```suggestion

func TestAccApplication_passwordUpdate(t *testing.T) {
	data := acceptance.BuildTestData(t, "azuread_application", "test")
	startDate := time.Now().AddDate(0, 0, 7).UTC().Format(time.RFC3339)
	endDate := time.Now().AddDate(0, 5, 27).UTC().Format(time.RFC3339)
	updateEndDate := time.Now().AddDate(0, 11, 28).UTC().Format(time.RFC3339)
	r := ApplicationResource{}

	data.ResourceTest(t, r, []acceptance.TestStep{
		{
			Config: r.passwordComplete(data, startDate, endDate),
			Check: acceptance.ComposeTestCheckFunc(
				check.That(data.ResourceName).ExistsInAzure(r),
				check.That(data.ResourceName).Key("password.#").HasValue("1"),
				check.That(data.ResourceName).Key("password.0.key_id").Exists(),
				check.That(data.ResourceName).Key("password.0.value").Exists(),
				check.That(data.ResourceName).Key("password.0.start_date").Exists(),
				check.That(data.ResourceName).Key("password.0.end_date").Exists(),
				check.That(data.ResourceName).Key("password.0.display_name").HasValue(fmt.Sprintf("acctest-appPasswordComplete-%d", data.RandomInteger)),
			),
		},
		{
			Config: r.passwordComplete(data, startDate, updateEndDate),
			Check: acceptance.ComposeTestCheckFunc(
				check.That(data.ResourceName).ExistsInAzure(r),
				check.That(data.ResourceName).Key("password.#").HasValue("1"),
				check.That(data.ResourceName).Key("password.0.key_id").Exists(),
				check.That(data.ResourceName).Key("password.0.value").Exists(),
				check.That(data.ResourceName).Key("password.0.start_date").Exists(),
				check.That(data.ResourceName).Key("password.0.end_date").Exists(),
				check.That(data.ResourceName).Key("password.0.display_name").HasValue(fmt.Sprintf("acctest-appPasswordComplete-%d", data.RandomInteger)),
			),
		},
	})
}

```

internal/services/applications/application_resource_test.go
```diff
@@ -1572,3 +1594,20 @@ resource "azuread_application" "test" {
...
+resource "azuread_application" "test" {
+  display_name = "acctest-APP-%[1]d"
+  owners = [ data.azuread_client_config.current.object_id ]
+
+  password {
+    display_name = "acctest-APP-%[1]d"
+  }
+}
+`, data.RandomInteger)
+}
```
manicminer, line 1613: Configuration for the additional test

```suggestion
}

func (r ApplicationResource) passwordComplete(data acceptance.TestData, startDate, endDate string) string {
	return fmt.Sprintf(`
provider "azuread" {}

data "azuread_client_config" "current" {}

resource "azuread_application" "test" {
  display_name = "acctest-APP-%[1]d"
  owners       = [data.azuread_client_config.current.object_id]

  password {
    display_name = "acctest-appPasswordComplete-%[1]d"
    start_date   = "%[2]s"
    end_date     = "%[3]s"
  }
}
`, data.RandomInteger, startDate, endDate)
}
```

internal/services/applications/applications.go
```diff
@@ -793,6 +853,26 @@ func flattenApplicationSpa(in *msgraph.ApplicationSpa) []map[string]interface{}
...
+		}
+	}
+
+	// create case
+	for _, appCred := range *passwordCredentials {
+		return helpers.FlattenCredential(&appCred, d)
+	}
+
+	return []map[string]interface{}{}
+}
```
manicminer, line 874: By moving this logic into the resource for improved readability, we can simplify this function

```suggestion
func flattenApplicationPasswordCredentials(input *[]msgraph.PasswordCredential) (output []map[string]interface{}) {
	output = make([]map[string]interface{}, 0)

	if input == nil {
		return
	}

	for _, in := range *input {
		var startDate, endDate string

		if in.StartDateTime != nil {
			startDate = in.StartDateTime.Format(time.RFC3339)
		}
		if in.EndDateTime != nil {
			endDate = in.EndDateTime.Format(time.RFC3339)
		}

		output = append(output, map[string]interface{}{
			"key_id":       pointer.From(in.KeyId),
			"display_name": pointer.From(in.DisplayName),
			"start_date":   startDate,
			"end_date":     endDate,
			"value":        pointer.From(in.SecretText),
		})
	}

	return
}
```

internal/services/applications/application_resource_test.go
```diff
@@ -1572,3 +1594,20 @@ resource "azuread_application" "test" {
...
+
+func (r ApplicationResource) password(data acceptance.TestData) string {
+	return fmt.Sprintf(`
+provider "azuread" {}
+
+data "azuread_client_config" "current" {}
+
+resource "azuread_application" "test" {
+  display_name = "acctest-APP-%[1]d"
+  owners = [ data.azuread_client_config.current.object_id ]
```
manicminer, line 1606: (linting)

```suggestion
owners       = [data.azuread_client_config.current.object_id]
```

manicminer (changes requested):
Thanks for working on this @HappyTobi, this is looking good but there are some changes we will need to make to separate the logic for Create and Read, which I have added inline below.

Additionally, there is currently no Update implementation, so I've put together an initial suggestion with which the tests seem to pass. In the Update function, after setting the application logo, we can add something like this to ensure any password changes are effected. This is quite lengthy in order to handle eventual consistency issues.

```go
	// Remove and/or set a new application password, if changed
	if d.HasChange("password") {
		oldPasswordRaw, newPasswordRaw := d.GetChange("password")
		oldPasswordBlock := oldPasswordRaw.(*pluginsdk.Set).List()
		oldPassword := make(map[string]interface{})
		if len(oldPasswordBlock) > 0 {
			oldPassword = oldPasswordBlock[0].(map[string]interface{})
		}

		if oldPassword["key_id"] != nil {
			keyIdToRemove := oldPassword["key_id"].(string)
			if _, err = client.RemovePassword(ctx, id.ApplicationId, keyIdToRemove); err != nil {
				return tf.ErrorDiagF(err, "Removing password credential %q from application with object ID %q", id.ApplicationId, keyIdToRemove)
			}

			// Wait for application password to be deleted
			if err = helpers.WaitForDeletion(ctx, func(ctx context.Context) (*bool, error) {
				defer func() { client.BaseClient.DisableRetries = false }()
				client.BaseClient.DisableRetries = true

				app, _, err := client.Get(ctx, id.ApplicationId, odata.Query{})
				if err != nil {
					return nil, err
				}

				credential := helpers.GetPasswordCredential(app.PasswordCredentials, keyIdToRemove)
				if credential == nil {
					return pointer.To(false), nil
				}

				return pointer.To(true), nil
			}); err != nil {
				return tf.ErrorDiagF(err, "Waiting for deletion of password credential %q from application with object ID %q", keyIdToRemove, id.ApplicationId)
			}
		}

		newPasswordBlock := newPasswordRaw.(*pluginsdk.Set).List()
		if len(newPasswordBlock) > 1 {
			return tf.ErrorDiagPathF(errors.New("`password` must have one element"), "password", "Multiple passwords are not supported with this resource")
		}

		// Proceed to add a new password to replace the now-removed one, if the password block is present in the configuration
		if len(newPasswordBlock) > 0 {
			newPassword := newPasswordBlock[0].(map[string]interface{})

			credential, err := helpers.PasswordCredential(newPassword)
			if err != nil {
				attr := ""
				if kerr, ok := err.(helpers.CredentialError); ok {
					attr = kerr.Attr()
				}
				return tf.ErrorDiagPathF(err, attr, "Generating password credential for %s", id.ApplicationId)
			}

			newCredential, _, err := client.AddPassword(ctx, id.ApplicationId, *credential)
			if err != nil {
				return tf.ErrorDiagF(err, "Adding password for application with object ID %q", id.ApplicationId)
			}
			if newCredential == nil {
				return tf.ErrorDiagF(errors.New("nil credential received when adding password"), "API error adding password for application with object ID %q", id.ApplicationId)
			}
			if newCredential.KeyId == nil {
				return tf.ErrorDiagF(errors.New("nil or empty keyId received"), "API error adding password for application with object ID %q", id.ApplicationId)
			}
			if newCredential.SecretText == nil || len(*newCredential.SecretText) == 0 {
				return tf.ErrorDiagF(errors.New("nil or empty password received"), "API error adding password for application with object ID %q", id.ApplicationId)
			}

			// Wait for the credential to appear in the application manifest, this can take several minutes
			timeout, _ := ctx.Deadline()
			polledForCredential, err := (&pluginsdk.StateChangeConf{ //nolint:staticcheck
				Pending:                   []string{"Waiting"},
				Target:                    []string{"Done"},
				Timeout:                   time.Until(timeout),
				MinTimeout:                1 * time.Second,
				ContinuousTargetOccurence: 5,
				Refresh: func() (interface{}, string, error) {
					app, _, err := client.Get(ctx, id.ApplicationId, odata.Query{})
					if err != nil {
						return nil, "Error", err
					}

					if app.PasswordCredentials != nil {
						for _, cred := range *app.PasswordCredentials {
							if cred.KeyId != nil && strings.EqualFold(*cred.KeyId, *newCredential.KeyId) {
								return &cred, "Done", nil
							}
						}
					}

					return nil, "Waiting", nil
				},
			}).WaitForStateContext(ctx)

			if err != nil {
				return tf.ErrorDiagF(err, "Waiting for password credential for application with object ID %q", id.ApplicationId)
			} else if polledForCredential == nil {
				return tf.ErrorDiagF(errors.New("password credential not found in application manifest"), "Waiting for password credential for application with object ID %q", id.ApplicationId)
			}

			// Ensure the new value is persisted to state
			newPassword["key_id"] = pointer.From(newCredential.KeyId)
			newPassword["value"] = pointer.From(newCredential.SecretText)
			tf.Set(d, "password", []interface{}{newPassword})
		}
	}
```

Additionally, in the Update function, we should add an internal lock against the application object to prevent clobbering when the `azuread_application_password` resource is also used. This should possibly have been added here already, but it will definitely be needed now.

```go
	// add near the top of the Update function, after parsing the resource ID but before reading any values from config

	tf.LockByName(applicationResourceName, id.ApplicationId)
	defer tf.UnlockByName(applicationResourceName, id.ApplicationId)
```

We will also need to document the `password` property. Can you write up a proposal for explaining the use case for this? It doesn't need to be long, ideally something that will fit into an info box, for example:



If you can look at incorporating these changes, and there are no test breakages in the `applications` package, this should be good to merge. Thanks!

HappyTobi:
ACC tests runs fine:
```
--- PASS: TestAccApplication_password (25.09s)
--- PASS: TestAccApplication_passwordUpdate (52.37s)
```

manicminer (approved):

manicminer:
Test results

Screenshot 2024-06-25 at 14 52 14

PR #1397, opened 2024-05-31
docs: correct `identity` snippet in doc

lonegunmanb:
The `identity` in `azurerm_linux_virtual_machine` resource should be a nested block, so the `=` is incorrect.


katbyte (approved):
LGTM 🔆

PR #1406, opened 2024-06-11
`azuread_application`: work around very buggy API when instantiating from template

manicminer:
> [!IMPORTANT]
> Depends on https://github.com/manicminer/hamilton/pull/285, requires rebase prior to merging

The `/instantiate` operation can return a 404 whlst also processing the request completely normally. This leads to orphaned objects in the directory, and a resource that cannot successfully Create.

Work around this by polling for the application and service principal that you'd expect to see created out-of-band, whenever a 404 is received. Also set a temporary `displayName` for the application, as this is the only means we have to identify the resulting object is the one we are looking for.

Unfortunately this means that this workaround cannot be implemented for the `azuread_application_from_template` resource, since that resource intentionally avoids changing the implicitly created `user_impersonation` scope - this will get created with nonsensical display names / descriptions in the consent flow. Since the whole point of the standalone resource is to inherit scopes (and other fields) from the template, this makes it infeasible to add this workaround there.

Tested locally by intercepting the response to `POST /v1.0/applicationTemplates/4601ed45-8ff3-4599-8377-b6649007e876/instantiate` and sending a 404 back to the provider.

Screenshot 2024-06-11 at 18 23 59
  <br>
  <br>

**Test results:**

Screenshot 2024-06-11 at 18 19 26

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,3 +63,5 @@ require (
 )
 
 go 1.21.3
+
+replace github.com/manicminer/hamilton => github.com/MarkDordoy/hamilton v0.17.1-0.20240611151114-899c6ce169f6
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -63,3 +63,5 @@ require (
 )
 
 go 1.21.3
+
+replace github.com/manicminer/hamilton => github.com/MarkDordoy/hamilton v0.17.1-0.20240611151114-899c6ce169f6
```
tombuildsstuff, line 67: revert?

tombuildsstuff (approved):
One comment but otherwise 👍

go.mod
manicminer: Sorry forgot to note that this depends on https://github.com/manicminer/hamilton/pull/285, will rebase prior to merge

MarkDordoy (approved):

PR #1419, opened 2024-06-26
docs: timeouts and notes on restrictive API limits

manicminer:
- Ensure timeouts are documented for all resources, and all resources have correct timeouts
- Remove use of deprecated `application_id` (renamed to `client_id`) from examples
- Add info notes about restrictive API limits for `azuread_conditional_access_policy` and `azuread_named_location`

Screenshot 2024-06-26 at 19 07 10


katbyte (approved):
LGTM ⏳

Issue #1421, opened 2024-06-28
Introduction of password {} block in v2.53.0 for azuread_application forces sensitive = true on output even when not utilized


philmph:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

Running on Terraform Cloud hosted runner

```bash
Terraform v1.8.5
on linux_amd64
```

Updating via Renovate Bot `chore(deps): update terraform azuread to v2.53.0`:

| Package | Type | Update | Change |
|---|---|---|---|
| [azuread](https://registry.terraform.io/providers/hashicorp/azuread) ([source](https://github.com/hashicorp/terraform-provider-azuread)) | required_provider | minor | `2.52.0` -> `2.53.0` |

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

I am running a module which deploys apps + spns + secrets. The module is called by different environments with the respective root module. The `azuread_application` does NOT utilize the new `password` block.

`variables.tf`
```hcl
variable "service_principals" {
  description = "The Service Principals to create."
  type = map(object({
    create_password = optional(bool, false)
    display_name    = string
    description     = string

    required_resource_access = optional(list(object({
      resource_app_id = string
      resource_access = list(object({
        id   = string
        type = string
      }))
    })), [])
  }))
}
```

`main.tf`
```hcl
resource "azuread_application" "this" {
  for_each = var.service_principals

  display_name = each.value.display_name
  description  = each.value.description

  owners = local.owners

  # Required resource access
  dynamic "required_resource_access" {
    # for_each = each.value.required_resource_access
    for_each = { for i, o in each.value.required_resource_access : o.resource_app_id => o }

    content {
      resource_app_id = required_resource_access.value.resource_app_id

      dynamic "resource_access" {
        for_each = { for i, o in required_resource_access.value.resource_access : o.id => o }

        content {
          id   = resource_access.value.id
          type = resource_access.value.type
        }
      }
    }
  }
}

locals {
  azuread_applications_with_password = {
    for k, v in azuread_application.this : k => v if var.service_principals[k].create_password
  }
}

[...]

PR #1422, opened 2024-06-28
regression: `azuread_application` - don't populate the `password` block unless specified in config

manicminer:
Avoids marking all attributes as sensitive if the entire resource is used for an output variable

Fixes: #1421


tombuildsstuff (approved):
LGTM 👍

manicminer:
Before and after:

SCR-20240628-mkk

manicminer:
Test results

Screenshot 2024-06-28 at 17 34 46

PR #1532, opened 2024-10-16
`applications` - fix-up deprecation change in applications test for many owners

jackofallops:


stephybun (approved):
Thanks @jackofallops LGTM 🦘

PR #1543, opened 2024-10-22
docs: update oauth2_permission_scope documentation

robertmarsal:
Fix field marked as optional that is in fact required. When trying to create this resource without the field the following error gets returned:
```
16:47:34.579 STDERR terraform: │ Error: Could not update application with object ID: "<redacted>"
16:47:34.579 STDERR terraform: │
16:47:34.579 STDERR terraform: │   with module.<redacted>.azuread_application.this,
16:47:34.579 STDERR terraform: │   on ../../../../../../modules/<redacted>/main.tf line 16, in resource "azuread_application" "this":
16:47:34.579 STDERR terraform: │   16: resource "azuread_application" "this" {
16:47:34.579 STDERR terraform: │
16:47:34.579 STDERR terraform: │ unexpected status 400 (400 Bad Request) with error: ValueRequired: Property
16:47:34.579 STDERR terraform: │ value is required but is empty or missing.
16:47:34.580 STDERR terraform: ╵
16:47:34.660 ERROR  terraform invocation failed in .
```

Once the field is provided the error no longer occurs. The field is also required in the UI as can be seen in the below screenshot

image

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -265,7 +265,7 @@ The following arguments are supported:
 * `type` - (Optional) Whether this delegated permission should be considered safe for non-admin users to consent to on behalf of themselves, or whether an administrator should be required for consent to the permissions. Defaults to `User`. Possible values are `User` or `Admin`.
 * `user_consent_description` - (Optional) Delegated permission description that appears in the end user consent experience, intended to be read by a user consenting on their own behalf.
 * `user_consent_display_name` - (Optional) Display name for the delegated permission that appears in the end user consent experience.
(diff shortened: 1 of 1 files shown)


jackofallops (changes requested):
HI @robertmarsal - Thanks for this PR. Unfortunately the docs need to reflect the schema code, so the scheme for this property would also need to be updated.  However, afaik, `value` is a conditionally required property, so I don't believe we can mark it as required.  If you have links to docs on this, I'm happy to reconsider and help out on getting this change correct?

robertmarsal:
Thanks Jack, if it's conditionally required it makes sense. I do not have any links for docs on this to justify it should be always required.

PR #1561, opened 2024-11-05
Fix example for directory_role_eligibility_schedule_request

daftping:


jackofallops (approved):
Thanks @daftping - LGTM 👍

Issue #1556, opened 2024-10-31
[BUG] Group naming policy is not respected for 365 groups


tibbing:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

* Terraform v1.9.8
* AzureAD Provider v3.0.2

### Affected Resource(s)

* `azuread_group`

### Terraform Configuration Files

```hcl
data "azuread_client_config" "current" {}

resource "azuread_group" "example" {
  display_name               = "#example" # Naming policy requires a prefix of "#"
  owners                     = [data.azuread_client_config.current.object_id]
  security_enabled           = false
  mail_enabled               = true
  mail_nickname              = "example"
  prevent_duplicate_names    = false

  types = ["Unified"]
}
```

### Debug Output

<details><summary><i>Debug log excerpt</i></summary>

```terraform
2024/10/30 16:19:53 [DEBUG] ============================ Begin AzureAD Request ============================
Request ID: 992d4788-1f23-9872-c51d-cf6ff0861e1c

PATCH /beta/groups/0377fd2e-8abf-4865-83c6-e008c3c45f6c HTTP/1.1
Host: graph.microsoft.com
User-Agent: HashiCorp Terraform/1.9.8 (+https://www.terraform.io) Terraform Plugin SDK/2.34.0 terraform-provider-azuread/3.0.2 HashiCorp/go-azure-sdk (Go-http-Client/1.1 MicrosoftGraph-group/beta) pid-222c6c49-1b0a-5959-a213-6608f9eb8820
Content-Length: 110
Accept: application/json; charset=utf-8; IEEE754Compatible=false
Content-Type: application/json; charset=utf-8
Odata-Maxversion: 4.0
Odata-Version: 4.0
Accept-Encoding: gzip

{"@odata.type":"#microsoft.graph.group","displayName":"TERRAFORM_UPDATE_1f8282a3-8edf-8125-8c16-68bce37ca75f"}
============================= End AzureAD Request =============================: timestamp="2024-10-30T16:19:53.391+0100"
[...]

PR #1562, opened 2024-11-05
fix: don't set temporary group names to avoid breaking naming policy

tibbing:
Closes https://github.com/hashicorp/terraform-provider-azuread/issues/1556

This simply removes the logic that sets a temporary display name for Microsoft 365 groups. It will still attempt to patch the new group using the desired display name, not sure if that makes a difference. 

This seems to work fine without the temporary name after some initial testing. Are there any edge cases that this might break?


webstean:
Is there a reason this has not been merged during the last YEAR

tibbing:
> Is there a reason this has not been merged during the last YEAR

It's a bit in limbo, please see @manicminer's [comment](https://github.com/hashicorp/terraform-provider-azuread/issues/1556#issuecomment-2547063432) in the issue. 

For our purposes the only way forward was to fork the provider with this fix. It has been working fine so far, but I wish we didn't have to maintain our own fork so I still hope for an official solution.

PR #1577, opened 2024-11-29
feat(azuread_service_principals): Support for argument `filter`

tiwood:
This adds support for filtering service principals by using a OData filter.

```terraform
data "azuread_service_principals" "example" {
  filter = "startswith(displayName, 'myPrincipal')"
}
```

```terraform
data "azuread_service_principals" "example" {
  filter = "displayName eq 'foo'"
}
```

### DIFF
--- a/internal/services/serviceprincipals/service_principals_data_source.go
+++ b/internal/services/serviceprincipals/service_principals_data_source.go
@@ -40,7 +40,7 @@ func servicePrincipalsDataSource() *pluginsdk.Resource {
 				Type:         pluginsdk.TypeList,
 				Optional:     true,
 				Computed:     true,
-				ExactlyOneOf: []string{"client_ids", "display_names", "object_ids", "return_all"},
+				ExactlyOneOf: []string{"client_ids", "display_names", "object_ids", "return_all", "filter"},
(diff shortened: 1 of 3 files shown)


internal/services/serviceprincipals/service_principals_data_source.go
```diff
@@ -85,7 +85,16 @@ func servicePrincipalsDataSource() *pluginsdk.Resource {
...
 				ConflictsWith: []string{"ignore_missing"},
-				ExactlyOneOf:  []string{"client_ids", "display_names", "object_ids", "return_all"},
+				ExactlyOneOf:  []string{"client_ids", "display_names", "object_ids", "return_all", "filter"},
+			},
+
+			"filter": {
+				Description:   "Use a OData filter to search for service principals.",
+				Type:          pluginsdk.TypeString,
+				Optional:      true,
+				Default:       false,
```
jackofallops, line 95: I think this will need to be removed as this is a string value, not a bool?
```suggestion
```

jackofallops (changes requested):
Thanks @tiwood - Thanks for this. Just one comment to take a look at which I think will cause problems for cases that don't specify the new property if you can take a look?

Thanks!

Issue #1580, opened 2024-12-03
Unable to create basic groups


pgerhard:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

```shell
2024-12-03T20:53:07.856+0100 [INFO]  Terraform version: 1.9.8
2024-12-03T20:53:07.862+0100 [DEBUG] using github.com/hashicorp/go-tfe v1.58.0
2024-12-03T20:53:07.862+0100 [DEBUG] using github.com/hashicorp/hcl/v2 v2.20.0
2024-12-03T20:53:07.862+0100 [DEBUG] using github.com/hashicorp/terraform-svchost v0.1.1
2024-12-03T20:53:07.862+0100 [DEBUG] using github.com/zclconf/go-cty v1.14.4
2024-12-03T20:53:07.862+0100 [INFO]  Go runtime version: go1.22.7
2024-12-03T20:53:07.862+0100 [INFO]  CLI args: []string{"./bin/terraform", "-v"}
2024-12-03T20:53:07.862+0100 [DEBUG] Attempting to open CLI config file: /Users/pgerhard/.terraformrc
2024-12-03T20:53:07.862+0100 [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2024-12-03T20:53:07.863+0100 [DEBUG] ignoring non-existing provider search directory terraform.d/plugins
2024-12-03T20:53:07.863+0100 [DEBUG] ignoring non-existing provider search directory /Users/pgerhard/.terraform.d/plugins
2024-12-03T20:53:07.863+0100 [DEBUG] ignoring non-existing provider search directory /Users/pgerhard/Library/Application Support/io.terraform/plugins
2024-12-03T20:53:07.863+0100 [DEBUG] ignoring non-existing provider search directory /Library/Application Support/io.terraform/plugins
2024-12-03T20:53:07.864+0100 [INFO]  CLI command args: []string{"version", "-v"}
Terraform v1.9.8
on darwin_arm64
+ provider registry.terraform.io/hashicorp/azuread v3.0.2
```

### Affected Resource(s)

[...]

PR #1593, opened 2024-12-16
azuread_group: fix bug preventing creation retry when caller owner causes duplicate value error

manicminer:
Fixes: #1580


stephybun (approved):
Thanks @manicminer LGTM 👍

Issue #1600, opened 2024-12-19
brk-multihub URIs are not accepted as SPA redirect uris


barclayadam:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* `azuread_application`

### Terraform Configuration Files

I use `Pulumi` which uses v3.0.2 of the Terraform provider. The configuration I'm attempting to use in TypeScript is below:

````typescript
import * as azuread from '@pulumi/azuread';

new azure.Application('app', {
    displayName: 'Test',

    singlePageApplication: {
        redirectUris: [
            // Required for Outlook Add-In NAA
            pulumi.interpolate`brk-multihub://my-add-in-domain.com`,
        ],
    },
});
````

### Expected Behavior

`brk-multihub` should be an allowed scheme for `singlePageApplication.redirectUris`, with a domain-only that needs no trailing slash

### Actual Behavior

Without a trailing slash:

````
azuread:index/application:Application resource 'main-application' has a problem: URI must have a trailing slash when there is no path segment for "single_page_application.0.redirect_uris.0". Examine values at 'main-application.singlePageApplication.redirectUris)
````

or, with the trailing slash:

````
azuread:index/application:Application resource 'main-application' has a problem: unexpected URI scheme for "single_page_application.0.redirect_uris.0", expected one of: http, https, ms-appx-web. Examine values at 'main-application.singlePageApplication.redirectUris'.
````

### References

* [pulumi-azuread#1686](https://github.com/pulumi/pulumi-azuread/issues/1686)
* https://learn.microsoft.com/en-gb/office/dev/add-ins/develop/enable-nested-app-authentication-in-your-add-in

barclayadam:
Hi @manicminer, I see a PR exists but has been stuck since the start of Jan. Can this be progressed so I can unblock our usage? Thanks :)

MrAdam:
[...]

Issue #1401, opened 2024-06-06
azuread_application_identifier_uri parsing error for hostnames exactly 10 characters long


bwrogo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version

`terraform 1.7.4`

### Affected Resource(s)

* `hashicorp/terraform-provider-azuread`
* `azuread_application_identifier_uri `

### Terraform Configuration Files

```hcl
  resource "azuread_application_registration" "example" {
    display_name = "test_saml_application"
  }
  
  resource "azuread_application_identifier_uri" "example" {
    application_id = azuread_application_registration.example.id
    identifier_uri = "https://methenwegoyouuse.mywebsit.net/some-thing/auth/saml-endpoint.cgi?p=sp4"
  }
```

### Debug Output

```sh
  [0m[1m[31mPlanning failed.[0m[1m Terraform encountered an error while generating this plan.[0m

  [0m[31m╷[0m[0m
  [31m│[0m [0m[1m[31mError: [0m[0m[1mparsing "/applications/{redacted_saml_app_id}/identifierUris/aHR0cHM6Ly9tZXRoZW53ZWdveW91dXNlLm15d2Vic2l0Lm5ldC9zb21lLXRoaW5nL2F1dGgvc2FtbC1lbmRwb2ludC5jZ2k/cD1zcDQ=": unexpected segment "cD1zcDQ=" present at the end of the URI (input "/applications/{redacted_saml_app_id}/identifierUris/aHR0cHM6Ly9tZXRoZW53ZWdveW91dXNlLm15d2Vic2l0Lm5ldC9zb21lLXRoaW5nL2F1dGgvc2FtbC1lbmRwb2ludC5jZ2k/cD1zcDQ=")[0m
  [31m│[0m [0m
  [31m│[0m [0m[0m  with azuread_application_identifier_uri.example,
  [31m│[0m [0m  on main.tf line 18, in resource "azuread_application_identifier_uri" "example":
  [31m│[0m [0m  18: resource "azuread_application_identifier_uri" "example" [4m{[0m[0m
[...]

PR #1610, opened 2025-01-07
`applications`: URI parsing bugfixes

manicminer:
* Fix encoding of `application_identifier_uri` IDs to avoid injecting forward slashes into the base64 slug (closes: #1401)
* Loosen URI schema restrictions for application redirect URIs as the number of supported schemes has expanded and their documentation is not consolidated, making it difficult for the provider to keep up (closes: #1600)

### Changelog

* `azuread_application` - allow more URL schemes for `redirect_uris` properties
* `azuread_application_identifier_uri` - fix an encoding bug that could generate invalid resource IDs
* `azuread_application_redirect_uris` - allow more URL schemes for `redirect_uris` properties

### DIFF
--- a/internal/services/applications/migrations/application_identifier_uri_resource.go
+++ b/internal/services/applications/migrations/application_identifier_uri_resource.go
@@ -1,0 +1,89 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package migrations
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-provider-azuread/internal/helpers/tf/pluginsdk"
+	"github.com/hashicorp/terraform-provider-azuread/internal/services/applications/parse"
+)
+
+type ResourceApplicationIdentifierUriStateUpgradeV0 struct{}
+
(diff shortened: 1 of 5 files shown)


internal/services/applications/migrations/application_identifier_uri_resource.go
```diff
@@ -0,0 +1,89 @@
...
+			ForceNew:     true,
+			ExactlyOneOf: []string{"end_date_relative"},
+		},
+
+		"end_date_relative": {
+			Type:         pluginsdk.TypeString,
+			Optional:     true,
+			ForceNew:     true,
+			ExactlyOneOf: []string{"end_date"},
+		},
```
stephybun, line 67: Doesn't this have to match the schema of the resource exactly? When I look at the resource schema I only see two properties https://github.com/hashicorp/terraform-provider-azuread/blob/8cead2417164648326d3ba16ab8f1cce8a6b4080/internal/services/applications/application_identifier_uri_resource.go#L55-L71
Unless I'm missing something? 🙈

jackofallops: @manicminer - This migration schema looks like it belongs to `azuread_application_certificate`?

Issue #1412, opened 2024-06-21
Error 400 (RoleAssignmentExists) when modifying existing azuread_privileged_access_group_eligibility_schedule resources


michvllni:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
* Terraform Version 1.8.5
* AzureAD Provider Version 2.52.0

### Affected Resource(s)

* azuread_privileged_access_group_eligibility_schedule

### Terraform Configuration Files

```hcl
resource "azuread_privileged_access_group_eligibility_schedule" "gdap_rolegroup_default_assignments" {
  for_each = { for assignment in flatten([
    for group_key, group in azuread_group.gdap_groups : [
      for user_key, user in data.azuread_user.gdap_rolegroup_default_members : {
        group_key    = group_key
        user_key     = replace(user_key, "/@.*/", "") # remove dots and domain
        group_id     = group.object_id
        principal_id = user.object_id
      }
    ]
    ]) : "${assignment.group_key}-${assignment.user_key}" => assignment
  }
  group_id             = each.value.group_id
  principal_id         = each.value.principal_id
  assignment_type      = "member"
  permanent_assignment = true
}
```

### Expected Behavior

Resource should be updatable

### Actual Behavior

Modifying fails on apply with 

```
│ Error: Could not create assignment schedule request, PrivilegedAccessGroupEligibilityScheduleRequestsClient.BaseClient.Post(): unexpected status 400 with OData error: RoleAssignmentExists: The Role assignment already exists.
│ 
│   with module.io-gdap-groups["xxx"].azuread_privileged_access_group_eligibility_schedule.gdap_rolegroup_default_assignments["bcadmin-name"],
│   on modules/io-gdap-groups/adgroups.tf line 52, in resource "azuread_privileged_access_group_eligibility_schedule" "gdap_rolegroup_default_assignments":
│   52: resource "azuread_privileged_access_group_eligibility_schedule" "gdap_rolegroup_default_assignments" {
│ 
│ Could not create assignment schedule request,
[...]

PR #1614, opened 2025-01-09
Fix azuread_privileged_access_group_eligibility_schedule resource update functionality 

h0597:
My goal with this pull request is to improve the azuread_privileged_access_group_eligibility_schedule resource. 

Updating the 'Update' functionality to use the `adminUpdate` instead of `adminAssign` [As per Microsoft graph documentation ](https://learn.microsoft.com/en-us/graph/api/privilegedaccessgroup-post-eligibilityschedulerequests?view=graph-rest-1.0&tabs=http#request-body)

> `adminUpdate`: For administrators to change existing eligible assignments. instead of `adminAssign`: For administrators to assign group membership or ownership eligibility to principals.

I found this issue in relation to this problem: https://github.com/hashicorp/terraform-provider-azuread/issues/1412

As this is my first fork pull request, I hope this way is correct to do it. If not, please let me know so I can improve the pull request.

### DIFF
--- a/internal/services/identitygovernance/privileged_access_group_eligiblity_schedule_resource.go
+++ b/internal/services/identitygovernance/privileged_access_group_eligiblity_schedule_resource.go
@@ -228,7 +228,7 @@ func (r PrivilegedAccessGroupEligibilityScheduleResource) Update() sdk.ResourceF
 				AccessId:      stable.PrivilegedAccessGroupRelationships(resourceId.Relationship),
 				PrincipalId:   nullable.Value(model.PrincipalId),
 				GroupId:       nullable.Value(resourceId.GroupId),
-				Action:        pointer.To(stable.ScheduleRequestActions_AdminAssign),
+				Action:        pointer.To(stable.ScheduleRequestActions_AdminUpdate),
 				Justification: nullable.NoZero(model.Justification),
 				ScheduleInfo:  schedule,
 			}
--- a/internal/services/identitygovernance/privileged_access_group_eligiblity_schedule_resource_test.go
+++ b/internal/services/identitygovernance/privileged_access_group_eligiblity_schedule_resource_test.go
@@ -34,6 +34,9 @@ func TestPrivilegedAccessGroupEligibilitySchedule_member(t *testing.T) {
 			Config: r.member(data, endTime),
 			Check: acceptance.ComposeTestCheckFunc(
 				check.That(data.ResourceName).ExistsInAzure(r),
+				// There is a minimum life of 5 minutes for a schedule request to exist.
+				// Attempting to delete the request within this time frame will result in
+				// a 400 error on destroy, which we can't trap.
 				helpers.SleepCheck(5*time.Minute+15*time.Second),
 			),
 		},
--- a/internal/services/identitygovernance/privileged_access_group_eligiblity_schedule_resource_test.go
+++ b/internal/services/identitygovernance/privileged_access_group_eligiblity_schedule_resource_test.go
@@ -34,9 +34,6 @@ func TestPrivilegedAccessGroupEligibilitySchedule_member(t *testing.T) {
 			Config: r.member(data, endTime),
 			Check: acceptance.ComposeTestCheckFunc(
 				check.That(data.ResourceName).ExistsInAzure(r),
-				// There is a minimum life of 5 minutes for a schedule request to exist.
-				// Attempting to delete the request within this time frame will result in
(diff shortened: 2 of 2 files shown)


SPALLADI:
Any luck with merging the PR?

relent0r:
I'm interested in this one too as I've hit the issue where you get 400 errors and are unable to apply after any changes are required. I need to manually delete all the assignments and reapply to make it work which given how slow the azure interface is for this is really painful.

farflungfish:
This is definitely the solution we are looking for; total pain in the rear without it at the moment having to recreate role assignments.

xzuttz:
An issue for us as well. Looking forward to a solution soon.

kevinfalconer-seccl:
Issue that we are facing at the moment as well, hopefully will be looked at soon!

xgryph:
bump!

aaron-warner-nzta:
I've been doing testing on this fix today. It looks like it causes a problem where the state is not being updated after the change is applied.

I think its because when the update takes place a new assignment is created on the MS side but the state is not updated with this new assignment information and still has the original ID reference. So every time it reads from the API it uses the previous assignments ID which contains the original configuration so Terraform see's no changes being required.

csg1001:
Is there an update on this?

jackofallops (changes requested):
Thanks @h0597 - Can you add an update test to show that this change correctly succeeds on the update call?  My reasoning is that we may also need to change the API call to use the update method. In either case, we should have test coverage that shows the change is correct.  If you can take a look, I can re-review.

Thanks!

feliasson:
@jackofallops The PR author has added tests and requested another review. It would be greatly appreciated if this could be merged to resolve the issue.

codethief:
Looks like @h0597 abandoned the PR. (They haven't been active on GitHub since creating the PR.) What do we do here? I'd be happy to submit an updated PR, however I don't have an AzureAD tenant to run the acceptance tests with, so I'd need some help.

@jackofallops Just to make sure I understand your previous comment correctly: The broken test showed up *after* rebasing the current PR on the changes from https://github.com/hashicorp/terraform-provider-azuread/pull/1790 (which are now on main)? **UPDATE:** See my next comment below.

On another note: Claude reports that a very similar bug exists in `internal/services/identitygovernance/privileged_access_group_assignment_schedule_resource.go`: Here, too, `Update()` uses `adminAssign` instead of the actually correct `adminUpdate`, compare https://learn.microsoft.com/en-us/graph/api/privilegedaccessgroup-post-eligibilityschedulerequests?view=graph-rest-1.0&tabs=http#request-body . Should the PR address this in one fell swoop?

EdisIbrovic15:
i have the same issue for this resource as descripted and solved here: https://github.com/hashicorp/terraform-provider-azuread/pull/1682   after some time the resource disapears from my statefile but still present in Azure so terraform tries to create the resource but fails since it already exists.

codethief:
So I've finally got a test tenant with an Entra ID P2 license through https://developer.microsoft.com/en-us/microsoft-365/dev-program (had to find someone with a Visual Studio Pro subscription…) and have been able to test the changes made in this PR.

> @jackofallops Just to make sure I understand your previous comment correctly: The broken test showed up *after* rebasing the current PR on the changes from https://github.com/hashicorp/terraform-provider-azuread/pull/1790 (which are now on main)?

The answer to my question is _yes_ – kind of: The error showed up after applying the PR changes to `main`. However, the issue has nothing to do with the PR and has been present for a long time – the bug being fixed by the PR was merely hiding it: Updating an eligibility schedule (or assignment schedule) will change the resource ID, so the state needs to be updated. I'm working on a new PR which fixes this as well as an additional issue that showed up.

codethief:
@EdisIbrovic15 

> i have the same issue for this resource as descripted and solved here: https://github.com/hashicorp/terraform-provider-azuread/pull/1682 after some time the resource disapears from my statefile but still present in Azure so terraform tries to create the resource but fails since it already exists.

How is this related to https://github.com/hashicorp/terraform-provider-azuread/issues/1412 which the present PR is trying to fix? Also, if I understand correctly, your issue has already been solved by the PR you linked?

codethief:
As mentioned before, it looks like the present PR has been abandoned and additional changes are needed to make the tests green. I therefore created a new PR: https://github.com/hashicorp/terraform-provider-azuread/pull/1856 .

sreallymatt:
superseded by #1856

PR #1621, opened 2025-01-15
[Docs] azuread_group_member - update example usage code block

joshvito:
Updates example usage to be match the attributes of the azuread_group and azuread_user resources.


stephybun (approved):
Thanks @joshvito LGTM 👍

PR #1623, opened 2025-01-16
New Resource: `azuread_group_without_members`

### LINKED ISSUE #954: Group Resource CRUD slowdown on large member diff
### Community Note

terraform version: v0.14.11
hashicorp/azuread: latest (2.31.0)

### Affected Resource(s)

* `azuread_group`

```hcl
// group has ~1400 members
data "azuread_group" "source" {
  display_name = "Group_With_Tons_Of_Members"
}

resource "azuread_group" "testing" {
  display_name = "testing-time-complexity"
  security_enabled = true
  members = data.azuread_group.source.members
}

```

### Debug Output

None

### Panic Output

None

### Context

I've been testing the time complexity of group creation/update/delete with very large member diffs. I spotted the [following linear relationship between members and GET requests in the provider group_resource.go](https://github.com/hashicorp/terraform-provider-azuread/blob/main/internal/services/groups/group_resource.go#L774)

```golang
// Add members after the group is created
members := make(msgraph.Members, 0)
if v, ok := d.GetOk("members"); ok {
	for _, memberId := range v.(*schema.Set).List() {
		memberObject, _, err := directoryObjectsClient.Get(ctx, memberId.(string), odata.Query{})
		if err != nil {
			return tf.ErrorDiagF(err, "Could not retrieve member principal object %q", memberId)
		}
		if memberObject == nil {
			return tf.ErrorDiagF(errors.New("memberObject was nil"), "Could not retrieve member principal object %q", memberId)
		}
		// TODO: remove this workaround for https://github.com/hashicorp/terraform-provider-azuread/issues/588
		//if memberObject.ODataId == nil {
		//	return tf.ErrorDiagF(errors.New("ODataId was nil"), "Could not retrieve member principal object %q", memberId)
		//}
		memberObject.ODataId = (*odata.Id)(utils.String(fmt.Sprintf("%s/v1.0/%s/directoryObjects/%s",
			client.BaseClient.Endpoint, client.BaseClient.TenantId, memberId)))

		members = append(members, *memberObject)
	}
}
```
I was wondering if there was a reason we wern't using the [`directoryObjectsClient.GetByIds` wrapper](https://github.com/manicminer/hamilton/blob/main/msgraph/directory_objects.go#L58) in place of the iteration over the memberIDs and `directoryObjectsClient.Get`. We could possibly speed-up the apply for this resource by cutting-down on the API calls.

jackofallops:
This PR introduces a Group resource that omits management and exposure of group members. This resource is intended for use cases where a Group needs to be managed via terraform, but the members of the group are not of concern, typically where the members list is sufficiently large that performance problems are encountered, and/or members are managed by other means.

closes #954 - Since there is no current way to improve the linear performance problem with the List() call, the `azuread_group` resource cannot be changed to satisfy a performance improvement.

### DIFF
--- a/.github/labeler-issue-triage.yaml
+++ b/.github/labeler-issue-triage.yaml
@@ -25,7 +25,7 @@ feature/domains:
   - '### (|New or )Affected Resource\(s\)\/Data Source\(s\)((.|\n)*)azuread_domains((.|\n)*)###'
 
 feature/groups:
-  - '### (|New or )Affected Resource\(s\)\/Data Source\(s\)((.|\n)*)azuread_(group\W+|group_member\W+|groups\W+)((.|\n)*)###'
+  - '### (|New or )Affected Resource\(s\)\/Data Source\(s\)((.|\n)*)azuread_(group\W+|group_member\W+|group_without_members\W+|groups\W+)((.|\n)*)###'
 
 feature/identity-governance:
   - '### (|New or )Affected Resource\(s\)\/Data Source\(s\)((.|\n)*)azuread_(access_package|privileged_access_group_)((.|\n)*)###'
--- a/internal/services/groups/group_without_members_resource.go
+++ b/internal/services/groups/group_without_members_resource.go
@@ -182,9 +182,9 @@ func groupWithoutMembersResource() *pluginsdk.Resource {
 				Type:        pluginsdk.TypeSet,
 				Optional:    true,
 				Computed:    true,
-				MinItems:    1,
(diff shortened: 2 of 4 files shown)


Andrei-Predoiu:
This is something we really need. We don't set members inside the azuread_group resource but the provider still tries to read all the members.

jackofallops:
as per `azuread_group` some tests are failing due to a required change on the underlying CI server image.

image

katbyte (approved):
LGTM 🍰

PR #1655, opened 2025-03-02
fix import example to remove incorrect /groups/ prefix

drdamour:


jackofallops:
Thanks @drdamour - The SDK uses the `/groups/xxx` format as defined by the underlying Go type (see [here](https://github.com/hashicorp/terraform-provider-azuread/blob/c3a4ff171ecf31bca32e7aadb10641baff6405b5/vendor/github.com/hashicorp/go-azure-sdk/microsoft-graph/groups/beta/group/method_getgroup.go#L63)) to perform reads, so the prefix is required to correctly format the ID for import. Are you experiencing a problem with this?

drdamour:
yeah it doesnt work, you have to omit the groups prefix

sreallymatt:
Hi @drdamour, I've just tested and confirmed that importing as documented works (when using the latest version of the provider).

Prior to v3.0.0, imports for this resource were done with just the UUID, and if you are using an older version of the provider this is likely causing your issue. (Changing the version selector in the Terraform registry documentation to 2.53.1 or older shows the import example with just the UUID)

```
terraform import azuread_group.test /groups/cbd383bc-160a-4020-99d1-7de16c9f61b2
azuread_group.test: Importing from ID "/groups/cbd383bc-160a-4020-99d1-7de16c9f61b2"...
azuread_group.test: Import prepared!
  Prepared azuread_group for import
azuread_group.test: Refreshing state... [id=/groups/cbd383bc-160a-4020-99d1-7de16c9f61b2]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

Based on this, I'm going to close this PR. If you are still encountering the import issue on the latest version of the provider, please open an issue that includes a reproduction and the Terraform + AzureAD provider versions used (`terraform version` output)

Issue #1584, opened 2024-12-09
Removal of "azuread_application_pre_authorized" also removes pre-existing pre-authorized clients not managed by Terraform


l33tCod-er:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritise this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritise the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and AzureAD Provider) Version
* provider version 2.53.1

### Affected Resource(s)

* `azuread_application_pre_authorized`

### Terraform Configuration Files

```hcl

# This null_resource adds clients for which there is no app registration. Thus, we can't use "azuread_application_pre_authorized" for them, anyway, the need to be allowed to access call a certain scope on the "target_app"
resource "null_resource" "patched_app_reg" {
  triggers = {
    scope_1 = lookup(<target_app>, "scope1", <tenant>)
    scope_2 = lookup(<target_app>, "scope2", <tenant>)
  }

  provisioner "local-exec" {
    interpreter = ["pwsh", "-Command"]
    command     = <<EOF
        az rest --method Patch --uri "https://graph.microsoft.com/v1.0/applications/<target_app>" --headers 'Content-Type=application/json' --body '{"api": {"preAuthorizedApplications": [{"appId": <some_guid_1>, "delegatedPermissionIds": [<all_scope_id_lookup>, "scope1", <tenant>)}"]},{"appId": <some_guid_2>, "delegatedPermissionIds": [<all_scope_id_lookup>, "scope1", <tenant>)}"]},{"appId": <some_guid_3>, "delegatedPermissionIds": [<all_scope_id_lookup>, "scope1", <tenant>)}"]}]}}'
EOF

    environment = {
      POWERSHELL_TELEMETRY_OPTOUT = 1
    }
  }
}

resource "azuread_application_pre_authorized" "new_app_reg" {
  application_object_id = <target_app>
  authorized_client_id  = <some_guid_with_app_reg>
  permission_ids        = [<scope_lookup>, "scope1", <tenant>)]
}
```

### Expected Behavior

[...]

PR #1659, opened 2025-03-06
`azuread_application_pre_authorized` - Fix unexpected removal of unmanaged pre-authorized apps

l33tCod-er:
Range breaks at first pre_authed app with a different id during deletion, but we must preserve all existing pre_authorized apps with a different id.

Shall fix [#1584](https://github.com/hashicorp/terraform-provider-azuread/issues/1584)


jackofallops (approved):
Thanks @l33tCod-er - LGTM 👍

Issue #1653, opened 2025-02-27
Conditional Access Authentication flows Device Code Flow


mtx10:
We are currently using the azuread_conditional_access_policy resource to deploy our Azure Conditional Access Rules. However, we have encountered a limitation where we are unable to create a rule that disallows authentication flows using the Device Code Flow.

As of now, there is no documentation or implementation available to support this functionality. We believe that having the ability to block the Device Code Flow would enhance our security posture by allowing us to enforce stricter access controls.

We kindly request the addition of this feature to the azuread_conditional_access_policy resource. It would be beneficial to have detailed documentation and examples on how to implement this once available.

Thank you for considering this request. We look forward to seeing this feature in future updates.

![Image](https://github.com/user-attachments/assets/a139d3d2-26f6-40d3-a2ec-1105a713b410)

Keith-EMP:
I would also like to see this option added

biggles007:
I was about to add an issue on this, but you beat me to it. This is a blocker to putting all policies as Terraform.

richardlock:
I've started a PR for this. Needs some more work but hopefully some of the suggested code is useful.

https://github.com/hashicorp/terraform-provider-azuread/pull/1662

jackofallops:
@richardlock - This feature is only available the `beta` API afaik, so we need to investigate the migration (or hybridisation with `stable`) to provide this, which will need a fair amount of regression testing to cover. It's not on our roadmap at the moment, but if you want to look into it, that's the place to start.

richardlock:
Hi Jack. Thanks for replying. It looks like the feature is in the normal 1.0 API as well as the beta API:

https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessauthenticationflows?view=graph-rest-1.0

mysteq:
Is this currently missing due to the microsoft-graph definitions in hashicorp/go-azure-sdk not being updated for a while, due to this issue https://github.com/hashicorp/pandora/pull/4707 in hashicorp/pandora?

enorlando:
Is there any update on this?

PR #1662, opened 2025-03-11
#1653 Start work to add Conditional Access authentication flows feature

richardlock:
This needs more work and testing, but started the work to add this feature.


jackofallops:
Thanks @richardlock - unfortunately this will require a later version of the SDK which is currently blocked by a change in the upstream MSFT source data for generating that new version. We're working diligently with Microsoft to get this unblocked, so we'll loop back asap!

richardlock:
Thanks @jackofallops.

PR #1745, opened 2025-08-01
use `go-version` for `link-milestone.yaml`

sreallymatt:
Link Milestone action always fails due to it not finding a `.go-version` file. Update to match `link-milestone.yaml` action on azurerm.


catriona-m (approved):
Thanks @sreallymatt LGTM!

PR #1906, opened 2026-08-04
`testing` - fix-up template config for change in mandatory default policy

jackofallops:


catriona-m (approved):
Thanks @jackofallops LGTM!

PR #1911, opened 2026-08-06
groups: select only `id` when listing members and owners

### LINKED ISSUE #1910: data.azuread_group: members read fails with "invalid character '{' after object key:value pair" (Graph beta returns a truncated body for app-only tokens without User.Read.All)
### Community Note

```
Terraform v1.15.8
on linux_amd64
+ provider registry.terraform.io/hashicorp/azuread v3.9.0
```

### Affected Resource(s)

* `azuread_group` (data source)

```hcl
provider "azuread" {}

# Authenticated app-only (ARM_CLIENT_ID / ARM_CLIENT_SECRET / ARM_TENANT_ID) with a service
# principal holding Group.Read.All, Group.Create, Application.Read.All,
# Application.ReadWrite.OwnedBy — and *not* User.Read.All.
data "azuread_group" "deployers" {
  object_id = var.deployers_group
}

output "members" {
  value = data.azuread_group.deployers.members
}
```

### Debug Output

The relevant part, from `TF_LOG_PROVIDER=DEBUG` (IDs redacted). The provider requests:

```
GET https://graph.microsoft.com/beta/groups/<group-id>/members
```

and Graph replies `200` with a body that **stops mid-object** and has an error object appended
to the bytes already sent:

```json
{"@odata.context":"https://graph.microsoft.com/beta/$metadata#directoryObjects","value":[{"@odata.type":"#microsoft.graph.user","id":"<user-id>","deletedDateTime":null,"accountEnabled":null,...,"userPrincipalName":null,"userType":null,"identityParentId":null,"agentIdentityBlueprintId":null{"error":{"code":"InternalServerError","message":"The property 'isProvisionedToOnPremises[Nullable=False]' of type 'microsoft.graph.onPremisesProvisioningState' has a null value, which is not allowed.","innerError":{"date":"2026-08-06T08:31:46","request-id":"<request-id>","client-request-id":"<request-id>"}}}
```

Note `"agentIdentityBlueprintId":null{"error":` — two JSON fragments spliced together, which is
what produces the `invalid character '{' after object key:value pair` unmarshalling error.

### Panic Output

n/a

### Expected Behavior

The data source reads the group's members, or fails with an error naming the actual cause
(Graph returned `InternalServerError` while serializing the member list).

### Actual Behavior

```
Error: Could not retrieve group members for group with object ID: "/groups/<group-id>"

  with data.azuread_group.deployers,
  on main.tf line 1, in data "azuread_group" "deployers":
   1: data "azuread_group" "deployers" {

invalid character '{' after object key:value pair
```

The message points at the group and at a JSON syntax problem, which sends you looking for a bad
group or a broken HTTP path. The real cause is a server-side serialization failure, invisible
unless you enable provider debug logging and read the raw body.

### Steps to Reproduce

1. Authenticate app-only as a service principal that has `Group.Read.All` but **not**
   `User.Read.All` (so it may list group members but not read user objects — reading
   `/beta/directoryObjects/{id}` for a member returns `403`).
2. `terraform plan` against a `data "azuread_group"` for a security group with user members.

Reduced to plain HTTP, same group, same tenant, same Graph slice — only the request or the token
varies:

| request | token | result |
| --- | --- | --- |
| `/v1.0/groups/{id}/members` | app-only, no `User.Read.All` | 200, well-formed |
| `/beta/groups/{id}/members` | app-only, no `User.Read.All` | 200, **truncated at 1994 bytes + `InternalServerError`** |
| `/beta/groups/{id}/members?$select=id,displayName,userPrincipalName` | app-only, no `User.Read.All` | 200, well-formed |
| `/beta/groups/{id}/members` | delegated (user) | 200, well-formed (79 properties per member) |

The pattern is consistent with the cause: without `User.Read.All`, the app-only projection comes
back with every user property `null` (`displayName`, `userPrincipalName`, `accountEnabled` …),
and beta's own non-nullable `isProvisionedToOnPremises` then holds `null`, which its serializer
rejects part-way through the response. Restricting the projection with `$select`, or asking
`v1.0`, never reaches that property. Failure is deterministic, not slice luck (8/8 attempts).

### Important Factoids

* Nothing atypical: Azure global (not China/Germany/Government), a plain security group
  (`groupTypes: []`) with ordinary user members.
* Granting the service principal `User.Read.All` is a viable workaround, because the member
  objects then come back populated. But it means a data source that only needs member IDs
  requires directory-wide user read permission.
* `data "azuread_group"` always reads members and owners, so there is no way to opt out from
  configuration when only `object_id` is needed.

### Suggestions

The calls in question are `ListMembers` / `ListOwners` in
[`internal/services/groups/group_data_source.go`](https://github.com/hashicorp/terraform-provider-azuread/blob/main/internal/services/groups/group_data_source.go#L473-L489),
both against `beta` with `DefaultList{Members,Owners}OperationOptions()` — so no `$select`.

Any one of these would make the data source robust here, independent of the Graph bug:

1. Send `$select` on the members/owners reads for the fields the data source actually exposes.
   Per the table above this alone avoids the failure, and it makes the responses dramatically
   smaller (819 KB → a few KB for 39 members).
2. Read members from `v1.0` rather than `beta`. I appreciate this was tried in 2.34.0 and
   reverted in 2.34.1, so (1) seems the safer option.
3. Failing both, detect a truncated body / embedded `"error"` object and surface Graph's
   `InternalServerError` message instead of the JSON syntax error.

The serialization failure itself is a Graph beta bug and is Microsoft's to fix; this issue is
about the provider being resilient to it, and about the error message pointing somewhere useful.

### References

* n/a

orgads:
## Community Note
* Please vote on this PR by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original PR to help the community and maintainers prioritize for review
* Please do not leave comments along the lines of "+1", "me too" or "any updates", they generate extra noise for PR followers and do not help prioritize for review

## Description

The group data source and the group / group_member / group_without_members resources read members
and owners purely for their object IDs — every consumer is
`pointer.From(object.DirectoryObject().Id)` — but they ask Graph for the full directory object
projection. This requests `$select=id` instead, through three small helpers next to the existing
`groupGetMember` in `internal/services/groups/groups.go`.

**The payload difference is substantial.** For a group of 39 users:

| request | response size |
| --- | --- |
| `/beta/groups/{id}/members` | 819 KB (79 properties per member) |
| `/beta/groups/{id}/members?$select=id` | 694 bytes |

That is ~1000x less to transfer and unmarshal, for a data source that then discards all of it
except one field per member. It applies on every read of a group's `members` and `owners`,
including refreshes of `azuread_group` resources.

**How I got here** (#1910): the full projection was also what tripped a Microsoft Graph beta
serialization bug for app-only callers lacking `User.Read.All`. Every user property in that
restricted projection comes back `null`, including the non-nullable `isProvisionedToOnPremises` of
`microsoft.graph.onPremisesProvisioningState`, so Graph aborted mid-object and appended an error to
the bytes it had already sent, which the provider surfaced as:

```
Error: Could not retrieve group members for group with object ID: "/groups/<id>"
invalid character '{' after object key:value pair
```

Microsoft have since fixed that serialization failure, so **this PR is no longer needed to fix
#1910** — I have confirmed the failure no longer reproduces and said so on the issue. I still think
the change is worth taking on the two remaining grounds: the payload saving above, and not asking
Graph for properties the provider never reads, which is what exposed it to a server-side bug in a
projection it had no use for.

`@odata.type` is still returned under `$select=id`, so the polymorphic directory-object
unmarshalling is unaffected — responses come back as `{"@odata.type": "...", "id": "..."}`.

## Changes to existing Resource / Data Source

- [x] I have added an explanation of what my changes do and why I'd like you to include them (This may be covered by linking to an issue above, but may benefit from additional explanation).
- [ ] I have written new tests for my resource or datasource changes & updated any relevant documentation. — no new tests: this changes the request sent to Graph, not observable schema or behaviour, and the existing acceptance tests for these resources and the data source already assert the `members` / `owners` values this code path produces. Happy to add coverage if you can suggest a shape for asserting `$select` that fits the suite.
- [x] I have successfully run tests with my changes locally. See below.
- [ ] (For changes that include a **state migration only**). I have manually tested the migration path between relevant versions of the provider.

## Testing

Unit tests, build and vet pass:

```
$ go build ./... && go vet ./internal/services/groups/... && go test ./internal/...
ok  	github.com/hashicorp/terraform-provider-azuread/internal/services/groups	0.018s
```

I do not have a tenant I can point the acceptance tests at. Instead I verified against a real
tenant with a `dev_overrides` build, reading two security groups (40 and 7 members) with an
app-only service principal, both with `include_transitive_members = false` and via the `owners`
path:

```
data.azuread_group.deployers: Read complete after 2s [id=/groups/7e9a2227-...]
data.azuread_group.admin: Read complete after 2s [id=/groups/11478e71-...]

Changes to Outputs:
  + admin_members     = 40
  + admin_owners      = 7
  + deployers_members = 7
  + deployers_sample  = "2612466e-..."
```

The member and owner IDs are identical to what the unfiltered request returns for the same groups,
i.e. narrowing the projection loses nothing the data source exposes. Note this is no longer a
before/after comparison: when I first ran it, these two reads failed on released 3.9.0 with the
error quoted above and succeeded with this change — but with Microsoft's fix in place, stock 3.9.0
now reads them successfully too.

## Change Log

* `azuread_group` - request only the `id` property when reading members and owners, greatly reducing the response size [GH-1910]
* `azuread_group_member` - request only the `id` property when listing existing members [GH-1910]
* `azuread_group_without_members` - request only the `id` property when reading owners [GH-1910]

This is a (please select all that apply):

- [ ] Bug Fix
- [ ] New Feature (ie adding a service, resource, or data source)
- [x] Enhancement
- [ ] Breaking Change

## Related Issue(s)
Relates to #1910

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the provider.

## Changes to Security Controls

No changes to security controls. The requests are narrowed, not widened: the provider now asks
Graph for strictly fewer properties than before, and no permission requirements change.

### DIFF
--- a/internal/services/groups/group_data_source.go
+++ b/internal/services/groups/group_data_source.go
@@ -15,9 +15,6 @@ import (
 	"github.com/hashicorp/go-azure-sdk/microsoft-graph/common-types/beta"
 	"github.com/hashicorp/go-azure-sdk/microsoft-graph/common-types/stable"
 	groupBeta "github.com/hashicorp/go-azure-sdk/microsoft-graph/groups/beta/group"
-	memberBeta "github.com/hashicorp/go-azure-sdk/microsoft-graph/groups/beta/member"
-	ownerBeta "github.com/hashicorp/go-azure-sdk/microsoft-graph/groups/beta/owner"
-	transitivememberBeta "github.com/hashicorp/go-azure-sdk/microsoft-graph/groups/beta/transitivemember"
 	"github.com/hashicorp/terraform-provider-azuread/internal/clients"
 	"github.com/hashicorp/terraform-provider-azuread/internal/helpers/tf"
 	"github.com/hashicorp/terraform-provider-azuread/internal/helpers/tf/pluginsdk"
@@ -458,7 +455,7 @@ func groupDataSourceRead(ctx context.Context, d *pluginsdk.ResourceData, meta in
 	includeTransitiveMembers := d.Get("include_transitive_members").(bool)
 	var members *[]string
 	if includeTransitiveMembers {
-		resp, err := transitiveMemberClient.ListTransitiveMembers(ctx, beta.GroupId(id), transitivememberBeta.DefaultListTransitiveMembersOperationOptions())
+		resp, err := transitiveMemberClient.ListTransitiveMembers(ctx, beta.GroupId(id), listTransitiveMembersOptions())
 		if err != nil {
 			return tf.ErrorDiagF(err, "Could not retrieve transitive group members for group with object ID: %q", d.Id())
 		}
@@ -470,7 +467,7 @@ func groupDataSourceRead(ctx context.Context, d *pluginsdk.ResourceData, meta in
 			members = &transitiveMembers
 		}
 	} else {
-		resp, err := memberClient.ListMembers(ctx, beta.GroupId(id), memberBeta.DefaultListMembersOperationOptions())
+		resp, err := memberClient.ListMembers(ctx, beta.GroupId(id), listMembersOptions())
 		if err != nil {
 			return tf.ErrorDiagF(err, "Could not retrieve group members for group with object ID: %q", d.Id())
 		}
@@ -484,7 +481,7 @@ func groupDataSourceRead(ctx context.Context, d *pluginsdk.ResourceData, meta in
 	}
 	tf.Set(d, "members", members)
 
-	resp, err := ownerClient.ListOwners(ctx, beta.GroupId(id), ownerBeta.DefaultListOwnersOperationOptions())
+	resp, err := ownerClient.ListOwners(ctx, beta.GroupId(id), listOwnersOptions())
 	if err != nil {
 		return tf.ErrorDiagF(err, "Could not retrieve group owners for group with object ID: %q", d.Id())
 	}
--- a/internal/services/groups/group_member_resource.go
+++ b/internal/services/groups/group_member_resource.go
@@ -77,7 +77,7 @@ func groupMemberResourceCreate(ctx context.Context, d *pluginsdk.ResourceData, m
 		return tf.ErrorDiagPathF(err, "object_id", "Retrieving %s", groupId)
 	}
 
-	resp, err := memberClient.ListMembers(ctx, groupId, memberBeta.DefaultListMembersOperationOptions())
+	resp, err := memberClient.ListMembers(ctx, groupId, listMembersOptions())
 	if err != nil {
 		return tf.ErrorDiagF(err, "Listing existing members for %s", groupId)
 	}
--- a/internal/services/groups/group_resource.go
+++ b/internal/services/groups/group_resource.go
@@ -1089,7 +1089,7 @@ func groupResourceUpdate(ctx context.Context, d *pluginsdk.ResourceData, meta in
 	}
 
 	if d.HasChange("members") {
-		resp, err := memberClient.ListMembers(ctx, *id, memberBeta.DefaultListMembersOperationOptions())
+		resp, err := memberClient.ListMembers(ctx, *id, listMembersOptions())
 		if err != nil {
 			return tf.ErrorDiagF(err, "Could not retrieve members for %s", id)
 		}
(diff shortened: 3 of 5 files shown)


orgads:
ping?

PR #1922, opened 2026-08-12
docs: fix import command format for service principals

Fuco1:
## Description

Updated the import command for service principals to remove the leading '/servicePrincipals/'.  Using the full ID 
returns `Error: specified ID ("/servicePrincipals/00000....") is not valid: uuid string is wrong length`


toddgiguere (approved):
@Fuco1 Thank you for the correction! I'll merge this in and it will reflect in the next version.

PR #1939, opened 2026-09-10
lint: golangci with tfproviderlint plugin and azurerm's linter set; typos; reviewer assignment; pin all tooling and actions

katbyte:
Fourth PR in the CI alignment series, stacked on #1938 (base is `kt/ci-teamcity-workflows`); rebase onto `main` after that merges. Brings the golangci-lint setup in line with terraform-provider-azurerm, minus `azproviderlint` which is deferred to a follow-up.

### Pinned tooling in `.tools` (as in hashicorp/terraform-provider-azurerm#33429)

Every dev tool is a pinned install under `.tools`, so contributors and CI need only Go and make (plus `python3`/`npm` for yamllint and markdownlint-cli2).

- Go tools (golangci-lint, actionlint, gofumpt, misspell, tfproviderdocs, terrafmt, tctest) pinned via `tool` directives in `.tools/go.mod` (own module, nothing lands in `vendor/`); dependabot bumps them quarterly.
- non-Go tools (shellcheck, typos, yamllint, markdownlint-cli2) pinned in the `GNUmakefile` and installed into `.tools`.
- make targets list the binaries they need as prerequisites and build/install them on demand; `.tools/bin` is prepended to `PATH` for the check scripts. `make tools` installs everything up front.
- `depscheck` also tidies `.tools/go.mod` and checks the golangci-lint version there matches `.tools/.custom-gcl.yml`.
- workflows drop their ad-hoc `go install` / curl / npm steps and build from `.tools`; the golint jobs cache `.tools/bin` keyed on the pins.
- tfproviderdocs is built in isolation (`go install pkg@version`) rather than from the shared module: the goldmark bump golangci-lint brings in makes v0.12.1's contents check report spurious "arguments section is not sorted" errors. The pin still lives in `.tools/go.mod`.

### Plugin-built binary

- `.tools/.custom-gcl.yml` pins the `tfproviderlint-golangci` module plugin; its golangci-lint `version` must match the `.tools/go.mod` pin (depscheck enforces this).
- `make golangci-with-modules` builds `.tools/bin/golangci-with-modules` (gitignored), rebuilding when either pin changes. `make lint` / `lint-fix` / `tfproviderlint` use it.
- tfproviderlint now runs inside golangci-lint with **all** checks (`extended: true`, so the X* checks too) instead of the hand-listed `tfproviderlintx` invocation. The standalone `tfproviderlint` CI job and its failure-comment entry are removed.
- `terrafmt-acctests.sh` moves from the `tfproviderlint` target into `quick-checks`.

### tfproviderlint policy

Disabled, same rationale as azurerm: `AT004` (test configs intentionally declare `provider "azuread" {}`, 141 findings), `R009` (intentional panics in the test framework and generators), `R018` (deliberate eventual-consistency sleeps in test helpers). The `V*` checks are exempted for `internal/helpers/tf/validation/` only, since that package *is* the validation helpers those checks want callers to use. `AT001` is told about the `_data_source_test.go` suffix.

### New stock linters

copyloopvar, errname, fatcontext, gocheckcompilerdirectives, makezero, mirror, modernize, nilnesserr, noctx, nosprintfhostport, govet `nilness`, and the `gci` formatter, with azurerm's settings. modernize's `any` (~740 `interface{}` rewrites) and `newexpr` (`pointer.To` to `new()`) sub-checks are disabled, as in azurerm.

### Findings fixed

- modernize / copyloopvar / gci autofixes across 38 files (strings.Builder, `SplitSeq`, `slices.Contains`, range-over-int, loop var copies, import grouping)
- `R005`: 7 `HasChange() || HasChange()` chains combined into `HasChanges(...)`
- `XS001`: `triggers` on `azuread_synchronization_job_provision_on_demand` gets a Description. `XS002` (alphabetical schema order) is disabled: it only matches schemas written as `schema.Schema`, not the `pluginsdk.Schema` alias every other resource uses, so it would enforce ordering on that single file
- `AT003` / `AT012`: `TestAccRandTimeInt` was a unit test wearing the acceptance prefix, now `TestRandTimeInt`; `TestAccSynchronizationJob` (sequential runner) becomes `TestAccSynchronizationJob_sequential`; the two `TestAccGroup_preventDuplicateNames*` tests living in the administrative unit test file are renamed `TestAccAdministrativeUnit_preventDuplicateNames*`. The last two change TeamCity test names.

`make lint`, `make quick-checks`, `make test`, `make gencheck` all pass.

### Next

`azproviderlint` as a fifth PR: add the plugin to `.custom-gcl.yml`, enable, measure, disable-with-comment the noisy checks.


mstroob (dismissed):
LGTM 🎲

mstroob (approved):

PR #1953, opened 2026-09-16
typos: use UK spelling in prose

katbyte:
Makes `locale = "en"` explicit (it was relying on the default) and records why both spellings stay: schema descriptions are already British ("unique in the organisation"), while published names are American and cannot change — `azuread_access_package_catalog`, the `behaviors` attribute on `azuread_group`.

Then switches prose to UK spelling where the word is prose, not an attribute name.

Matches azurerm, whose issue template already reads "Expected Behaviour".


catriona-m (approved):
Thanks @katbyte LGTM!

